Issue #8, opened 2020-01-17
Open "Send" Page when a wallet exists


oshorefueled:
### Can the feature be described as a User story?

Given that a wallet has been created or restored.
When I’m on the send page, I should be able to:
* input the address I want to send DCR to
* input the amount I want to send in DCR and see it’s USD equivalent
* see the transaction fee
* see the total number of DCR I’ll be sending which includes transaction fees and my balance after a successful transaction
* see a confirmation page or popup modal asking me to confirm my inputs before sending DCR

### Additional context
Mobile mockups
<img width="324" alt="Screenshot 2020-01-17 at 12 33 31 PM" src="https://user-images.githubusercontent.com/15804688/72609575-9c797480-3925-11ea-9eb1-b8295f1d01a6.png">

confirmation modal
<img width="325" alt="Screenshot 2020-01-17 at 12 34 21 PM" src="https://user-images.githubusercontent.com/15804688/72609655-cd59a980-3925-11ea-8c8d-23b48f608a49.png">

PR #12, opened 2020-01-23
Create Send page

hydra0210:
### Resolves

Issue #8 

### What's new

This adds the send page to the app. The send page is displayed when at least one wallet is loaded upon starting the app

decred
decred2

### Additional context

_Add any other context about the problem here._


LICENCE
```diff
@@ -1,15 +1,15 @@
...
-copyright notice and this permission notice appear in all copies.
-
-THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
-WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
-MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
-ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
-WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
-ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
-OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ISC License
```
oluwandabira: Did anything change here?

ui/helper/clicker.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+)
+
+func NewClicker() Clicker {
+	return Clicker{
+		history: []clickItem{},
+	}
+}
+
+func (c *Clicker) Clicked(ctx *layout.Context) bool {
```
oluwandabira: Why use this instead of the widget's inbuilt clicked function?

ui/helper/color.go
```diff
@@ -0,0 +1,21 @@
...
+	SuccessColor = color.RGBA{227, 98, 9, 255}
+
+	DarkBlueColor  = color.RGBA{9, 20, 64, 255}
+	LightBlueColor = color.RGBA{41, 112, 255, 255}
+
+	OrangeColor = color.RGBA{237, 109, 71, 255}
+	GreenColor  = color.RGBA{46, 214, 161, 255} //color.RGBA{65, 191, 83, 255}
+
+	BackgroundColor = color.RGBA{248, 249, 250, 255}
+)
```
oluwandabira: Move out of helper, to just `ui/color`

ui/widgets/line.go
```diff
@@ -0,0 +1,59 @@
...
+type (
+	// Line represents a rectangle widget with an initial thickness of 1
+	Line struct {
+		height float32
+		color  color.RGBA
+	}
+)
+
+// NewLine returns a line widget instance
+func NewLine() *Line {
```
oluwandabira: The theme should be where widgets are created from.

ui/page/send.go
```diff
@@ -0,0 +1,481 @@
...
+			inset := layout.Inset{
+				Top:    unit.Dp(1),
+				Bottom: unit.Dp(1),
+			}
+			inset.Layout(gtx, func() {
+				pg.modalWidgets.line.Draw(gtx)
+			})
+		},
+		func() {
+			gtx.Constraints.Width.Min = gtx.Constraints.Width.Max
```
oluwandabira: Why?

ui/helper/clicker.go
yaowork: we need custom made widgets that are clickable. There's no inbuilt way of doing that. This helper gives us the feature of making our own widgets clickable

ui/page/send.go
yaowork: this is to make the following widget take up all available width

ui/helper/clicker.go
oluwandabira: Buttons are clickable, why not use buttons?

ui/page/send.go
oluwandabira: Why not change the widgets size instead of changing the layout size?

yaowork: need the layout to take up all width, rather

oluwandabira: ok

ui/themes/materialplus/line.go
```diff
@@ -0,0 +1,59 @@
...
+func (l *Line) SetColor(color color.RGBA) *Line {
+	l.color = color
+	return l
+}
+
+// SetHeight sets the line thickness
+func (l *Line) SetHeight(height float32) *Line {
+	l.height = height
+	return l
+}
```
oluwandabira: Why not export `height` and `color`?

ui/themes/materialplus/modal.go
```diff
@@ -0,0 +1,53 @@
...
+	overlayColor.A = 200
+
+	return &Modal{
+		overlayColor: overlayColor,
+		isOpen:       true,
+	}
+}
+
+// Draw renders the modal instamce to screen
+func (m *Modal) Draw(gtx *layout.Context, renderFunc func()) {
```
oluwandabira: This can be a single draw modal function, the page already tracks if the modal is open

oshorefueled:
can we have lines or some other form of indication for where to add inputs for a transaction? Right now the page looks like it's for displaying data rather than inputting them.

oshorefueled (changes requested):
Can we have the wallets dropdown list not shift the entire page content downwards when it's clicked?
Screenshot 2020-02-03 at 3 21 11 PM

oshorefueled:
Not sure if it's OS specific, but here's how the send page looks like when the popup is active. Notice how the covered page looks like? 
Screenshot 2020-02-03 at 3 20 47 PM

hydra0210:
> Not sure if it's OS specific, but here's how the send page looks like when the popup is active. Notice how the covered page looks like?
> Screenshot 2020-02-03 at 3 20 47 PM

fixed

ui/helper/helper.go
```diff
@@ -0,0 +1,99 @@
+package helper
```
oshorefueled: you can remove the helper package.

ui/send_page.go
```diff
@@ -1,21 +1,724 @@
...
+				}),
+				layout.Flexed(1, func() {
+					layout.Stack{Alignment: layout.NE}.Layout(gtx,
+						layout.Stacked(func() {
+							pg.balanceAfterSendValueLabel.Layout(gtx)
+						}),
+					)
+				}),
+			)
+		},
```
oshorefueled: These look like they have the same layout structure. Why not create a function of the structure and pass both widgets to the function? It will definitely eliminate repeated code.

README.md
```diff
@@ -1,18 +1,18 @@
...
+
+Note: You need to have [Go 1.13](https://golang.org/dl/) or above to build.
+
+Follow the [installation](https://gioui.org/doc/install) instructions for gio.
+
+Then `go build`.
+
+## Contributing
+
+See [CONTRIBUTING.md](https://github.com/raedahgroup/godcr-gio/blob/master/.github/CONTRIBUTING.md)
```
oshorefueled: There doesn't seem to be a difference between these two files.

wallet/wallet_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/decred/dcrd/dcrutil"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
```
oshorefueled: is there a need for space here? They're all packages from github.com.

ui/send_page.go
```diff
@@ -1,21 +1,724 @@
...
+func (pg *Send) setWallets(wallets []wallet.InfoShort) {
+	pg.wallets = wallets
+
+	for i := range wallets {
+		if len(wallets[i].Accounts) == 0 {
+			continue
+		}
+
+		pg.setSelectedAccount(wallets[i], wallets[i].Accounts[0])
+		pg.balanceAfterSendValueLabel.Text = dcrutil.Amount(pg.selectedWallet.SpendableBalance).String()
```
oshorefueled: "SpendableBalance" should be preprocessed from the wallet method. You shouldn't need to process values in the UI.

ui/send_page.go
```diff
@@ -1,21 +1,724 @@
...
+
+	//pg.passwordModal = theme.Password()
+	pg.destinationAddressEditor = &editor{
+		editor:   new(widget.Editor),
+		material: pg.theme.Editor("Destination Address"),
+		line:     pg.theme.Line(),
+	}
+
+	// TODO remvoe following line
+	pg.destinationAddressEditor.editor.SetText("TsfWXQ6M2QdtEGRd1mpGScWpEk8pgf8e9J5")
```
oshorefueled: are you still planning to remove this line?

ui/send_page.go
```diff
@@ -1,21 +1,724 @@
...
+	pg.isPasswordModalOpen = false
+
+	pg.wallet.BroadcastTransaction(pg.txAuthor, password)
+}
+
+func (pg *Send) cancel() {
+	pg.isPasswordModalOpen = false
+}
+
+func reduceWidgetsWrapper(widgetsWrapper [][]func()) []func() {
```
oshorefueled: You can add a comment on what this does.

ui/state.go
```diff
@@ -37,6 +43,12 @@ func (win *Window) updateStates(update interface{}) {
 	case wallet.DeletedWallet:
 		win.selected = 0
 		win.states.dialog = false
+	case *dcrlibwallet.TxAuthor:
+		win.states.txAuthor = update.(*dcrlibwallet.TxAuthor)
+	case *wallet.Broadcast:
+		fmt.Println("fff")
```
oshorefueled: this should be removed maybe?

ui/state.go
```diff
@@ -9,6 +12,9 @@ import (
 type states struct {
 	loading bool // true if the window is in the middle of an operation that cannot be stopped
 	dialog  bool // true if the window dialog modal is open
+
+	txAuthor        *dcrlibwallet.TxAuthor
```
oshorefueled: these values are not booleans for determining what the wallet is displaying. You should probably create a key for these in the Window struct and update them there.

ui/wallets_page.go
```diff
@@ -31,7 +33,7 @@ func (win *Window) WalletsPage() {
 								win.theme.Body1(acct.Name).Layout(win.gtx)
 							}),
 							layout.Rigid(func() {
-								win.theme.Body1(acct.TotalBalance).Layout(win.gtx)
+								win.theme.Body1(strconv.Itoa(int(acct.TotalBalance))).Layout(win.gtx)
```
oshorefueled: You seem to have chosen your version of the code over that on master when you rebased.

ui/window.go
```diff
@@ -59,6 +61,7 @@ func CreateWindow(wal *wallet.Wallet) (*Window, error) {
 	win.tabs.Flex.Spacing = layout.SpaceBetween
 	win.current = win.WalletsPage
 	win.dialog = func() {}
+	win.sendPage = win.NewSendPage(win.theme)
```
oshorefueled: You shouldn't need the "sendPage" in the Window struct. Also, if "NewSpendPage" is a method on "Window", there's no need to pass "win.theme" as an argument to it.

wallet/commands.go
```diff
@@ -172,38 +197,39 @@ func (wal *Wallet) GetMultiWalletInfo() {
...
-				BlockTimestamp:  wall.GetBestBlockTimeStamp(),
-				IsWaiting:       wall.IsWaiting(),
+				ID:               wall.ID,
+				Name:             wall.Name,
+				Accounts:         walletAccounts,
+				TotalBalance:     totalWalletBalance,
+				SpendableBalance: spendableWalletBalance,
+				BestBlockHeight:  wall.GetBestBlock(),
+				BlockTimestamp:   wall.GetBestBlockTimeStamp(),
+				IsWaiting:        wall.IsWaiting(),
```
oshorefueled: any reason this implementation should replace that on master?

oshorefueled (changes requested):

oluwandabira:
- There is a bit of lag when changing accounts.
- The send page should look and function similar to the receive page with regards to selecting an account and wallet.
- Sending should be disabled if the wallet is not synced (we should probably add back that sync button in the header).
- For the password entry, there should be only one field.

oluwandabira:
An example of the lag:
GoDcr-decred-wallet-2020-04-10-1(2)

oluwandabira:
Needs a rebase and destination address needs a copy-from-clipboard button

hts-log.txt
```diff
@@ -0,0 +1,16 @@
...
+	do not share these files/folders if you want these information to remain private
+
+08:16:29	Warning: 	Cache: damaged cache, trying to repair
+08:16:29	Warning: 	Cache: 0 bytes successfully recovered in 0 entries
+08:16:29	Warning: 	Cache: error trying to open the cache
+08:16:29	Error: 	"Unable to get server's address: Name or service not known" (-5) after 2 retries at link gui/robots.txt (from primary/primary)
+08:16:29	Warning: 	Retry after error -5 (Unable to get server's address: unknown error) at link gui/ (from primary/primary)
+08:16:29	Warning: 	Retry after error -5 (Unable to get server's address: unknown error) at link gui/ (from primary/primary)
+08:16:29	Error: 	"Unable to get server's address: unknown error" (-5) after 2 retries at link gui/ (from primary/primary)
+08:16:29	Warning: 	No data seems to have been transferred during this session! : restoring previous one!
```
oshorefueled: why do we need this?

version.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+// Version returns the current (semver) version number
+// of the app
+func Version() string {
+	return "0.0.1"
+}
```
oshorefueled: I've mentioned this before, this file shouldn't be on the commit.

ui/decredmaterial/editor.go
```diff
@@ -216,3 +225,19 @@ func (e Editor) handleEvents(gtx *layout.Context) {
 		e.editor.SetText("")
 	}
 }
+
+func (e *Editor) SetRequiredErrorText(txt string) {
+	e.requiredErrorText = txt
```
oshorefueled, line 230: Why do we need to set required error text? aren't all required message for input fields "field is required"?

ui/decredmaterial/theme.go
```diff
@@ -132,13 +137,52 @@ func (t *Theme) Surface(gtx *layout.Context, w layout.Widget) {
...
 	return ic
 }
 
+func mustDcrIcon(icon image.Image, err error) image.Image {
+	if err != nil {
+		panic(err)
+	}
+	return icon
+}
+
```
oshorefueled: any reason why this was added?

ui/send_page.go
```diff
@@ -1,18 +1,614 @@
...
+	isPasswordModalOpen       bool
+	isBroadcastingTransaction bool
+	hasInitializedTxAuthor    bool
+	hasCopiedTxHash           bool
+}
+
+const (
+	PageSend                 = "send"
+	PageSendTxAuthorErrChan  = "send-txauthor-err"
+	PageSendBroadcastErrChan = "send-broadcast-err"
```
oshorefueled: these should be in the page struct

oshorefueled (changes requested):

oshorefueled (approved):

PR #48, opened 2020-02-17
add custom text editor

hydra0210:
### What's new

This creates a custom editor that can mask it's input when set. It can also be set to only allow numeric characters by calling  `editor.IsNumeric()` function

### DIFF
--- a/ui/themes/materialplus/buffer.go
+++ b/ui/themes/materialplus/buffer.go
@@ -48,7 +48,7 @@ func (e *editBuffer) deleteRunes(runes int) {
 		e.gapend += s
 		e.changed = e.changed || s > 0
 	}
-	e.dump()
+	//e.dump()
 }
 
 // moveGap moves the gap to the caret position. After returning,
@@ -84,7 +84,7 @@ func (e *editBuffer) moveGap(space int) {
 		e.gapstart = e.caret
 		e.gapend = e.gapstart + l
 	}
-	e.dump()
+	//e.dump()
 }
 
 func (e *editBuffer) len() int {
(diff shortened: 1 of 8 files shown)


ui/themes/materialplus/buffer.go
```diff
@@ -0,0 +1,179 @@
...
+	e.moveGap(len(s))
+	copy(e.text[e.caret:], s)
+	e.gapstart += len(s)
+	e.changed = e.changed || len(s) > 0
+	e.dump()
+}
+
+func (e *editBuffer) dump() {
+	if bufferDebug {
+		fmt.Printf("len(e.text) %d e.len() %d e.gapstart %d e.gapend %d e.caret %d txt:\n'%+x'<-%d->'%+x'\n", len(e.text), e.len(), e.gapstart, e.gapend, e.caret, e.text[:e.gapstart], e.gapLen(), e.text[e.gapend:])
```
oluwandabira, line 150: Use the `log`

ui/themes/materialplus/editor.go
```diff
@@ -0,0 +1,868 @@
...
+	"gioui.org/op/paint"
+	"gioui.org/text"
+	"gioui.org/unit"
+	"gioui.org/widget"
+
+	"golang.org/x/image/math/fixed"
+)
+
+// Editor implements an editable and scrollable text area.
+type Editor struct {
```
oluwandabira: Is gioui's Editor not editable and scroll-able?

yaowork: this is a modification of the gioui's Editor codes to add the functionality of masking characters for use as a password editor. Gio doesn't provide this functionality out of the box

oluwandabira: Then the comment should reflect that. And if it is for passwords, why does it need to be scrollable?

oshorefueled:
password masking feature to be pushed upstream

PR #49, opened 2020-02-19
Internally manage required confirmations

oluwandabira:
Also makes change wallet creation and restoration to only use passwords

### DIFF
--- a/main.go
+++ b/main.go
@@ -49,7 +49,13 @@ func main() {
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -49,7 +49,13 @@ func main() {
 		}
 	}
 
-	wal, _ := wallet.NewWallet(cfg.HomeDir, cfg.Network, make(chan wallet.Response, 3))
+	var confirms int32 = 2
```
beansgum: use `dcrlibwallet.DefaultRequiredConfirmations`

beansgum (changes requested):

beansgum (approved):

PR #52, opened 2020-02-22
Rework how the front end ui works

oluwandabira:
`Window` now has a set of states(`bool`) and a set of inputs (`widget.Button` and `widget.Editor`'s). 
Each input is mapped to a specific action that can change the wallet state set and call multiwallet commands.
Updates received from the `wallet.Send` channel can change the state set.
After each update, the window determines what to display from the set of channels.
The current UI page is composed of a gioui's widgets and is itself a widget.
A special widget is the `Loading` widget, which is overlayed over the current page whenever the ui is in the middle of an operation.


oluwandabira:
Moved to https://github.com/RogueElement/godcr-gio/tree/rework

PR #55, opened 2020-03-01
Simplify front end code

oluwandabira:
- Pages are now methods of `Window`.
- Pages can use reusable UI components, which are also methods of Window.
- UI inputs (`widget.Button`'s and `widget.Editor`'s) and outputs (display elements) are declared on `Window` creation.
- All inputs are handled after the frame is laid out.
- Specific functionality is tied to only one UI input but can be laid out by multiple UI outputs.
- Dialogs are special pages that are laid out stacked on the current page.
- Loading is a special dialog laid out during wallet operations.
- Use a customized version of gioui's `material` : `decredmaterial`.


oluwandabira:
Loading dialog
Screenshot from 2020-03-06 00-07-42

oluwandabira:
Wallets page
Screenshot from 2020-03-06 00-07-49

oluwandabira:
Create Wallet dialog
Screenshot from 2020-03-06 00-12-26
Delete Wallet dialog
Screenshot from 2020-03-06 00-12-35

oluwandabira:
Current design is not final. It can be easily changed, the important part is that it's much easier to change now.

oshorefueled (approved):

PR #78, opened 2020-04-16
Add TabContainer widget support

### LINKED ISSUE #77: Add tabcontainer widget support
### Can the feature be described as a User story?
Adds support for tabcontainer support for trailing, leading, top and bottom layout.

metaclips:
### Resolves

Fixes #77

### What's new
Add tabcontaner widget support. Solves issue #77.
Tabcontainer icons supports both jpg, png and iconvg. Icons are set to 24 dp as to material design https://material.io/components/tabs#specs

### DIFF
--- a/ui/decredmaterial/tabcontainer.go
+++ b/ui/decredmaterial/tabcontainer.go
@@ -1,7 +1,6 @@
 package decredmaterial
 
 import (
-	"fmt"
 	"image"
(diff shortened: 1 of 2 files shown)


oshorefueled:
the same feature can be implemented with less code. The current tab implementation would be improved instead.

Issue #82, opened 2020-04-18
Seperate page widgets


oluwandabira:
Currently most widgets live in `Window.inputs` or `Window.outputs` and this has gotten quite cumbersome ( `Window.handleinputs` is also very long as a result).
This need to be improved.

oluwandabira:
I propose changing the window to hold a `map[string]layout.Widget` for all pages.

- Pages themselves should be a structure that has a method (something along the lines of a `page.Construct`) that takes pointers to the window data it needs and returns a `layout.Widget` that lays out the page when called.
- The construction method should only take the window data used.
- For widget data that can change the page should construct the widget within the layout function as the window will not communicate data changes to the pages.
- Dialogs will remain the same (a single `dialog` variable) but we expand the number of pages.
- There would pretty much only be about 3 dialogs (confirm, info, enter password) but other more specific dialogs can also be included.
-  If a dialog needs to call another dialog it should be a page.

PR #83, opened 2020-04-20
Split

oluwandabira:
### Resolves

Issue #82 

### What's new

- Added a `pages` field to Window for all the pages.
- Changed Window.current to a string representing the current page map key.
- Added string key constants for each page.
- Switched the Wallets page to a self contained page that declares
 and handles it's own widgets (this page is still in its preliminary
 stage).
 - Other pages were left as is but will be switched over in pull requests.
 - Dialogs are controlled by the pages themselves.
* Error handling for slow wallet operations is not supported yet


ui/wallet_page.go
```diff
@@ -0,0 +1,173 @@
...
+			},
+		}
+
+		page.dialog.LayoutIfActive(gtx, func() {
+			common.LayoutWithWallets(gtx, func() {
+				page.container.Layout(common.gtx, len(wdgs), func(i int) {
+					wdgs[i]()
+				})
+			})
+		})
```
oshorefueled: The page layout should be wrapping the dialog layout. Although currently it still produces the intended result, the page layout in the dialog is counter-intuitive.

ui/wallet_page.go
```diff
@@ -0,0 +1,173 @@
...
+				log.Error(err)
+			} else {
+				common.info.Wallets[*common.selectedWallet].Name = name
+				page.renaming = false
+			}
+		}
+
+		if page.delete.Clicked(gtx) {
+			page.dialog.SetDialog(page.dialog.deleteWallet(common))
+		}
```
oshorefueled: The event handling conditions should be in a separate method, the result is a much cleaner code. To make sure every page has an event handling method, you can define an interface which every valid page must implement.

oshorefueled (changes requested):

ui/wallet_page.go
oluwandabira: I actually prefer this to a separate method given that the only requirement a page has to satisfy now is returning a `layout.Widget` so adding an interface seems like unnecessary complexity.

ui/wallet_page.go
oluwandabira: It doesn't seem counter-intuitive to me since the dialog is not part of the page but rather a layer above it.

oshorefueled (approved):

Issue #75, opened 2020-04-16
Make Dedicated "create and restore wallet" page similar to wallet page implementation


oshorefueled:
The wallet page has a main layout method "WalletPage" which returns the entire layout of the page. The method is passed a "PageCommon" data type as an argument and added to a map of all pages.

The "create and restore wallet" page should:
* Have a main layout method which is also passed the common data type an argument and any other data type that might be needed in the page.
* Have a "Handle" method that handles the events on the page

### Can the feature be described as a User story?
When I open page the "create and restore" wallet page, I should be
*  see two buttons one to create wallet and the other for restore wallet
*  I should be able to access this page from the wallets page
*  I should also see this page when I have a no wallets 

A feature with the same goal was worked on in PR #22 

### Additional context
Mobile mockup
<img width="283" alt="Screenshot 2020-04-16 at 10 15 56 AM" src="https://user-images.githubusercontent.com/15804688/79438372-5aff7080-7fcb-11ea-8d21-d811670aa7c8.png">

PR #84, opened 2020-04-21
multi: create dedicated page for adding wallet

oshorefueled:
### Resolves

Issue #75 

### What's new
This PR:
* adds a dedicated page for accessing the create and restore functionalities.
* It also adds the page as the default page when no wallet exists. 
* adds a button on the wallets page to access the page
* removes the create and restore buttons in the header

Screenshot 2020-04-21 at 8 03 49 PM

Screenshot 2020-04-21 at 8 04 19 PM

Screenshot 2020-04-21 at 8 04 31 PM

Screenshot 2020-04-21 at 8 04 45 PM


Sirmorrison (changes requested):
image

the input fields and the suggestion buttons are not aligned.

Sirmorrison (changes requested):
image

this looks like you used the password modal instead of a page with password input fields. Any resaon why?

Sirmorrison (commented):
image

Do we really need the input field title label? we can set the `isTitleLabel = false` for the input fields. reason for this is the `...` after the `Input word 12`

oshorefueled:
> image
> 
> this looks like you used the password modal instead of a page with password input fields. Any resaon why?

I didn't use a password modal, I used a modal with input fields for the password and confirm password fields. There's no reason to have them on a separate page as the layout is slim and it follows the normal password input convention of other pages.

ui/create_restore_page.go
```diff
@@ -0,0 +1,555 @@
...
+type createRestore struct {
+	gtx          *layout.Context
+	theme        *decredmaterial.Theme
+	info         *wallet.MultiWalletInfo
+	wal          *wallet.Wallet
+	keyEvent     chan *key.Event
+	errChan      chan error
+	showRestore  bool
+	restoring    bool
+	showPassword bool
```
Sirmorrison, line 43: I think you meant `showCreate`

ui/create_restore_page.go
```diff
@@ -0,0 +1,555 @@
...
+								pg.Restore()()
+							} else {
+								pg.mainContent()()
+							}
+						}),
+					)
+				})
+			}),
+		)
+		if pg.showPassword {
```
Sirmorrison, line 155: This is supposed to show the create modal i believe. it should be `showCreat` instead.

Sirmorrison (changes requested):

oshorefueled:
> Do

the label should denote what number of seed is being inputted, since they have to be done in an order. You can check the mobile mockups to see how it was done there.  It can however be simplified to look exactly like that.

Sirmorrison:
> > Do
> 
> the label should denote what number of seed is being inputted, since they have to be done in an order. You can check the mobile mockups to see how it was done there. It can however be simplified to look exactly like that.

cool. Then lets remove the `...`

ui/create_restore_page.go
```diff
@@ -0,0 +1,555 @@
...
+			}),
+			layout.Flexed(1, func() {
+				layout.Center.Layout(pg.gtx, func() {
+					layout.Flex{}.Layout(pg.gtx,
+						layout.Rigid(func() {
+							pg.gtx.Constraints.Width.Max = pg.gtx.Constraints.Width.Max / 2
+							pg.inputsGroup(pg.seedListLeft, 16, 0)
+						}),
+						layout.Rigid(func() {
+							//fmt.Printf("max %v min %v \n", pg.gtx.Constraints.Width.Max, pg.gtx.Constraints.Width.Min)
```
Sirmorrison: If this is not used, lets remove it. no need keeping commented codes.

ui/create_restore_page.go
```diff
@@ -0,0 +1,555 @@
...
+
+	"github.com/raedahgroup/dcrlibwallet"
+	"github.com/raedahgroup/godcr/wallet"
+
+	"gioui.org/layout"
+	"gioui.org/text"
+	"gioui.org/unit"
+	"gioui.org/widget"
+	"github.com/raedahgroup/godcr/ui/decredmaterial"
+	"golang.org/x/exp/shiny/materialdesign/icons"
```
Sirmorrison, line 18: On a lighter note, could we try rearranging the imports? standard imports at the top, other imports below?. I think it makes our work cleaner and more orderly.

oshorefueled: > On a lighter note, could we try rearranging the imports? standard imports at the top, other imports below?. I think it makes our work cleaner and more orderly.

doesn't really matter. As long as it builds on the CI, it's fine.

Sirmorrison (changes requested):
image

* After a wallet restore seed has been successfully add, the password mode that shows up has a title of `create wallet` and the button at the bottom say `restore wallet`. I think this is misleading. Kindly have the modal title read `Restore Wallet`.

* Also when a user navigates out of the restore page, the input fields should be cleared.

Sirmorrison (approved):
PR looks good.

beansgum (approved):

PR #93, opened 2020-05-05
Icon generator 

oshorefueled:
### What's new

This PR adds a script to convert Decred icons from png to byte format. The source icon files would be added in the "ui/assets/decredicons" directory. 

To convert icons from png to bytes, navigate into the "ui" directory and use the command "go generate"


oshorefueled:
bundling the icons using pkger was implemented on #115

PR #115, opened 2020-05-26
use pkger in bundling decred icons

oshorefueled:
### What's new
This PR adds icon bundling for decred icons using pkger.


pkged.go
```diff
@@ -1,10 +1,8 @@
-// Code generated by pkger; DO NOT EDIT.
-
 package main
```
beansgum: Why is this file in the repo? Can it be generated by anyone?

beansgum (changes requested):

pkged.go
oshorefueled: Yes, but the user would need to install pkger globally first and then generate it. This way, you just build and get the app running. The bundled is already on the master branch, currently it bundles just the fonts.

beansgum: We can't have a generated binary checked into the repo. Each dev should have pkger and generate the binary locally(instructions should be added to the readme). Adding binary to commits will make the repo heavy in size.

beansgum (dismissed):

beansgum (approved):

Issue #123, opened 2020-05-30
Could not create wallet


song50119:
### Describe the bug
When syncing wallet, could not create new wallet

### To Reproduce
 - *On page create wallet*
 - *When I click on create new wallet*

### Expected behavior
When syncing wallet should disable button `add wallet` from `wallet page`

### Relevant screenshots or logs
```
2020-05-30 16:57:38.264 [ERR] UI: Wallet Error: Could not create wallet : sync_already_in_progress
2020-05-30 16:57:38.614 [INF] CMGR: DNS discovery failed on seed x5.testnet-seed.decred.netpurgatory.com: lookup x5.testnet-seed.decred.netpurgatory.com: no such host
```

PR #124, opened 2020-06-01
ui: Fix could not create wallet while syncing

song50119:
- Add (*walletPage).walletInfo
- Disable button "Add Wallet" while syncing

### Resolves

Issue #123

### DIFF
--- a/ui/wallet_page.go
+++ b/ui/wallet_page.go
@@ -2,6 +2,7 @@ package ui
 
 import (
 	"image/color"
+	"time"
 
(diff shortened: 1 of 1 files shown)


oshorefueled:
A notification should be displayed when the user clicks the add wallet button when its greyed out. You can have a look at the send page for how "alert" is used to display notifications to the user. The message could be "you have to stop sync to create a new wallet"

oshorefueled (dismissed):

oshorefueled (approved):

Issue #126, opened 2020-06-01
restore page flickers


oshorefueled:
### Describe the bug
the restore page starts to flicker when I scroll the inputs without click

### To Reproduce
On the restore page, type in an input field, when the suggestion input is displayed, scroll the row of inputs.

### Expected behavior
Page should display without flickering

### Actual behavior
Page flickers

### Relevant screenshots or logs
![restore page glitch](https://user-images.githubusercontent.com/15804688/83417705-62af9300-a41a-11ea-8632-cbc96fe0b43b.gif)

PR #128, opened 2020-06-03
use universal suggestion widgets for all editors

oshorefueled:
### Resolves

Issue #126 

### What's new

This PR replaces the previous seed suggestion implementation. Rather than defining suggestion seeds for each editor, it uses the same suggestion widgets for all editors only displaying them on active editors. It fixes the flickering of the restore page on scroll. It also fixes briefly displayed suggestions for the previously active editor when selecting another one.

### DIFF
--- a/ui/create_restore_page.go
+++ b/ui/create_restore_page.go
@@ -379,14 +379,8 @@ func (pg *createRestore) onSuggestionSeedsClicked() {
 }
 
 func (pg *createRestore) editorSeedsEventsHandler() {
-	var focused []int
-
 	for i := 0; i < len(pg.seedEditorWidgets.editors); i++ {
 		editor := &pg.seedEditorWidgets.editors[i]
-
-		if editor.Focused() && editor.Text() == "" {
-			focused = append(focused, i)
-		}
 
 		for _, e := range editor.Events(pg.gtx) {
 			switch e.(type) {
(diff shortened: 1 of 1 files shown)


Sirmorrison (changes requested):
image


So there is an issue here. if the seed suggestion shows up and i dont select any option at first and it leaves the screen, After clearing the editor field and trying again no suggestions show up.
Even if i move to the next editor nothing shows up. i looks like the code breaks at the point if there is no suggestion in the first editor.

To have this to work again i will need to restart the application.

Sirmorrison (changes requested):
image

Nice work so far just a couple of observation that could be addressed.
* i noticed that the initial implementation of this page, when a seed suggestion is selected, the editor focus is moved to the next available editor field. However, currently i will have to select the next editor field manually. 
for example the screenshot below i just selected a suggestion; the focus is supposed to move to number 2 howver its still on number 1.
image

* After the wallet is successfully restored, if you try to restore a wallet again, i noticed the seeds from the previous wallet restoration still exist (first screenshot). I believe the fields are supposed to be cleared after a successful wallet restoration.
* In the even a user mixes seed phrases up while inputting them, he gets an error telling him his seed phrase is not accurate however say which seed position is the giving the issue and with this the user will have to enter the seed phrases again from scratch. Currently, the user will have to manually delete all 33 seed phrases from the input fields which is lot of work. Can we add a `RESET` button to the page to clear all the fields?

oshorefueled:
your first and second bullet points was implemented on the last push. Your last bullet point would be on a separate pull request. This PR is only to fix the suggestion seeds flickering on scroll.

Sirmorrison (dismissed):
PR has passed all checks so far Looks good to go.

beansgum (approved):

Issue #121, opened 2020-05-29
sort list of wallets when "GetMultiWalletInfo" wallet command is called


oshorefueled:
### Describe the bug
Each time the "GetMultiWalletInfo" wallet command is called, it returns the list of wallets in a new order. This affects pages using the list as it also makes dependent UI components rendered randomly as well.
### To Reproduce
Steps to reproduce the behaviour:
When syncing wallets, in the wallet details section, observe the order of the wallet cards. 

### Expected behaviour
The order of the wallet detail cards shouldn't change at any point of the sync

### Actual behaviour
The order of the wallet detail card should be the same through out the sync process in an increasing order of their wallet IDs

PR #129, opened 2020-06-04
sort wallet order

hydra0210:
### Resolves

Issue #121 

### What's new

This sorts the wallet slice returned by "GetMultiWalletInfo" such that the order of wallets returned is always maintained.

### DIFF
--- a/wallet/commands.go
+++ b/wallet/commands.go
@@ -117,19 +117,26 @@ func (wal *Wallet) CreateTransaction(walletID int, accountID int32, errChan chan
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


oshorefueled (changes requested):
I believe the not sorted wallets is originally from the call of " wal.multi.OpenedWalletIDsRaw()" in the wallets method. It would be better to sort the wallet id from there since the wallets method is called by other methods other than "GetMultiWalletInfo"

oshorefueled (approved):

Issue #125, opened 2020-06-01
Move add account feature to the new wallet layout.


Sirmorrison:
Add the create account feature to the new wallet layout.

PR #131, opened 2020-06-05
Add account

Sirmorrison:
### Resolves

Issue #125 

### What's new

This PR adds the wallet account creation feature.

image

### DIFF
--- a/wallet/commands.go
+++ b/wallet/commands.go
@@ -88,20 +88,38 @@ func (wal *Wallet) DeleteWallet(walletID int, passphrase []byte, errChan chan er
 
 // AddAccount adds an account to a wallet.
 // It is non-blocking and sends its result or any error to wal.Send.
-func (wal *Wallet) AddAccount(walletID int, name string, pass []byte) {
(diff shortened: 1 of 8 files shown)


wallet/commands.go
```diff
@@ -88,20 +88,38 @@ func (wal *Wallet) DeleteWallet(walletID int, passphrase []byte, errChan chan er
...
-func (wal *Wallet) AddAccount(walletID int, name string, pass string) {
+func (wal *Wallet) AddAccount(walletID int, name string, pass []byte, errChan chan error) {
 	go func() {
+		var resp Response
 		wall := wal.multi.WalletWithID(walletID)
 		if wall == nil {
+			go func() {
+				errChan <- ErrIDNotExist
+			}()
+			resp.Err = ErrIDNotExist
```
oshorefueled, line 99: There should be a return statement if wallet id doesn't exist

oshorefueled (changes requested):

oshorefueled (approved):

Issue #143, opened 2020-06-08
balance after send inaccurate until new field input


oshorefueled:
### Describe the bug
The value of balance after send displays "0" when I land on the send page. The correct value isn't computed until I start typing into the address or amount field. Also when I switch between wallets, the balance after send is that of the previous wallet until I start typing into an input field. 

### To Reproduce
 -  navigate to the send page, balance after send is "0". 
-  switch to another wallet

### Expected behavior
-On landing on the send page, the balance after send should be the active wallet balance. It shouldn't wait until an input field is filled.
-When I switch between wallets, I should get the balance after send of the current wallet not that of the previously computed wallet.

### Actual behavior
 -  navigate to the send page, balance after send is "0". After typing into an input field, only then is it inputted correctly.
- switch to another wallet, the balance after send remains the same.

### Relevant screenshots or logs
*If applicable, add screenshots or log entries to help explain your problem.*
<img width="800" alt="Screenshot 2020-06-08 at 8 18 10 PM" src="https://user-images.githubusercontent.com/15804688/84071201-3caf6300-a9c5-11ea-847d-4c22f23563bb.png">

PR #146, opened 2020-06-10
Fix: corrects balance after send before and after transaction

sam-0x:
### Resolves

Issue #143


oshorefueled:
you have commits that aren't related to your PR.

PR #147, opened 2020-06-11
Add Seed reset button to restore page.

Sirmorrison:
### Resolves

Issue #148 

### What's new

This PR adds a reset button to the restore page to enable users reset the seed fields in the event he makes a mistake.

image

image

### DIFF
--- a/ui/decredmaterial/theme.go
+++ b/ui/decredmaterial/theme.go
@@ -102,14 +102,17 @@ func (t *Theme) Modal(gtx *layout.Context, title string, wd []func()) {
 					t.H4(title).Layout(gtx)
 				},
 				func() {
-					t.Line().Layout(gtx)
+					line := t.Line()
+					// line.Color = t.Color.Text
+					line.Width = gtx.Constraints.Width.Max
+					line.Layout(gtx)
 				},
 			}
 			w = append(w, wd...)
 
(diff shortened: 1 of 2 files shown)


oshorefueled:
The warning message would be better as a modal. You should give users an option to cancel the seed reset. Currently all a user can do is click "Yes" to proceed, there's no way to cancel or clear the warning.

Sirmorrison:
> The warning message would be better as a modal. You should give users an option to cancel the seed reset. Currently all a user can do is click "Yes" to proceed, there's no way to cancel or clear the warning.

I initially used a modal however the size of the modal was too big for the text. Also to hide the message all the user need to do is continue typing and it will reset. 

May be i can add that to the message that is displayed.

ui/decredmaterial/theme.go
```diff
@@ -102,14 +102,17 @@ func (t *Theme) Modal(gtx *layout.Context, title string, wd []func()) {
 					t.H4(title).Layout(gtx)
 				},
 				func() {
-					t.Line().Layout(gtx)
+					line := t.Line()
+					// line.Color = t.Color.Text
```
oshorefueled: any reason why the commented line was left there?

oshorefueled (changes requested):

oshorefueled (approved):

Issue #142, opened 2020-06-08
amount in USD


oshorefueled:
### Can the feature be described as a User story?
When I'm on send page and I have sufficient balance, 
* I should be able to switch between inputting the amount in DCR and USD. 
* See what currency I'm sending in USD or DCR

### Additional context
Amount input on mobile mock up
<img width="228" alt="Screenshot 2020-06-08 at 7 45 04 PM" src="https://user-images.githubusercontent.com/15804688/84068386-c6106680-a9c0-11ea-98a4-df0a8329fd64.png">

song50119:
This is can close?

PR #152, opened 2020-06-14
Send Page amount USD.

Sirmorrison:
### Resolves

Issue #142 

### What's new
This PR adds the send amount USD equivalence for all transactions fields to the send Page.

image

image

image

### DIFF
--- a/ui/send_page.go
+++ b/ui/send_page.go
@@ -1,13 +1,12 @@
 package ui
 
 import (
+	"encoding/json"
 	"fmt"
 	"image/color"
 	"net/http"
 	"strconv"
 	"time"
-	// "io/ioutil"
-	"encoding/json"
 
 	"gioui.org/layout"
 	"gioui.org/unit"
@@ -46,27 +45,10 @@ type SendPage struct {
 	activeTransactionFeeValueLabel   decredmaterial.Label
 	inactiveTransactionFeeValueLabel decredmaterial.Label
 
-	activeTransactionFeeValue   string
-	inactiveTransactionFeeValue string
-
-	activeTotalCostValue   string
-	inactiveTotalCostValue string
-
-	balanceAfterSendValue string
-
 	activeTotalCostValueLabel   decredmaterial.Label
 	inactiveTotalCostValueLabel decredmaterial.Label
 
 	balanceAfterSendValueLabel decredmaterial.Label
-
-	// activeExchLabel decredmaterial.Label
-	// inactiveExchLabel decredmaterial.Label
-
-	// transactionFeeValueLabel   decredmaterial.Label
-	// totalCostValueLabel        decredmaterial.Label
-	// balanceAfterSendValueLabel decredmaterial.Label
-	// txFeeLabelUSD decredmaterial.Label
-	// totalCostLabelUSD decredmaterial.Label
 
 	destinationAddressEditorMaterial     decredmaterial.Editor
 	sendAmountEditorMaterial             decredmaterial.Editor
@@ -84,10 +66,20 @@ type SendPage struct {
 	txHashText         string
 	txHash             string
 	calculateErrorText string
-	activeAmount       string
-	inactiveAmount     string
-	selectedExchange   string
-	unselectedExchange string
+
+	activeTotalAmount   string
+	inactiveTotalAmount string
+
+	activeExchange   string
+	inactiveExchange string
+
+	activeTransactionFeeValue   string
+	inactiveTransactionFeeValue string
+
+	activeTotalCostValue   string
+	inactiveTotalCostValue string
+
+	balanceAfterSendValue string
 
 	passwordModal *decredmaterial.Password
 
@@ -130,17 +122,12 @@ func (win *Window) SendPage(common pageCommon) layout.Widget {
 		sendErrorText: "",
 		txHashText:    "",
 
-		selectedExchange:   "DCR",
-		unselectedExchange: "USD",
+		activeExchange:   "DCR",
+		inactiveExchange: "USD",
 
 		closeConfirmationModalButtonMaterial: common.theme.Button("Close"),
 		nextButtonMaterial:                   common.theme.Button("Next"),
 		confirmButtonMaterial:                common.theme.Button("Confirm"),
-		activeTransactionFeeValueLabel:       common.theme.Body2("- DCR"),
-		activeTotalCostValueLabel:            common.theme.Body2("- DCR"),
-		balanceAfterSendValueLabel:           common.theme.Body2("- DCR"),
-		inactiveTransactionFeeValueLabel:     common.theme.Body2("(- USD)"),
-		inactiveTotalCostValueLabel:          common.theme.Body2("(- USD)"),
 
 		copyIconMaterial: common.theme.IconButton(mustIcon(decredmaterial.NewIcon(icons.ContentContentCopy))),
 
@@ -161,15 +148,6 @@ func (win *Window) SendPage(common pageCommon) layout.Widget {
 	page.walletsTab.Position = decredmaterial.Top
 	page.accountsTab.Position = decredmaterial.Top
 
-	page.activeAmount = "- " + page.selectedExchange
-	page.inactiveAmount = "- " + page.unselectedExchange
-
-	page.activeTransactionFeeValue = "- " + page.selectedExchange
-	page.inactiveTransactionFeeValue = fmt.Sprintf("(- %s)", page.unselectedExchange)
-
-	page.activeTotalCostValue = "- " + page.selectedExchange
-	page.inactiveTotalCostValue = fmt.Sprintf("(- %s)", page.unselectedExchange)
-
 	page.balanceAfterSendValue = "- DCR"
 
 	page.inactiveTransactionFeeValueLabel.Color = common.theme.Color.Hint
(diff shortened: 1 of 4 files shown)


oshorefueled:
The amount-usd feature doesn't work. The currency conversion feature isn't just to show a user DCR equivalent in dollars but to also allow the user input an amount in dollars and have its DCR equivalent sent. Currently, it doesn't do that.

oshorefueled:
The position of the conversion button and values look off, we should be able to make it look just like the mobile. Kindly have a look at the related issue for the mobile mockup. 

Screenshot 2020-06-19 at 7 24 41 AM

Sirmorrison:
> The position of the conversion button and values look off, we should be able to make it look just like the mobile. Kindly have a look at the related issue for the mobile mockup.
> 
> Screenshot 2020-06-19 at 7 24 41 AM

Resolved.

ezgif com-video-to-gif(9)

ui/send_page.go
```diff
@@ -16,6 +19,10 @@ import (
 	"golang.org/x/exp/shiny/materialdesign/icons"
 )
 
+type data struct {
```
oshorefueled: why do we need a new data type to handle just one property?

ui/send_page.go
```diff
@@ -260,6 +305,9 @@ func (pg *SendPage) setAccountTabs() {
 }
 
 func (pg *SendPage) Layout(common pageCommon) {
+	go pg.getUSDValues(&pg.data)
```
oshorefueled: This method is going to be called repeatedly, and even though it's called from a go routine for optimal reasons that's still a lot of go routines to spin up. Wouldn't it be better if this was only called when a user types the amount into the amount field?

ui/send_page.go
```diff
@@ -400,13 +448,13 @@ func (pg *SendPage) drawSelectedAccountSection(gtx *layout.Context) {
 func (pg *SendPage) drawTransactionDetailWidgets(gtx *layout.Context) {
 	w := []func(){
 		func() {
-			pg.tableLayoutFunc(gtx, pg.theme.Body2("Transaction Fee"), pg.transactionFeeValueLabel)
+			pg.tableLayoutFunc(gtx, pg.theme.Body2("Transaction Fee"), pg.activeTransactionFeeValue, pg.inactiveTransactionFeeValue)
```
oshorefueled: you don't need to add "Func" as a suffix to a method name.

ui/send_page.go
```diff
@@ -560,44 +655,138 @@ func (pg *SendPage) validateAmount(ignoreEmpty bool) bool {
...
+func (pg *SendPage) getUSDValues(target interface{}) {
+	url := "https://api.bittrex.com/v3/markets/DCR-USDT/ticker"
+	resp, err := http.Get(url)
+	if err != nil {
+		return
+	}
+
+	defer resp.Body.Close()
+
+	json.NewDecoder(resp.Body).Decode(target)
```
oshorefueled: This can be in the wallet package and add a comment about its function.

ui/send_page.go
Sirmorrison: The idea is to ensure the exchange is readily available at the point of typing the input and also not to make the app to freeze while trying to get the exchange data.

Sirmorrison: Removing to go routine and moving it to the calculatedvalue function.

oshorefueled: > The idea is to ensure the exchange is readily available at the point of typing the input and also not to make the app to freeze while trying to get the exchange data.

Still, calling an endpoint that frequent isn't an optimal solution to that. If real time time updates is what is needed, using sockets would be a better approach. Recursive calls to an endpoint, is a No No. Plus I don't think the exchange rates change that frequently to be called multiple times in a fraction of a second.

wallet/commands.go
```diff
@@ -488,6 +490,19 @@ func (wal *Wallet) VerifyMessage(address string, message string, signature strin
...
+	resp, err := http.Get(url)
+	if err != nil {
+		return
+	}
+
+	defer resp.Body.Close()
+
+	json.NewDecoder(resp.Body).Decode(target)
+}
+
```
oshorefueled: this doesn't belong here, it's not a wallet command. It should be in wallet.go just like the "GetBlockExplorerURL" method. Also, the comment should be more descriptive.   "GetUSDExchangeValues gets the exchange rate of DCR - USDT from a specified endpoint" would be more appropriate.

ui/send_page.go
```diff
@@ -560,44 +649,127 @@ func (pg *SendPage) validateAmount(ignoreEmpty bool) bool {
...
+		txFeeValueUSD := dcrutil.Amount(txFee).ToCoin() * usdExchangeRate
+
+		totalAmountUSDTostring := fmt.Sprintf("%s USD", strconv.FormatFloat(totalAmountUSD, 'f', 7, 64))
+
+		pg.activeTotalAmount = dcrutil.Amount(amountAtoms).String()
+		pg.inactiveTotalAmount = totalAmountUSDTostring
+		pg.activeTransactionFeeValue = dcrutil.Amount(txFee).String()
+		pg.inactiveTransactionFeeValue = fmt.Sprintf("(%f USD)", txFeeValueUSD)
+		pg.activeTotalCostValue = dcrutil.Amount(totalCostDCR).String()
+		pg.inactiveTotalCostValue = fmt.Sprintf("(%s USD)", strconv.FormatFloat(totalAmountUSD+txFeeValueUSD, 'f', 7, 64))
```
oshorefueled: This looks repeated. To make it cleaner, you can create a data type for it. For each case, assign the respective data properties their respective methods and pass that to a single method that initiates those values to the SendPage properties.
 Also, all that would be cleaner if they were in a separate method.

oshorefueled (changes requested):

ui/send_page.go
```diff
@@ -560,44 +658,137 @@ func (pg *SendPage) validateAmount(ignoreEmpty bool) bool {
...
-	amountAtoms := int64(amount)
 
-	// set destination address
+	amountAtoms := int64(amount)
 	pg.txAuthor.RemoveSendDestination(0)
 	pg.txAuthor.AddSendDestination(pg.destinationAddressEditor.Text(), amountAtoms, false)
+	return amountAtoms
+}
+
+func (pg *SendPage) setUSDValues(usdExchangeRate, inputAmount float64, totalCostDCR, txFee, amountAtoms int64) {
```
oshorefueled: make this a data type and pass a single argument to the Set method.

ui/send_page.go
```diff
@@ -560,44 +658,137 @@ func (pg *SendPage) validateAmount(ignoreEmpty bool) bool {
...
+	// calculate total tx amount in DCR
+	txFeeValueUSD := dcrutil.Amount(txFee).ToCoin() * usdExchangeRate
+	totalAmountUSDTostring := fmt.Sprintf("%s USD", pg.sendAmountEditor.Text())
+
+	pg.activeTotalAmount = totalAmountUSDTostring
+	pg.inactiveTotalAmount = dcrutil.Amount(amountAtoms).String()
+	pg.activeTransactionFeeValue = fmt.Sprintf("%f USD", txFeeValueUSD)
+	pg.inactiveTransactionFeeValue = fmt.Sprintf("(%s)", dcrutil.Amount(txFee).String())
+	pg.activeTotalCostValue = fmt.Sprintf("%s USD", strconv.FormatFloat(inputAmount+txFeeValueUSD, 'f', 7, 64))
+	pg.inactiveTotalCostValue = fmt.Sprintf("(%s )", dcrutil.Amount(totalCostDCR).String())
```
oshorefueled: you only need one method to set these.

ui/send_page.go
```diff
@@ -560,44 +658,137 @@ func (pg *SendPage) validateAmount(ignoreEmpty bool) bool {
...
+	txFeeValueUSD := dcrutil.Amount(txFee).ToCoin() * usdExchangeRate
+
+	totalAmountUSDTostring := fmt.Sprintf("%s USD", strconv.FormatFloat(totalAmountUSD, 'f', 7, 64))
+
+	pg.activeTotalAmount = dcrutil.Amount(amountAtoms).String()
+	pg.inactiveTotalAmount = totalAmountUSDTostring
+	pg.activeTransactionFeeValue = dcrutil.Amount(txFee).String()
+	pg.inactiveTransactionFeeValue = fmt.Sprintf("(%f USD)", txFeeValueUSD)
+	pg.activeTotalCostValue = dcrutil.Amount(totalCostDCR).String()
+	pg.inactiveTotalCostValue = fmt.Sprintf("(%s USD)", strconv.FormatFloat(totalAmountUSD+txFeeValueUSD, 'f', 7, 64))
```
oshorefueled: you only need one method to set these.

ui/send_page.go
```diff
@@ -560,44 +658,137 @@ func (pg *SendPage) validateAmount(ignoreEmpty bool) bool {
...
+}
+
+func (pg *SendPage) updateDefaultValues(totalCostDCR, txFee, amountAtoms int64, noExchangeText string) {
+	pg.activeTotalAmount = dcrutil.Amount(amountAtoms).String()
+	pg.inactiveTotalAmount = noExchangeText
+	pg.activeTransactionFeeValue = dcrutil.Amount(txFee).String()
+	pg.inactiveTransactionFeeValue = ""
+	pg.activeTotalCostValue = dcrutil.Amount(totalCostDCR).String()
+	pg.inactiveTotalCostValue = ""
+}
```
oshorefueled: this is the only method you need to set the data. And it only really needs to accept one argument.

oshorefueled (changes requested):

ui/send_page.go
```diff
@@ -260,6 +309,8 @@ func (pg *SendPage) setAccountTabs() {
 }
 
 func (pg *SendPage) Layout(common pageCommon) {
+	pg.calculateValues()
```
oshorefueled: this shouldn't be called on every frame.

oshorefueled (changes requested):

oshorefueled (approved):

Issue #171, opened 2020-06-30
tab indicator doesn't change when "See All" button is clicked


oshorefueled:
#### Describe the bug
When the "see all" button on the overview page is clicked, the page navigates to the transactions page but the tab indicator remains on "overview".

#### Expected behavior
The tab indictor should switch to "transactions"

### Relevant screenshots or logs
<img width="502" alt="Screenshot 2020-06-30 at 2 34 09 PM" src="https://user-images.githubusercontent.com/15804688/86132460-d064e280-bade-11ea-9615-dcd40eee1504.png">

PR #176, opened 2020-07-02
Tab indicator change

oshorefueled:
####  Resolves
Issue #171. 

This PR replaces the previous tab change event implementation that was based on change variable with a slice implementation. Whenever there's a change event on the tab widget, the event is added to the event slice. The event slice is cleared on the next frame, enough time for every part of the app watching the event.

### DIFF
--- a/ui/overview_page.go
+++ b/ui/overview_page.go
@@ -66,6 +66,8 @@ type overviewPage struct {
 	listContainer, walletSyncList *layout.List
 	gtx                           *layout.Context
 	theme                         *decredmaterial.Theme
+	tab  					      *decredmaterial.Tabs
+
 	walletInfo                    *wallet.MultiWalletInfo
 	walletSyncStatus              *wallet.SyncStatus
 	walletTransactions            **wallet.Transactions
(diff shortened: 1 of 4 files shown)


ui/overview_page.go
```diff
@@ -84,15 +86,17 @@ type overviewPage struct {
...
+		tab:   c.navTab,
+
 		walletInfo:         win.walletInfo,
 		walletSyncStatus:   win.walletSyncStatus,
 		walletTransactions: &win.walletTransactions,
 		walletTransaction:  &win.walletTransaction,
 		listContainer:      &layout.List{Axis: layout.Vertical},
 		walletSyncList:     &layout.List{Axis: layout.Horizontal},
-		toTransactions:     c.theme.Button("more"),
+		toTransactions:     c.theme.Button("see all"),
```
Sirmorrison: Mobile says `See all` not `see all`

Sirmorrison (changes requested):

Sirmorrison:
Kind rebase the PR.

ui/overview_page.go
Sirmorrison: image

this review was not implemented.

Sirmorrison (approved):
PR looks good

beansgum (approved):

Issue #169, opened 2020-06-29
amount input should not take entire page width


oshorefueled:
#### describe the feature
Currently, the amount input field takes the full width of the send page. It should only be as wide as the maximum input size acceptable.

oshorefueled:
PR #180 resolves this

PR #179, opened 2020-07-07
Limit editor width of Send page

song50119:
### Resolves

Issue #169 

### What's new

Limit editor width of Send page

Screen Shot 2020-07-07 at 21 17 14


oshorefueled:
#180 removes the need for this.

Issue #178, opened 2020-07-06
ui: dcr-usd conversion on send page


oshorefueled:
the design of DCR to USD conversion on send page should look like the mobile. Wrapped in a border and the currency symbol on the same line as the editor. 

<img width="319" alt="Screenshot 2020-07-06 at 10 22 03 PM" src="https://user-images.githubusercontent.com/15804688/86646503-38cc2c00-bfd7-11ea-9329-34055a9a56d9.png">

PR #180, opened 2020-07-08
Send amount field

Sirmorrison:
### Resolves

Issue #178

### What's new

This PR implements the send page amount input field just as the mobile app.
* Moved the  amount input to the front of the currency symbol.
* Made the input amount only as wide as the available amount to be sent.

image

image

### DIFF
--- a/ui/send_page.go
+++ b/ui/send_page.go
@@ -138,7 +138,7 @@ func (win *Window) SendPage(common pageCommon) layout.Widget {
 		isPasswordModalOpen:       false,
 		hasCopiedTxHash:           false,
 		isBroadcastingTransaction: false,
-		width:                     27,
+		width:                     29,
 
 		passwordModal:    common.theme.Password(),
 		broadcastErrChan: make(chan error),
@@ -166,7 +166,7 @@ func (win *Window) SendPage(common pageCommon) layout.Widget {
 	page.sendAmountEditor.IsUnderline = false
 	page.sendAmountEditor.Editor.SingleLine = true
 	page.sendAmountEditor.Editor.SetText("0")
-	page.sendAmountEditorMaterial.TextSize = values.TextSize24
+	page.sendAmountEditor.TextSize = values.TextSize24
 
 	pg.closeConfirmationModalButton.Background = common.theme.Color.Gray
 
@@ -465,6 +465,7 @@ func (pg *SendPage) tableLayout(gtx layout.Context, leftLabel decredmaterial.Lab
 	)
 }
 
+<<<<<<< HEAD
 func (pg *SendPage) destinationAddrSection(gtx layout.Context) layout.Dimensions {
 	main := layout.UniformInset(values.MarginPadding20)
 	return pg.sectionLayout(gtx, main, func(gtx C) D {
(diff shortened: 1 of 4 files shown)


ui/send_page.go
```diff
@@ -444,19 +453,37 @@ func (pg *SendPage) sendAmountLayout(gtx layout.Context) layout.Dimensions {
...
+								return pg.sendAmountEditor.Layout(gtx)
+							}),
+							layout.Rigid(func(gtx C) D {
+								// this adjusts space between input and currency symbol.
+								m := values.MarginPadding5
+								e := pg.sendAmountEditor.Editor.Len()
+								if e > 0 && e < 7 {
+									m = values.EditorWidth
+								}
+								fmt.Println(pg.width)
```
oshorefueled: this should be removed.

ui/send_page.go
```diff
@@ -755,8 +780,26 @@ func (pg *SendPage) watchForBroadcastResult() {
 }
 
 func (pg *SendPage) changeEvt(evt widget.EditorEvent) {
+	// this adjusts the width of the amount editor per text input.
```
oshorefueled: This would be better as the comment to the method. The comment should also include what values specified in the switch statement do. Why those values?

oshorefueled (changes requested):

ui/send_page.go
```diff
@@ -754,9 +778,29 @@ func (pg *SendPage) watchForBroadcastResult() {
...
+	cal := func(val int) int {
+		return n - val
+	}
 	switch evt.(type) {
 	case widget.ChangeEvent:
+		textLength = textLength + 1
+		switch {
+		case pg.width <= 80:
+			pg.width = textLength * cal(0)
+		case pg.width > 80 && pg.width < 137:
```
oshorefueled: *  why are those specific values, 80, 137 and 167 used? the comment doesn't state that. 
*  "handle" would be a better description for what changeEvt does, rather than track it.
*  also, "ssend" in the comment.
*  you can also state very briefly why the editor width is adjusted on a change event

oshorefueled (changes requested):

raedah:
Horizontal line between the two currency values is confusing. Its is how its implemented on mobile, but our other input fields are marked by a line so look the same. Input fields on mobile are boxes.

Sirmorrison:
> Horizontal line between the two currency values is confusing. Its is how its implemented on mobile, but our other input fields are marked by a line so look the same. Input fields on mobile are boxes.

The mobile design had the dividing line. The line would make more sense after #184 is merged as it places the ui elements into different sections.

ui/send_page.go
```diff
@@ -912,9 +946,21 @@ func (pg *SendPage) watchForBroadcastResult() {
 	}
 }
 
-func (pg *SendPage) changeEvt(evt widget.EditorEvent) {
+// handleEditorChange tracks changes on the editor and adjust its width of the send amount input field
```
oshorefueled: "handles" describes the function of the method better than "track"

ui/send_page.go
```diff
@@ -912,9 +946,21 @@ func (pg *SendPage) watchForBroadcastResult() {
...
+// it also updates the DCR - USD exchange rate value
+func (pg *SendPage) handleEditorChange(evt widget.EditorEvent) {
```
oshorefueled, line 950: Why does a single method handle events for both the amount editor and the destination address editor?

oshorefueled (changes requested):

ui/send_page.go
Sirmorrison: How do you mean address editor? It handles only events in the amount editor.

oshorefueled: handleEditorChange method  is called on lines 299 and 308 to handle change events on the amount editor and destination address editor. On change event of the address field, it calls calculateNextWidth function and calls the GetUSDExchangeValues as well. Shouldn't these methods on be called on change event of the amount field?

ui/send_page.go
```diff
@@ -912,9 +946,21 @@ func (pg *SendPage) watchForBroadcastResult() {
...
 
-func (pg *SendPage) changeEvt(evt widget.EditorEvent) {
+// handleEditorChange handles changes on the editor and adjust its width of the send amount input field
+// it also updates the DCR - USD exchange rate value
+func (pg *SendPage) handleEditorChange(evt widget.EditorEvent) {
+	editorTextLength := pg.sendAmountEditor.Editor.Len()
+
+	// calulateNextWidth use the values of the defualtEditorWidth(the editor text size) and
+	// total number of text in the editor to determine the width of the amount field
+	calulateNextWidth := func() {
```
oshorefueled: fix typo, "calulate"

oshorefueled (changes requested):

oshorefueled (approved):

Issue #177, opened 2020-07-06
notifications on send page


oshorefueled:
#### Describe the bug
* There's no way to clear the notification on the send page when a transactions is successful created. The notification doesn't need to show the transaction hash for a copy, it can simple show "Transaction Sent" like the mobile and disappear after some seconds, 3 seconds is fine.

* The notification background shouldn't fill the entire page width. It should only take the width of the notification message with padding applied on the sides

* There shouldn't be an error notification on the confirm modal. A user probably has to close the modal to resolve the cause of the error anyway. 

#### Relevant screenshots or logs

<img width="597" alt="Screenshot 2020-06-30 at 7 41 11 PM" src="https://user-images.githubusercontent.com/15804688/86640759-7da19400-bfd2-11ea-9011-241e248613ad.png">

<img width="599" alt="Screenshot 2020-06-30 at 7 48 57 PM" src="https://user-images.githubusercontent.com/15804688/86640769-7ed2c100-bfd2-11ea-9be1-f493761c6065.png">

PR #183, opened 2020-07-11
move error alert to main send page

hydra0210:
### Resolves

Issue #177 

### What's new
This does the following: 

- Changes the send success alert message from ```transaction hash``` to "Transaction Sent"
- Clears the send success alert message after 3 seconds 
- Resizes the alert messages to take up the width of it's content plus padding 
- Moves the error message alert from the confirmation modal to the send page itself

### DIFF
--- a/ui/send_page.go
+++ b/ui/send_page.go
@@ -2,7 +2,6 @@ package ui
 
 import (
 	"fmt"
-	"image"
 	"image/color"
 	"strconv"
 	"time"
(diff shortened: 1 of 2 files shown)


ui/send_page.go
```diff
@@ -99,6 +93,10 @@ const (
 	PageSend = "send"
 )
 
+var (
+	invalidPassphraseError = "error broadcasting transaction: " + dcrlibwallet.ErrInvalidPassphrase
+)
```
oshorefueled: make this a constant

oshorefueled (changes requested):

oshorefueled (approved):

Issue #188, opened 2020-07-17
update: click event on wallet and accounts tabs


oshorefueled:
Currently, wallet and accounts tabs don't respond to click events. This should be fixed to behave just how it was before the update

PR #197, opened 2020-07-18
fix issue with Wallet and Account Tab click event

Sirmorrison:
### Resolves

Issue #188


oshorefueled:
PR #201 resolves the issue

Issue #194, opened 2020-07-17
update: fix modals


oshorefueled:
Currently, modals seem to be below the content of this page. When active, it  should be above all content of a page, not below it. 

#### Additional description
<img width="599" alt="Screenshot 2020-07-17 at 23 39 32" src="https://user-images.githubusercontent.com/15804688/87836174-1c59aa80-c887-11ea-957e-966c06cf0efe.png">

PR #198, opened 2020-07-18
update: fix modals

song50119:
### Resolves

Issue #194

### DIFF
--- a/ui/page.go
+++ b/ui/page.go
@@ -145,6 +145,18 @@ func (page pageCommon) Layout(gtx layout.Context, body layout.Widget) layout.Dim
 	})
 }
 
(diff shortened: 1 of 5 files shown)


ui/page.go
```diff
@@ -145,6 +145,18 @@ func (page pageCommon) Layout(gtx layout.Context, body layout.Widget) layout.Dim
 	})
 }
 
+func (page pageCommon) LayoutWithModal(gtx layout.Context, body layout.Widget, modal layout.Widget) layout.Dimensions {
```
oshorefueled: This can just be named "Modal" and you don't need to specify its second parameter as a widget, it can just be dimension. There's no layout there, all you need are the dimensions.

oshorefueled (changes requested):

oshorefueled (approved):

Issue #190, opened 2020-07-17
update: custom editor widget 


oshorefueled:
After the update, the behaviour of the custom editor changed. In the previous version, when the editor is focused, the label becomes smaller and is pushed upwards. Currently, the editor is pushed down when the editor is focused. The change in scale seems to have also affected the position of the paste icon. That should also be resolved.

#### Additional Description
Gif showing custom editor current behaviour
![editor-bug](https://user-images.githubusercontent.com/15804688/87802989-d9c5ad00-c849-11ea-9fcc-d8b8fef3d744.gif)

PR #199, opened 2020-07-18
fix editor positioning issue

hydra0210:
### Resolves

Issue #190 

### What's new

This fixes the linked issue by ensuring that the text editor widget position remains the same even when focused

### DIFF
--- a/ui/decredmaterial/editor.go
+++ b/ui/decredmaterial/editor.go
@@ -93,6 +93,31 @@ func (t *Theme) Editor(editor *widget.Editor, hint string) Editor {
 	return e
 }
 
+func (e *Editor) calculateDims(gtx layout.Context) {
+	layout.Flex{Axis: layout.Vertical}.Layout(gtx,
+		layout.Rigid(func(gtx C) D {
+			e.titleLabelDims = e.TitleLabel.Layout(gtx)
+			return e.titleLabelDims
+		}),
+		layout.Rigid(func(gtx C) D {
+			e.editorDims = e.EditorStyle.Layout(gtx)
+			return e.editorDims
+		}),
+		layout.Rigid(func(gtx C) D {
+			e.lineDims = layout.Inset{
+				Top:    unit.Dp(4),
+				Bottom: unit.Dp(4),
+			}.Layout(gtx, func(gtx C) D {
+				return e.editorLine(gtx)
+			})
+			return e.lineDims
+		}),
+	)
+}
+
+// Layout renders the editor to screen. The editor line is able to retain
+// it's relative position whether or not the hint or title labels are displayed
+// or not because their dimensions are pre-calculated before hand
 func (e *Editor) Layout(gtx layout.Context) layout.Dimensions {
 	e.handleEvents()
 	if e.IsVisible {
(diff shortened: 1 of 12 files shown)


ui/send_page.go
```diff
@@ -136,11 +136,12 @@ func (win *Window) SendPage(common pageCommon) layout.Widget {
 	page.destinationAddressEditor.SetRequiredErrorText("")
 	page.destinationAddressEditor.IsRequired = true
 	page.destinationAddressEditor.IsVisible = true
+	page.sendAmountEditor.IsTitleLabel = true
 
 	page.sendAmountEditor = common.theme.Editor(new(widget.Editor), "Amount to be sent")
 	page.sendAmountEditor.SetRequiredErrorText("")
 	page.sendAmountEditor.IsRequired = true
-	page.sendAmountEditor.IsTitleLabel = false
+	page.sendAmountEditor.IsTitleLabel = true
```
oshorefueled: Not sure why this is done twice.

ui/decredmaterial/editor.go
```diff
@@ -81,88 +88,112 @@ func (t *Theme) Editor(editor *widget.Editor, hint string) Editor {
...
+		e.Hint = e.StartHint
+	}
+
 	if e.IsRequired && !e.Editor.Focused() && e.Editor.Len() == 0 {
 		e.ErrorLabel.Text = e.requiredErrorText
 	}
 
-	return layout.UniformInset(unit.Dp(2)).Layout(gtx, func(gtx C) D {
-		return layout.Flex{Axis: layout.Vertical}.Layout(gtx,
+	if !e.hasCalculated {
```
oshorefueled: The Layout method seems to do a lot. Add comments to different sections so how it works can be clear.

oshorefueled (changes requested):

Sirmorrison (changes requested):
image

The editor when single line is false has the editor line overlaying on the editor.

ui/decredmaterial/editor.go
```diff
@@ -5,59 +5,69 @@ package decredmaterial
...
 	pasteBtnMaterial IconButton
 	clearBtMaterial  IconButton
+
+	hasCalculated                                   bool
+	editorLineHeight                                int
+	titleLabelDims, editorDims, lineDims, errorDims layout.Dimensions
 }
 
-func (t *Theme) Editor(editor *widget.Editor, hint string) Editor {
+func (t *Theme) Editor(editor *widget.Editor, hint string) *Editor {
```
oshorefueled: this shouldn't return a pointer

ui/decredmaterial/editor.go
```diff
@@ -5,59 +5,69 @@ package decredmaterial
...
 	m.Color = t.Color.Text
-	m.Hint = hint
 	m.HintColor = t.Color.Hint
 
-	return Editor{
+	e := Editor{
 		EditorStyle:       m,
+		shaper:            t.Shaper,
 		TitleLabel:        t.Body2(""),
+		StartHint:         hint,
```
oshorefueled: Any reason why this is made public? Plus, I don't think a "StartHint" field is needed.

ui/decredmaterial/editor.go
```diff
@@ -81,88 +91,128 @@ func (t *Theme) Editor(editor *widget.Editor, hint string) Editor {
...
 	if e.Editor.Focused() || e.Editor.Len() != 0 {
-		e.TitleLabel.Text = e.Hint
-		e.LineColor = color.RGBA{41, 112, 255, 255}
 		e.Hint = ""
 	}
 
+	if !e.Editor.Focused() {
+		e.Hint = e.StartHint
+	}
+
```
oshorefueled: since the Editor method isn't supposed to return a pointer, changing the value of e.Hint when focused would only change it temporarily. This means there wouldn't be a need to keep track of the previous "Hint" with the "StartHint" field. The hint value wouldn't just go back to the hint passed as an argument to the editor when not focused.

oshorefueled (changes requested):
The build for the widget test module should not be part of the PR.

oshorefueled:
PR #209 resolves the issue.

PR #201, opened 2020-07-20
Fix unresponsive tabs on accounts and wallets pages

oshorefueled:
### Resolves

Issue #188 

### What's new

This PR fixes the unresponsiveness of the accounts and wallets tabs when clicked. The SetTab method  was made to only be called when wallet info is updated rather than being set on every frame.


Sirmorrison (approved):
PR Looks good to go.

beansgum (approved):

Issue #216, opened 2020-07-31
send page should scroll upwards after an attempted transaction


oshorefueled:
Currently, when I scroll down the send page, the page remains there after an attempted transaction. This doesn't provide a good user experience as the notification for the attempted transaction, whether a successful or failed transaction is at the top of the page.  

After an attempted transaction, I should be able to see the response from my action. If i'm still at the bottom of the page, the page should be automatically scrolled up. 

#### Additional description
![send-page-bug](https://user-images.githubusercontent.com/15804688/89076452-3e1f5b00-d378-11ea-879b-aa4a0044f18d.gif)

oshorefueled:
issue #227 is a better solution to the problem.

PR #218, opened 2020-08-02
Scroll to top after an attempted transaction

song50119:
### Resolves

Issue #216


oshorefueled:
The send page only scrolls on a failed transaction. Nothing happens when a transaction is successful.

oshorefueled:
an implementation to resolve #227 is preferred.

Issue #222, opened 2020-08-06
rounded bordered editors


oshorefueled:
Currently most editors on godcr have a single bottom border. They should instead be similar to the mobile editors and have 4 borders rounded at the edges. 

![Screenshot 2020-08-06 at 09 35 39](https://user-images.githubusercontent.com/15804688/89510251-39b7cf80-d7c8-11ea-9bf9-54ffd653bca8.png)

* the borders of the editor should be blue when active and red when there's an error such as an invalid input
![Screenshot 2020-08-06 at 09 38 02](https://user-images.githubusercontent.com/15804688/89510500-8e5b4a80-d7c8-11ea-9026-36dd799c3eb6.png)

![Screenshot 2020-08-06 at 09 38 43](https://user-images.githubusercontent.com/15804688/89510581-a8952880-d7c8-11ea-82e1-240e69835098.png)

PR #223, opened 2020-08-06
implement editor with rounded border.

Sirmorrison:
### Resolves

Issue #222 

### What's new

This PR implements custom editor with rounded borders.

image

image

image

### DIFF
--- a/ui/decredmaterial/editor.go
+++ b/ui/decredmaterial/editor.go
@@ -29,6 +29,7 @@ const (
 )
 
 type Editor struct {
+	t *Theme
 	material.EditorStyle
 
 	LineStyle LineStyle
@@ -68,6 +69,7 @@ func (t *Theme) Editor(editor *widget.Editor, hint string) Editor {
 	m.HintColor = t.Color.Hint
 
 	return Editor{
+		t:                 t,
 		EditorStyle:       m,
 		TitleLabel:        t.Body2(""),
 		flexWidth:         0,
@@ -83,7 +85,7 @@ func (t *Theme) Editor(editor *widget.Editor, hint string) Editor {
 				Size:       unit.Dp(25),
 				Background: color.RGBA{},
 				Color:      t.Color.Text,
-				Inset:      layout.UniformInset(unit.Dp(5)),
+				Inset:      layout.UniformInset(unit.Dp(0)),
 				Button:     new(widget.Clickable),
 			},
 		},
@@ -94,7 +96,7 @@ func (t *Theme) Editor(editor *widget.Editor, hint string) Editor {
 				Size:       unit.Dp(25),
 				Background: color.RGBA{},
 				Color:      t.Color.Text,
-				Inset:      layout.UniformInset(unit.Dp(5)),
+				Inset:      layout.UniformInset(unit.Dp(0)),
 				Button:     new(widget.Clickable),
 			},
 		},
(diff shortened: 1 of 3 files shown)


ghost:
The rounded border shown in the screenshots should have a slightly bigger radius on the inner background box under the text, it looks almost square compared to the outer.

ui/decredmaterial/editor.go
```diff
@@ -173,27 +147,78 @@ func (e Editor) Layout(gtx layout.Context) layout.Dimensions {
...
+	var dims layout.Dimensions
+	switch e.LineStyle {
+	case RoundedRectangle:
+		dims = e.editorRectangle(gtx, func(gtx C) D {
+			return e.editorSection(gtx, false)
+		})
+	case SingleUnderLine:
+		dims = e.editorSection(gtx, true)
+	case NoLine:
+		dims = e.editorSection(gtx, false)
```
oshorefueled: Do we really need to have 3 editor layouts? The entire app would be using the editor with rounded borders. I'm not sure where we might need a bare editor with no border but a simple if statement with a property in the Editor struct should be able to resolve that.

ui/decredmaterial/editor.go
```diff
@@ -173,27 +147,78 @@ func (e Editor) Layout(gtx layout.Context) layout.Dimensions {
...
+			return layout.Dimensions{}
 		}),
 	)
 }
 
+func (e Editor) editorRectangle(gtx layout.Context, body layout.Widget) layout.Dimensions {
+	border := widget.Border{Color: e.LineColor, CornerRadius: values.MarginPadding5, Width: values.MarginPadding1}
+	return border.Layout(gtx, func(gtx C) D {
+		mtb := values.MarginPadding2
+		mlr := values.MarginPadding5
```
oshorefueled: you can define the margins and paddings within the widgets. decredmaterial widgets should be independent of variables and definitions outside the decredmaterial package.

oshorefueled (changes requested):

ui/decredmaterial/editor.go
Sirmorrison: The bare editor is what the send page amount editor uses.

ui/decredmaterial/editor.go
Sirmorrison: So basically we cannot use the dimensions values declared already.

oshorefueled: not in decredmaterial

ui/decredmaterial/editor.go
```diff
@@ -173,27 +149,56 @@ func (e Editor) Layout(gtx layout.Context) layout.Dimensions {
...
 		}),
 	)
 }
 
+func (e Editor) editorRectangle(gtx layout.Context, body layout.Widget) layout.Dimensions {
+	border := widget.Border{Color: e.LineColor, CornerRadius: e.m5, Width: unit.Dp(1)}
+	return border.Layout(gtx, func(gtx C) D {
+		return layout.Inset{Top: e.m2, Bottom: e.m2, Left: e.m5, Right: e.m5}.Layout(gtx, body)
+	})
+}
```
oshorefueled: is this method really needed? It seems to be used just once and the rest of the code is short enough to fit into the editorLayout method.

ui/decredmaterial/editor.go
```diff
@@ -173,27 +149,56 @@ func (e Editor) Layout(gtx layout.Context) layout.Dimensions {
...
+	if e.Bordered {
+		return e.editorRectangle(gtx, func(gtx C) D {
+			return e.editorSection(gtx)
+		})
+	}
+
+	return e.editorSection(gtx)
+}
+
+func (e Editor) editorSection(gtx layout.Context) layout.Dimensions {
```
oshorefueled: the entire code results in the editor widget. Having a method named "editorSection" is a little confusing.

oshorefueled (changes requested):

ui/decredmaterial/editor.go
Sirmorrison: Since i will be wrapping the editor in the border widget, the editorSection function is necessary. I will rename this to editor for simplicity.

oshorefueled (approved):

Issue #233, opened 2020-08-10
Shutdown/cleanup and interrupt handling


ghost:
Currently when the close button is clicked there is a remaining dangling thread that is immediately stopped by a sigstop (ctrl-c) and if this signal is sent from the controlling shell the entire application stops. 

I propose to use a library that I recently wrote: https://gitlab.com/stalker.loki/disrupt/ which I *think* (now to be honest I'm not even sure, it was so long ago) is derived from the Neutrino SPV client's interrupt handling code but additionally includes the ability to trigger a restart that I have used both in a controlling RPC as well as with a log monitor app that I was part way through writing for https://github.com/p9c/pod

oshorefueled:
thanks for the suggestion but #234 solves that.

PR #234, opened 2020-08-10
fix incomplete GUI triggered shutdown

oshorefueled:
### Resolves

Issue #233 

### What's new

This PR completely shuts down all process when Godcr is closed from the GUI window.


Sirmorrison (approved):
Tested and pr is good to go.

beansgum (approved):

PR #252, opened 2020-08-28
Replace clipboard library by Gio system clipboard

### LINKED ISSUE #251: replace external clipboard library with Gio system clipboard
Gio has its own clipboard library and is preferred over the currently used clipboard library (https://github.com/atotto/clipboard) as it compatible across more platforms.

song50119:
Resolves #251

### DIFF
--- a/ui/decredmaterial/editor.go
+++ b/ui/decredmaterial/editor.go
@@ -10,7 +10,6 @@ import (
 	"gioui.org/widget"
 	"gioui.org/widget/material"
 
-	"github.com/atotto/clipboard"
 	"golang.org/x/exp/shiny/materialdesign/icons"
 )
 
@@ -202,13 +201,19 @@ func (e Editor) editor(gtx layout.Context) layout.Dimensions {
 }
 
 func (e Editor) handleEvents() {
-	for e.pasteBtnMaterial.Button.Clicked() {
-		data, err := clipboard.ReadAll()
-		if err != nil {
-			panic(err)
-		}
-		e.Editor.SetText(data)
+	if e.pasteBtnMaterial.Button.Clicked() {
+		e.Editor.Focus()
+
+		go func() {
+			text := <-e.t.Clipboard
+			e.Editor.SetText(text)
+			e.Editor.Move(e.Editor.Len())
+		}()
+		go func() {
+			e.t.ReadClipboard <- ReadClipboard{}
+		}()
 	}
+
 	for e.clearBtMaterial.Button.Clicked() {
 		e.Editor.SetText("")
 	}
--- a/ui/window.go
+++ b/ui/window.go
@@ -49,6 +49,11 @@ type Window struct {
 	pages                   map[string]layout.Widget
 	walletTabs, accountTabs *decredmaterial.Tabs
 	keyEvents               chan *key.Event
+	clipboard               chan interface{}
+}
+
+type WriteClipboard struct {
+	Text string
 }
 
 // CreateWindow creates and initializes a new window with start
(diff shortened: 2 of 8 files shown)


ui/window.go
```diff
@@ -49,6 +49,9 @@ type Window struct {
 	pages                   map[string]layout.Widget
 	walletTabs, accountTabs *decredmaterial.Tabs
 	keyEvents               chan *key.Event
+	clipboardIsPasted       chan bool
```
oshorefueled: I don't think you need 3 channels to handle all clipboard events. A single channel should be sufficient. You can have one channel and create 3 different types, one to read from the clipboard and the other to write to it. The later would have a  text field to be written to the clipboard.

ui/decredmaterial/editor.go
```diff
@@ -203,11 +203,12 @@ func (e Editor) editor(gtx layout.Context) layout.Dimensions {
...
-		data, err := clipboard.ReadAll()
-		if err != nil {
-			panic(err)
-		}
-		e.Editor.SetText(data)
+		go func() {
+			e.Editor.Focus()
+			time.AfterFunc(time.Millisecond*60, func() {
+				e.t.Clipboard.IsPasted <- true
+			})
```
oshorefueled: I'm not sure why a delay of 60 milliseconds

ui/window.go
```diff
@@ -159,7 +170,10 @@ func (win *Window) Loop(shutdown chan int) {
 			case wallet.BlockConfirmed:
 				win.updateSyncProgress(update.ConfirmedTxn)
 			}
-
+		case <-win.clipboardIsPasted:
+			win.window.ReadClipboard()
```
oshorefueled: When the clipboard channel receives the signal of the read type, the window.ReadClipBoard() method should be called. And when it receives a write type, the window.WriteClipBoard should be called with the text received on the channel.

ui/window.go
```diff
@@ -185,6 +199,10 @@ func (win *Window) Loop(shutdown chan int) {
 				go func() {
 					win.keyEvents <- &evt
 				}()
+			case system.ClipboardEvent:
+				go func() {
+					win.clipboardPasteText <- evt.Text
```
oshorefueled: When this event is triggered by calling window.ReadClipBoard, the text should be sent back to the same channel with the third type. It could be named "Paste". All Editors would listen to the same channel and they would receive the signal when it is sent here. But only the focused editor would have the text pasted in its field.

oshorefueled (changes requested):

ui/window.go
song50119: If single channel, the loop in window will read the value first, then there is nothing on specific pages

oshorefueled: you're correct,  one channel would not suffice. Two would work though. One in the loop to call the read and write methods when it receives a signal from the editor. And another for the editors, only listened when an editor in focus.

ui/decredmaterial/editor.go
```diff
@@ -203,11 +202,9 @@ func (e Editor) editor(gtx layout.Context) layout.Dimensions {
...
 func (e Editor) handleEvents() {
 	for e.pasteBtnMaterial.Button.Clicked() {
-		data, err := clipboard.ReadAll()
-		if err != nil {
-			panic(err)
-		}
-		e.Editor.SetText(data)
+		go func() {
+			e.t.IsEditorPasted <- true
+		}()
```
oshorefueled: change this to an "if". Clicked returns a bool

oshorefueled (approved):

PR #271, opened 2020-12-02
Implement side nav in tab widget

hydra0210:
This moves the side navigation bar code to the tab widget


oshorefueled:
The tab widget is still in use by the send and receive page that haven't got a mockup designed for them yet. This would instead be created as a nav widget rather than replace the tab widget.

PR #275, opened 2020-12-07
One menu on the transactions page should be closed when the other is active

### LINKED ISSUE #272: one menu on the transactions page should be closed when the other is active
I can currently use two of the menus below at the same time. And since I currently can't click on the page to close the menus, one of them should be closed when the other is open. 

Screenshot 2020-12-03 at 14 33 44

song50119:
Resolves #272 
Hide all opened menu before show new

### DIFF
--- a/ui/decredmaterial/theme.go
+++ b/ui/decredmaterial/theme.go
@@ -74,6 +74,8 @@ type Theme struct {
 
 	Clipboard     chan string
 	ReadClipboard chan interface{}
+
+	dropDownMenus []*DropDown
 }
 
 func NewTheme(fontCollection []text.FontFace) *Theme {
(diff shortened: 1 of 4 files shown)


ui/decredmaterial/theme.go
```diff
@@ -75,6 +75,8 @@ type Theme struct {
 
 	Clipboard     chan string
 	ReadClipboard chan interface{}
+
+	dropDownMenus []*DropDown
```
oshorefueled, line 78: I don't think closing drop down menus should be done through out the entire app. The dropdowns that are related can be in a group and only the dropdowns in the group will be closed when one of the items is selected.

ui/decredmaterial/theme.go
```diff
@@ -224,3 +226,9 @@ func mulAlpha(c color.RGBA, alpha uint8) color.RGBA {
 		B: uint8(uint16(c.B) * a / 255),
 	}
 }
+
+func (t *Theme) HideAllDropdownMenus() {
```
oshorefueled: A better name for this would probably be "closeAllDropdownMenus"

oshorefueled (changes requested):

oshorefueled (approved):

Issue #281, opened 2020-12-23
implement new mockup design on verify message page


oshorefueled:
update the verify message page to the design as seen in the screenshot below 

<img width="694" alt="Screenshot 2020-12-23 at 08 45 04" src="https://user-images.githubusercontent.com/15804688/102972435-56e10680-44fb-11eb-9c9b-cc6cc06e6420.png">

PR #286, opened 2020-12-28
update verify message page to new UI design

oshorefueled:
This PR resolves issue #281

### DIFF
--- a/ui/modal_templates.go
+++ b/ui/modal_templates.go
@@ -15,8 +15,10 @@ const RenameAccountTemplate = "RenameAccount"
 const PasswordTemplate = "Password"
 const ChangePasswordTemplate = "ChangePassword"
 const ConfirmRemoveTemplate = "ConfirmRemove"
+const InfoTemplate = "Info"
 
 type ModalTemplate struct {
+	th                    *decredmaterial.Theme
 	walletName            decredmaterial.Editor
 	spendingPassword      decredmaterial.Editor
 	matchSpendingPassword decredmaterial.Editor
@@ -43,6 +45,7 @@ func (win *Window) LoadModalTemplates() *ModalTemplate {
 	icon.Background = win.theme.Color.Surface
 
 	return &ModalTemplate{
+		th:                    win.theme,
 		confirm:               win.theme.Button(new(widget.Clickable), "Confirm"),
 		cancel:                win.theme.Button(new(widget.Clickable), "Cancel"),
 		walletName:            win.theme.Editor(new(widget.Editor), ""),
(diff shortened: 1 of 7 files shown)


ui/page.go
```diff
@@ -630,6 +639,72 @@ func (page pageCommon) SelectedAccountLayout(gtx layout.Context) layout.Dimensio
...
+				cancelText: "Got it",
+			}
+		}()
+	}
+
+	if page.subPageBackButton.Button.Clicked() {
+		sp.back()
+	}
+
+	return decredmaterial.Card{Color: page.theme.Color.Surface, Rounded: true}.Layout(gtx, func(gtx C) D {
```
Sirmorrison: kindly rebase as the Card widget has been updated. `Color` and `Rounded` params no longer exist in card struct.

Sirmorrison (changes requested):
* A little extra padding around the modal will be nice.
image

The valid and invalid text colors does not align with the mockup design.
image

* The gray text background is way darker than the UI design. you can use the gray color used on the transaction details page.
image

tx details page..
image

* The color used for the clear all button is gray as againt the the blue used here.
image

image

ui/page.go
```diff
@@ -669,6 +678,72 @@ func (page pageCommon) SelectedAccountLayout(gtx layout.Context) layout.Dimensio
...
+	if page.subPageInfoButton.Button.Clicked() {
+		go func() {
+			page.modalReceiver <- &modalLoad{
+				template:   InfoTemplate,
+				title:      "Verify message",
+				cancel:     page.closeModal,
+				cancelText: "Got it",
+			}
+		}()
+	}
```
Sirmorrison: From what i understand about this function, its supposed to used by all the subpages. However whenever the `subPageInfoButton` info button is clicked, it will only render the verify message modal template. I believe this should be made flexible.

Sirmorrison (changes requested):

ui/modal_templates.go
```diff
@@ -275,6 +296,12 @@ func (m *ModalTemplate) handle(th *decredmaterial.Theme, load *modalLoad) (templ
 		}
 		template = m.removeWallet(th)
 		return
+	case InfoTemplate:
+		if m.cancel.Button.Clicked() {
+			load.cancel.(func())()
+		}
+		template = m.verifyMessageInfo()
+		return
```
Sirmorrison: This `InfoTemplate` is supposed to render for all the information on the pages i presume. currently, this renders only the `verifyMessageInfo`. We can either make this dynamic so it works with all info on the pages or have the `InfoTemplate` renamed to `verifyMsgInfoTemplate`.

oshorefueled: you're correct. My intention was to do that before pushing, I'll do that on my next commit.

ui/modal_templates.go
```diff
@@ -146,6 +149,18 @@ func (m *ModalTemplate) changePassword() []func(gtx C) D {
...
 }
 
+func (m *ModalTemplate) verifyMessageInfo() []func(gtx C) D {
+	return []func(gtx C) D{
+		func(gtx C) D {
+			text := m.th.Body1("After you or your counterparty has genrated a signature, you can use this form to verify the" +
+				" validity of the  signature. \n \nOnce you have entered the address, the message and the corresponding " +
+				"signature, you will see VALID if the signature appropriately matches the address and message, otherwise INVALID.")
+			text.Color = m.th.Color.Gray
+			return text.Layout(gtx)
```
Sirmorrison, line 159:
```suggestion
func (m *ModalTemplate) verifyMessageInfo(infoText) []func(gtx C) D {
	return []func(gtx C) D{
		func(gtx C) D {
			text := m.th.Body1(infoText)
			text.Color = m.th.Color.Gray
			return text.Layout(gtx)
```

Rather than we declare multiple info modals, we can have one model that takes the info text as a parameter and can be used by all subpages with the info option.

Sirmorrison:
image
the address is correct but the invalid error message is still displayed.

Sirmorrison (approved):
PR looks good.

beansgum (approved):

PR #305, opened 2021-01-22
Implement godcr settings page

### LINKED ISSUE #295: settings page
Implement settings page as in the mockup below
Screenshot 2021-01-11 at 14 45 09

Sirmorrison:
Resolves #295 

This PR adds the settings page ui and implements the wallet startup password.


ui/privacy_page.go
```diff
@@ -50,14 +50,15 @@ func (win *Window) PrivacyPage(common pageCommon) layout.Widget {
 func (pg *privacyPage) Layout(gtx layout.Context, c pageCommon) layout.Dimensions {
 	d := func(gtx C) D {
 		load := SubPage{
-			title:      "Privacy",
-			walletName: c.info.Wallets[*c.selectedWallet].Name,
+			title:        "Privacy",
+			walletName:   c.info.Wallets[*c.selectedWallet].Name,
+			isInfoButton: true,
+			isWalletName: true,
```
oshorefueled: Is this necessary? why not check if the walletName is an empty string?. Same thing applies the infoTemplate.

ui/settings_page.go
```diff
@@ -2,36 +2,357 @@ package ui
...
+	wal           *wallet.Wallet
+
+	currencyConversion decredmaterial.IconButton
+	connectToPeer      decredmaterial.IconButton
+	userAgent          decredmaterial.IconButton
+	changeStartupPass  decredmaterial.IconButton
+
+	spendUnconfirm  *widget.Bool
+	startupPassword *widget.Bool
+	notificationW   *widget.Bool
```
oshorefueled, line 27: There doesn't seem to be any other field named "notification". This is fine as just "notification"

ui/settings_page.go
```diff
@@ -2,36 +2,357 @@ package ui
...
-// main settings layout
 func (pg *settingsPage) Layout(gtx layout.Context, common pageCommon) layout.Dimensions {
-	return common.Layout(gtx, func(gtx C) D {
-		return layout.Center.Layout(gtx, func(gtx C) D {
-			return pg.dummyText.Layout(gtx)
+	pg.updateSetting()
+	body := func(gtx C) D {
+		page := SubPage{
+			title:      "Settings",
+			walletName: "",
```
oshorefueled: You don't need to pass an empty string to the field. You can simply leave the field out

ui/settings_page.go
```diff
@@ -2,36 +2,357 @@ package ui
...
+					pg.security(),
+					pg.notification(),
+					pg.connection(),
+				}
+
+				return pg.pageContainer.Layout(gtx, len(pageContent), func(gtx C, i int) D {
+					return layout.Inset{}.Layout(gtx, pageContent[i])
+				})
+			},
+			infoTemplate: "",
```
oshorefueled: You can leave the field out, it's the same as an empty string

ui/settings_page.go
```diff
@@ -2,36 +2,357 @@ package ui
...
+			e := "Password is incorrect"
+			common.Notify(e, false)
+			return
+		}
+		common.Notify(err.Error(), false)
+	default:
+	}
+}
+
+func (pg *settingsPage) updateSetting() {
```
oshorefueled: The method name should specify what setting it is updating

oshorefueled (changes requested):

ui/settings_page.go
Sirmorrison: There is a `notification()`

ui/privacy_page.go
Sirmorrison: Removing `isInfoButton` and `isWalletName` and doing this https://github.com/planetdecred/godcr/pull/305#discussion_r564375795 and https://github.com/planetdecred/godcr/pull/305#discussion_r564375852 leaves the info icon and a piece of the card. 

image

ui/page.go
```diff
@@ -722,21 +728,24 @@ func (page pageCommon) subpageHeader(gtx layout.Context, sp SubPage) layout.Dime
...
 		layout.Rigid(func(gtx layout.Context) layout.Dimensions {
-			return layout.Inset{Left: values.MarginPadding5, Top: values.MarginPadding5}.Layout(gtx, func(gtx C) D {
-				return decredmaterial.Card{
-					Color: page.theme.Color.Surface,
-				}.Layout(gtx, func(gtx C) D {
-					return layout.UniformInset(values.MarginPadding2).Layout(gtx, func(gtx C) D {
-						walletText := page.theme.Caption(sp.walletName)
-						walletText.Color = page.theme.Color.Gray
-						return walletText.Layout(gtx)
+			if sp.walletName == " " {
```
oshorefueled: This is a space and not an empty string. You should be checking if the walletName string is not empty.

ui/page.go
```diff
@@ -722,21 +728,24 @@ func (page pageCommon) subpageHeader(gtx layout.Context, sp SubPage) layout.Dime
...
 					})
 				})
-			})
+			}
+			return layout.Dimensions{}
 		}),
 		layout.Flexed(1, func(gtx layout.Context) layout.Dimensions {
 			return layout.E.Layout(gtx, func(gtx C) D {
-				if sp.isInfoButton {
+				if sp.infoTemplate == " " {
```
oshorefueled: The if check doesn't seem to be correct.

oshorefueled (changes requested):

ui/page.go
Sirmorrison: Apparently, testing with an empty string did not yield the desired result as the info icon and the `walletName` card was still displayed.
image
 
 i also just noticed that the info icon and the wallet name was removed from all the other pages that had them when the space was added.

Explicitly defining `walletName : ""` and `infoTemplate : ""` did not resolve it either.
What seems to fix the issue are the `isWalletName` and `isInfoTemplate` variables that were initially added.

oshorefueled (approved):

PR #311, opened 2021-02-04
make entire option rows clickable in settings page

### LINKED ISSUE #310: make entire option row clickable
Below is the screenshot of the settings page 
Screenshot 2021-02-04 at 11 53 00

Currently, to use the option row, the right navigation icon has to be clicked. A user should instead be able to click the entire row to use the option.

dreacot:
Resolves https://github.com/planetdecred/godcr/issues/310

This PR makes the entire option rows in the settings page clickable to enable the users to easily use/access the option


go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/decred/slog v1.1.0
 	github.com/jessevdk/go-flags v1.4.1-0.20200711081900-c17162fe8fd7
 	github.com/jrick/logrotate v1.0.0
+	github.com/markbates/pkger v0.17.1
```
oshorefueled: pkger doesn't seem to be used in this PR

ui/settings_page.go
```diff
@@ -12,13 +12,18 @@ import (
 
 const PageSettings = "Settings"
 
+type optionRowHandler struct {
```
oshorefueled: This isn't a handler. It should probably be just "optionRow"

go.sum
```diff
@@ -1,25 +1,35 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+decred.org/cspp v0.2.0 h1:SdwdoGT2wZenkczeDxzcKwoAA55Y0Ti3aZslabBORvA=
 decred.org/cspp v0.2.0/go.mod h1:KVnB49sueBFCldRa/ivZCaWZbrPNEiXWwxHCf1jTYKI=
 decred.org/cspp v0.3.0 h1:2AkSsWzA7HIMZImfw0gT82Gdp8OXIM4NsBn7vna22uE=
 decred.org/cspp v0.3.0/go.mod h1:UygjYilC94dER3BEU65Zzyoqy9ngJfWCD2rdJqvUs2A=
 decred.org/dcrwallet v1.6.0-rc4 h1:5IT6mFa+2YMqenu6aE2LetD0N8QSUVFyAFl205PvIIE=
 decred.org/dcrwallet v1.6.0-rc4/go.mod h1:lsrNbuKxkPGeHXPufxNTckwQopCEDz0r3t0a8JCKAmU=
 dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7gZCb22OMCxBHrMx7a5I7Hp++hsVxbQ4BYO7hU=
+gioui.org v0.0.0-20200722191435-e381ff40d66b h1:gnY/k6FSe0jYKiVFDG0ZSnBRry1zgh5c841NFifuQTA=
```
oshorefueled: The PR shouldn't make changes to the go.mod or go.sum files.

ui/settings_page.go
```diff
@@ -252,7 +276,7 @@ func (pg *settingsPage) agent() layout.Widget {
 func (pg *settingsPage) mainSection(gtx layout.Context, title string, body layout.Widget) layout.Dimensions {
 	return layout.Inset{Bottom: values.MarginPadding10}.Layout(gtx, func(gtx C) D {
 		return pg.theme.Card().Layout(gtx, func(gtx C) D {
-			return layout.UniformInset(values.MarginPadding15).Layout(gtx, func(gtx C) D {
+			return layout.Inset{Top: values.MarginPadding15, Left: values.MarginPadding15, Right: values.MarginPadding15}.Layout(gtx, func(gtx C) D {
```
oshorefueled: MarginPadding15 can be assigned to a variable to break the code to the next line.

ui/settings_page.go
```diff
@@ -123,15 +132,28 @@ func (pg *settingsPage) general() layout.Widget {
...
-								txt.Color = pg.theme.Color.Gray
-								return txt.Layout(gtx)
-							}),
-							layout.Rigid(func(gtx C) D {
-								return pg.currencyConversion.Layout(gtx)
+					return pg.theme.Card().Layout(gtx, func(gtx C) D {
+						return layout.Stack{}.Layout(gtx,
+							layout.Stacked(func(gtx C) D {
+								list := layout.List{Axis: layout.Vertical}
+								return list.Layout(gtx, 1, func(gtx C, i int) D {
```
oshorefueled: I'm not sure why a list layout is needed here.

oshorefueled (changes requested):

oshorefueled:
The entire row click doesn't seem to be applied to the change specific peer row

ui/settings_page.go
dreacot: i'm not sure what you mean here?

ui/settings_page.go
```diff
@@ -12,15 +12,20 @@ import (
 
 const PageSettings = "Settings"
 
+type optionRow struct {
+	clickable *widget.Clickable
+	icon      decredmaterial.IconButton
```
oshorefueled: Having an clickable widget and an icon button do the same thing seems redundant. The clickable widget and an icon might just be all that's needed. Reassess the code and see if you really need to create a new struct type for this.

ui/settings_page.go
```diff
@@ -123,17 +140,21 @@ func (pg *settingsPage) general() layout.Widget {
...
+										txt := pg.theme.Body2("None")
+										txt.Color = pg.theme.Color.Gray
+										return txt.Layout(gtx)
+									}),
+									layout.Rigid(func(gtx C) D {
+										return pg.currencyConversion.icon.Layout(gtx)
+									}),
+								)
+							})
+						})
```
oshorefueled: This code block seems to be reused for a number of option rows, you can consider making it a function or method and reusing that.

ui/settings_page.go
```diff
@@ -404,7 +438,7 @@ func (pg *settingsPage) handle(common pageCommon) {
 	}
 
 	userAgentKey := dcrlibwallet.UserAgentConfigKey
-	for pg.updateUserAgent.Button.Clicked() {
+	for pg.updateUserAgent.clickable.Clicked() || pg.updateUserAgent.icon.Button.Clicked() {
```
oshorefueled: Removing the icon button and using just an icon would remove the need for this or condition.

oshorefueled (changes requested):

ui/settings_page.go
```diff
@@ -123,17 +128,21 @@ func (pg *settingsPage) general() layout.Widget {
...
-								txt.Color = pg.theme.Color.Gray
-								return txt.Layout(gtx)
-							}),
-							layout.Rigid(func(gtx C) D {
-								return pg.currencyConversion.Layout(gtx)
-							}),
-						)
+					return decredmaterial.Clickable(gtx, pg.currencyConversion, func(gtx C) D {
+						return layout.Inset{Top: values.MarginPadding5, Bottom: values.MarginPaddingMinus5}.Layout(gtx, func(gtx C) D {
+							return pg.subSection(gtx, "Currency conversion", func(gtx C) D {
```
oshorefueled: why wasn't the clickableRow method used for currency conversion. It looks like it would work here.

ui/settings_page.go
```diff
@@ -283,6 +280,23 @@ func (pg *settingsPage) subSectionSwitch(gtx layout.Context, title string, optio
 	})
 }
 
+func (pg *settingsPage) clickableRow(gtx layout.Context, title string, clickable *widget.Clickable, icon *widget.Icon, label decredmaterial.Label) layout.Dimensions {
```
oshorefueled: If the arguments you'll be passing to a method is becoming too many, you can create a struct type and pass  that instead. For instance, for clickableRow, you can define the method to accept layout.Context and a type "row" as arguments. The type "row" would contain the remaining the parameters you defined on the method.

ui/settings_page.go
```diff
@@ -283,6 +280,23 @@ func (pg *settingsPage) subSectionSwitch(gtx layout.Context, title string, optio
...
+					layout.Rigid(func(gtx C) D {
+						return icon.Layout(gtx, values.MarginPadding22)
+					}),
+				)
+			})
+		})
+	})
+}
+
 func (pg *settingsPage) conditionalDisplay(gtx layout.Context, body layout.Widget) layout.Dimensions {
```
oshorefueled: Kindly add a boolean as one of the parameters of this method so it can accept the condition as an argument. With the "if" statement  inside the method.

oshorefueled (changes requested):

ui/settings_page.go
```diff
@@ -54,14 +60,15 @@ func (win *Window) SettingsPage(common pageCommon) layout.Widget {
...
 		line:     common.theme.Line(),
 		errChann: common.errorChannels[PageSettings],
 
-		currencyConversion:  common.theme.PlainIconButton(new(widget.Clickable), icon),
-		updateConnectToPeer: common.theme.PlainIconButton(new(widget.Clickable), icon),
-		updateUserAgent:     common.theme.PlainIconButton(new(widget.Clickable), icon),
-		changeStartupPass:   common.theme.PlainIconButton(new(widget.Clickable), icon),
+		currencyConversion:  currencyConversion,
+		updateConnectToPeer: updateConnectToPeer,
+		updateUserAgent:     updateUserAgent,
```
oshorefueled: You don't need to create variables for the Clickable types, you can just use new(widget.Cickable) here.

oshorefueled (changes requested):

ui/settings_page.go
dreacot: Yes it would, but the functionality hasn't been implemented yet, once the functionality is implemented, we can then pass in the required label as the parameter

Else I'd have to put an if else statement in the clickableRow function to accommodate the currency conversion

ui/settings_page.go
dreacot: I'm not sure I fully get what you mean here

oshorefueled: something like this 
`
func (pg *settingsPage) conditionalDisplay(gtx layout.Context, display bool, body layout.Widget) layout.Dimensions { 
  if display { 
    return layout.Flex{Axis: layout.Vertical}.Layout(gtx, 
      layout.Rigid(
       pg.lineSeparator()), 
       layout.Rigid(body), 
      ) 
  } 
  return layout.Dimensions{}
}
`

oshorefueled (approved):

PR #315, opened 2021-02-09
Implement validate address page and verify message page

### LINKED ISSUE #309: security tools: validate address and verify message
The security tools page was added in #299. 
* Verify Message already exists as a wallet feature. The security tools page simply needs to navigate to the page, the only difference is the wallet tag shouldn't be displayed when the page is navigated from the security tools page. 
* The validate address feature should be added as a security tool. It should be implemented as in the mockup below. 

Screenshot 2021-02-04 at 10 53 36

Screenshot 2021-02-04 at 10 52 52

Screenshot 2021-02-04 at 10 53 20

JustinBeBoy:
Resolves #309


oshorefueled:
There shouldn't be a "wallet" tag on the verify message page when navigating from the security tools page. 

Screenshot 2021-02-22 at 09 27 26

ui/page.go
```diff
@@ -270,12 +276,23 @@ func (win *Window) addPages(decredIcons map[string]image.Image) {
...
 	page.changePage(pg)
 }
 
+func (page pageCommon) PushNavigationPage(from, to string) {
+	page.returnPage = &from
+	page.pushNavigationPage(from)
+	page.changePage(to)
+}
+
+func (page pageCommon) PopNavigationPage() {
```
oshorefueled: This method name is a bit misleading. Pop and push are commonly used for removing or adding data to an array, this method has no association to an array. You're simply setting values of the pages to a field in the common struct.

oshorefueled (changes requested):

ui/validate_address.go
```diff
@@ -0,0 +1,288 @@
...
+
+	"gioui.org/layout"
+	"gioui.org/widget"
+
+	"github.com/planetdecred/godcr/ui/decredmaterial"
+	"github.com/planetdecred/godcr/ui/values"
+	"github.com/planetdecred/godcr/wallet"
+)
+
+const ValidateAddress = "Validate Address"
```
oshorefueled: Remove the space in the  page ID

ui/validate_address.go
```diff
@@ -0,0 +1,288 @@
...
+
+const (
+	none     = 0
```
oshorefueled: you can use iota to assign successive integer constants

ui/window.go
```diff
@@ -115,6 +115,11 @@ func (win *Window) changePage(page string) {
 	win.window.Invalidate()
 }
 
+func (win *Window) pushNavigationPage(from string) {
```
oshorefueled: "pushNavigationPage" is a bit misleading as push and pop are words associated with arrays. In this case you're simply setting a value. It can be "setReturnPage"

oshorefueled (changes requested):

ui/page.go
```diff
@@ -32,7 +32,7 @@ type pageIcons struct {
 	pendingIcon, logo, redirectIcon, confirmIcon, newWalletIcon, walletAlertIcon,
 	importedAccountIcon, accountIcon, editIcon, expandIcon, collapseIcon, copyIcon, mixer,
 	arrowFowardIcon, transactionFingerPrintIcon, settingsIcon, securityIcon, helpIcon,
-	aboutIcon, debugIcon, alert, verifyMessageIcon, locationPinIcon, alertGray, arrowDownIcon *widget.Image
+	aboutIcon, debugIcon, alert, verifyMessageIcon, locationPinIcon, alertGray, arrowDownIcon, checkMarkGreenIcon, crossMarkRed *widget.Image
```
oshorefueled: This should be broken into the next line

oshorefueled (changes requested):

oshorefueled (dismissed):

oshorefueled (approved):

PR #318, opened 2021-02-15
Import watch only wallet

### LINKED ISSUE #291: import watch-only wallet
Add the import watch-only wallet feature. 
NOTE: Issue #289 should be implemented before this is done

Screenshot 2021-01-05 at 17 06 33

Screenshot 2021-01-05 at 17 10 04

Screenshot 2021-01-05 at 17 10 18

hydra0210:
This adds the import watch only wallet feature. 
Fixes #291


ui/wallet_page.go
```diff
@@ -50,26 +60,42 @@ type walletPage struct {
 	optionsMenuCard                            decredmaterial.Card
 	optionsMenuItems                           []optionMenuItem
 	openPopupIndex                             int
+	addWalletMenuOptions                       []addWalletMenuOption
+	openAddWalletPopupButton                   *widget.Clickable
+	isAddWalletMenuOpen                        bool
+	watchOnlyWalletLabel                       decredmaterial.Label
+	noWatchOnlyWalletLabel                     decredmaterial.Label
```
oshorefueled: This doesn't seem to be used anywhere.

ui/wallet_page.go
```diff
@@ -28,6 +28,18 @@ type optionMenuItem struct {
 	page   string
 }
 
+type addWalletMenuOption struct {
```
oshorefueled: This looks like it does the same thing with optionMenuItem, their only difference being action and page. The page field is used to set the page to change to when one of the optionMenuItems is clicked. The change page action can instead be done in action.

ui/wallet_page.go
```diff
@@ -145,20 +186,40 @@ func (pg *walletPage) Layout(gtx layout.Context, common pageCommon) layout.Dimen
...
-		backupBtn.Inset = layout.UniformInset(values.MarginPadding0)
-		backupBtn.Size = values.MarginPadding20
-		pg.backupButton = append(pg.backupButton, backupBtn)
+		if common.info.Wallets[index].IsWatchingOnly {
+			if _, ok := pg.watchOnlyWalletMoreButtons[index]; !ok {
+				pg.watchOnlyWalletMoreButtons[index] = decredmaterial.IconButton{
+					IconButtonStyle: material.IconButtonStyle{
+						Button:     new(widget.Clickable),
+						Icon:       common.theme.NavMoreIcon,
+						Size:       unit.Dp(25),
```
oshorefueled: use a defined dimension from the values package

oshorefueled (changes requested):

go.sum
```diff
@@ -152,6 +152,7 @@ github.com/envoyproxy/go-control-plane v0.9.0/go.mod h1:YTl/9mNaCwkRvm6d1a2C3ymF
 github.com/envoyproxy/go-control-plane v0.9.4/go.mod h1:6rpuAdCZL397s3pYoYcLgu1mIlRU8Am5FuJP05cCM98=
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
 github.com/fsnotify/fsnotify v1.4.7/go.mod h1:jwhsz4b93w/PPRr/qN1Yymfu8t87LnFCMoQvtojpjFo=
+github.com/fsnotify/fsnotify v1.4.9 h1:hsms1Qyu0jgnwNXIxa+/V/PDsU6CfLf6CNO8H7IWoS4=
```
oshorefueled: This doesn't seem like it should be here since there's no change to the go.mod file

oshorefueled (changes requested):

oshorefueled:
UI BUGS
* The extended key field doesn't get cleared after successfully importing a wallet. 
* After importing a watch-only wallet, the sync progress never reaches the "synced" state. I observed though that It gets to "Synced" after I navigate to the send page and I get a "Loading" screen. Below is a screenshot of the overview page not displaying a synced state.
Screenshot 2021-02-16 at 16 41 10

oshorefueled:
Watch only wallet should not be displayed when there's no wallet imported. Below is the current implementation
Screenshot 2021-02-24 at 18 11 06

Below is how it should be
Screenshot 2021-02-24 at 18 10 59

ui/wallet_page.go
```diff
@@ -22,10 +22,17 @@ import (
 
 const PageWallet = "Wallet"
 
-type optionMenuItem struct {
+type menuItem struct {
 	text   string
-	button *widget.Clickable
 	page   string
+	button *widget.Clickable
+	action func(pageCommon)
```
oshorefueled, line 28: Is the action field needed? You could execute the switch to page within the action.

oshorefueled (changes requested):

oshorefueled (approved):

PR #323, opened 2021-02-22
display icons uniformly across all screen resolutions

### LINKED ISSUE #316: image icon size on screen resolutions greater than 1080p
After upgrading Gio versions, image icons behave different on different screen resolutions. Below is a screenshot of the overview page on a 1440 * 900 screen.  Although it looks fine on a 1080p screen, the icons on a screen with larger resolution are much smaller even when compared to the designed mockups.  
Screenshot 2021-02-11 at 16 03 36

Below is how the overview page should actually look like
Screenshot 2021-02-11 at 16 04 16

The "add wallet" button is also affected. It would have to be resized and scaled to be the intended size across all screen resolutions. 
Screenshot 2021-02-11 at 15 37 35

hydra0210:
This scales decred icons so that they are displayed in the size across screens with different resolutions. 

Fixes #316


oshorefueled:
#325 fixes the issue

PR #330, opened 2021-02-28
make entire wallet option row clickable

### LINKED ISSUE #327: entire row on wallet page dropdown should be clickable
Below is a screenshot of an active wallet menu. Currently to use any of the options, I have to click the text. A user should instead be able to click anywhere on the menu item row to use the option. Highlighting on click would also improve the user experience just like it was done in #311

dreacot:
Resolves https://github.com/planetdecred/godcr/issues/327 & https://github.com/planetdecred/godcr/issues/332

This PR;

- makes the entire rows in the wallet options drop down clickable and not just the texts
- separates the wallet menu into sections using line separators
- changes "Privacy to "StakeShuffle" 
- increases the font size of the wallet menu option titles
- fixes the right margin alignment of the wallet menu
- adds a separator widget and removes the pointer implementation of the line widget. There's no need to have the line
  widget as a pointer as there's no need for it to keep a state. The line widget is used mostly as separators with its 
  colour and height values repeatedly assigned everywhere it is used. Adding a separator widget removes the need to 
  this.

Screenshot from 2021-03-08 18-41-02


oshorefueled:
The update to the menu seems to work fine on my 1080p display. However, it doesn't display quite the same on my 1440 display. Below is a screenshot of how it looks. 

Screenshot 2021-03-09 at 06 03 35

Other thing to note:
* Remove the top shadow on the menu and put it below the menu. You can have a look at the mockups for an accurate representation. 
* Increase the margin between the two bordered options (view property and StakeShuffle). Also have a look at the mockups for an accurate representation.

ui/decredmaterial/collapsible.go
```diff
@@ -133,6 +133,7 @@ func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C)
 							)
 						}),
 						layout.Rigid(func(gtx C) D {
+							//fmt.Printf("MAX %v MIN %v\n",gtx.Constraints.Max, gtx.Constraints.Min)
```
oshorefueled: kindly remove the comment if it's not going to be used

ui/decredmaterial/line.go
```diff
@@ -36,7 +36,7 @@ func (t *Theme) Line(height, width int) Line {
 
 func (t *Theme) Separator() Line {
 	l := t.Line(1, 0)
-	l.Color = t.Color.Gray1
+	l.Color = t.Color.Background
```
oshorefueled: Background isn't the correct colour from the mockups. The correct colour is Gray1 with the hex #E6EAED, Background has the hex #444444

oshorefueled (changes requested):

oshorefueled (approved):

PR #348, opened 2021-03-16
update ui for account detail page

### LINKED ISSUE #342: defined dimensions from mockups on accounts detail page
Currently a number of items on the accounts detail page are not placed with the same dimensions as the mockups and create some inconsistencies with other pages. Although the mockup dimensions were defined in pixels (px), they work just fine in gio when specified in Density-Independent Pixels (dp). 

The dimensions on the accounts detail page should therefore be the same as that in the mockups.

JustinBeBoy:
resolve #342

### DIFF
--- a/ui/account_details_page.go
+++ b/ui/account_details_page.go
@@ -59,7 +59,7 @@ func (pg *acctDetailsPage) Layout(gtx layout.Context, common pageCommon) layout.
 		func(gtx C) D {
 			pg.line.Width = gtx.Constraints.Max.X
 			pg.line.Height = 2
-			m := values.MarginPadding5
+			m := values.MarginPadding10
 			return layout.Inset{Top: m, Bottom: m}.Layout(gtx, func(gtx C) D {
 				return pg.line.Layout(gtx)
 			})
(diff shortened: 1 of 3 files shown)


ui/account_details_page.go
```diff
@@ -249,13 +227,34 @@ func (pg *acctDetailsPage) accountInfoLayout(gtx layout.Context) layout.Dimensio
...
 		)
 	})
 }
 
+func (pg *acctDetailsPage) acctInfoLayout(gtx layout.Context, leftText, rightText string) layout.Dimensions {
+	return layout.Flex{}.Layout(gtx,
+		layout.Rigid(func(gtx C) D {
+			return layout.Flex{Axis: layout.Horizontal}.Layout(gtx,
+				layout.Rigid(func(gtx C) D {
+					leftTextLabel := pg.theme.Body1(leftText)
```
oshorefueled: This text size should be 14SP. Body1 label is 16SP

oshorefueled (changes requested):

oshorefueled (approved):

PR #353, opened 2021-03-27
Implement send to my account functionality on the send page.

Sirmorrison:
This Pr implements the send to my account functionality on the send page.
Update account selector widget.
Add info popup to the account selector widget.
Show only accounts necessary for either send or receive dcr.

image

image

image

image

### DIFF
--- a/ui/decredmaterial/switch.go
+++ b/ui/decredmaterial/switch.go
@@ -15,42 +15,42 @@ type Switch struct {
 	material.SwitchStyle
 }
 
+type SwitchItem struct {
+	Text   string
+	button Button
+}
+
 type SwitchButtonText struct {
-	t                        *Theme
-	activeTextColor          color.NRGBA
-	inactiveTextColor        color.NRGBA
-	activeBtn, inactiveBtn   *widget.Clickable
-	activeCard, inactiveCard Card
-	inactivetxt              string
-	activeTxt                string
-	isActive, isInactive     bool
+	t                                  *Theme
+	activeTextColor, inactiveTextColor color.NRGBA
+	active, inactive                   color.NRGBA
+	items                              []SwitchItem
+	selected                           int
 }
 
 func (t *Theme) Switch(swtch *widget.Bool) Switch {
 	return Switch{material.Switch(t.Base, swtch)}
 }
 
-func (t *Theme) SwitchButtonText(activeTxt, inactivetxt string, activeBtn, inactiveBtn *widget.Clickable) *SwitchButtonText {
+func (t *Theme) SwitchButtonText(i []SwitchItem) *SwitchButtonText {
 	sw := &SwitchButtonText{
-		t:           t,
-		activeBtn:   activeBtn,
-		inactiveBtn: inactiveBtn,
-
-		inactivetxt:  inactivetxt,
-		activeTxt:    activeTxt,
-		isActive:     true,
-		activeCard:   t.Card(),
-		inactiveCard: t.Card(),
+		t:     t,
+		items: make([]SwitchItem, len(i)+1),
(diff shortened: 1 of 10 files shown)


ui/decredmaterial/switch.go
```diff
@@ -63,68 +63,46 @@ func (s *SwitchButtonText) Layout(gtx layout.Context) layout.Dimensions {
...
 }
 
 func (s *SwitchButtonText) handleClickEvent() {
-	for s.inactiveBtn.Clicked() {
-		s.inactiveCard.Color = s.t.Color.Surface
-		s.activeCard.Color = color.NRGBA{}
-		s.isActive = false
-		s.isInactive = true
-	}
+	for i := range s.items {
```
dreacot: index can be assigned directly
`for index := range s.items {`

ui/values/dimensions.go
```diff
@@ -31,15 +31,21 @@ var (
...
 	TextSize10 = unit.Sp(10)
 	TextSize12 = unit.Sp(12)
 	TextSize14 = unit.Sp(14)
 	TextSize16 = unit.Sp(16)
 	TextSize18 = unit.Sp(18)
 	TextSize20 = unit.Sp(20)
+
+	STextSize20 = unit.Sp(20)
+
+	SMarginPadding16 = unit.Sp(16)
```
dreacot: these don't seem to have been used, they should be deleted if that's the case

wallet/commands.go
```diff
@@ -1056,3 +1056,11 @@ func (wal *Wallet) IsAccountMixerActive(walletID int) bool {
 func (wal *Wallet) AllWallets() []*dcrlibwallet.Wallet {
 	return wal.multi.AllWallets()
 }
+
+func (wal *Wallet) ReadMixerConfigValueForKey(key string, walletID int) int32 {
+	wall := wal.multi.WalletWithID(walletID)
+	if wall != nil {
+		return wall.ReadInt32ConfigValueForKey(key, -1)
```
dreacot:
```suggestion
	wallet := wal.multi.WalletWithID(walletID)
	if wallet != nil {
		return wallet.ReadInt32ConfigValueForKey(key, -1)
```

dreacot (changes requested):

dreacot:
while on the send page, if you switch the `To` section from `Address` to `My account`, you're mo longer able to see the info popup icon on the `sending account` dialog

Screenshot from 2021-04-13 19-05-35

Sirmorrison:
> while on the send page, if you switch the `To` section from `Address` to `My account`, you're mo longer able to see the info popup icon on the `sending account` dialog
> 
> Screenshot from 2021-04-13 19-05-35

Yes that is the design on the mock up... The info pop up is only visible when sending to address and on the receive account pop up.

ui/decredmaterial/switch.go
Sirmorrison: I dont follow. I also indicate the index. Kindly elaborate.

dreacot:
> > while on the send page, if you switch the `To` section from `Address` to `My account`, you're mo longer able to see the info popup icon on the `sending account` dialog
> > Screenshot from 2021-04-13 19-05-35
> 
> Yes that is the design on the mock up... The info pop up is only visible when sending to address and on the receive account pop up.

noted, thanks for pointing that out

ui/decredmaterial/switch.go
dreacot: you have i as the index, then reassign it to a more convenient naming of `index` on the next line
i was pointing out that it might be better to just rename `i` to `index` directly

oshorefueled (approved):

Issue #344, opened 2021-03-13
transactions page icons & dimensions update


oshorefueled:
Currently a number of items on the transactions page line the transaction direction, menu icons and margins are not uniform with the mockups. The transactions page should be adjusted to be in line with the mockups.

PR #356, opened 2021-03-29
Transaction, overview UI & code cleanup

oshorefueled:
This PR resolves issue #344. 

It also:
* contains code clean ups which include the addition of components.go, util.go and depricated.go file. Appropriate
   comments were added to each file describing the code that they contain. 
* modifies the progress bar widget. The default progress bar sets the max height to 4DP which doesn't allow a 8DP 
  height as specified in the overview mockup.
* adds a separator widget eliminating the need to initialize a Line object when initializing a page. And also repeatedly
  setting the width of the line. The same code is also present in PR #330.
* adds a Container type which is simply a wrapper to the layout.Inset type to help differentiate when it is used as 
   padding or margin. Its purpose is to help visualize the structure of a layout from the code.
* makes the transaction row layout reusable by both the transactions and overview page.


JustinBeBoy:
- The right margin is not aligned


- The button wrong text when wallet status is offline

ui/decredmaterial/progressbar.go
```diff
@@ -3,13 +3,97 @@
...
+	}
+
+	progressBarWidth := float32(gtx.Constraints.Max.X)
+	return layout.Stack{Alignment: layout.W}.Layout(gtx,
+		layout.Stacked(func(gtx layout.Context) layout.Dimensions {
+			return shader(progressBarWidth, p.TrackColor)
+		}),
+		layout.Stacked(func(gtx layout.Context) layout.Dimensions {
+			fillWidth := progressBarWidth * clamp1(p.Progress)
+			fillColor := p.Color
```
JustinBeBoy, line 56: The progress color is "41BE53" and background is E6EAED

JustinBeBoy (changes requested):

JustinBeBoy:
The margin is 24

ui/decredmaterial/progressbar.go
oshorefueled: Yes, the colours for the track and progress are set on the pages using the progress bar widget

oshorefueled:
> * The button wrong text when wallet status is offline
> 
> 

I'm not sure how to reproduce that error, you can create that as an issue and it will be resolved on another PR.

dreacot (approved):

dreacot (approved):

beansgum (approved):

JustinBeBoy (approved):

PR #365, opened 2021-04-08
Allow spending from unmixed modal

### LINKED ISSUE #360: warning popup after activating "allow spending from unmixed"
a warning popup containing a confirmation button should be displayed after toggling the "allow spending from unmixed account"  option. The settings should be saved to dcrlibwallet
Screenshot 2021-04-05 at 15 14 49

Icon Sizes should also be corrected
Screenshot 2021-04-05 at 15 16 27

song50119:
Resolves #360

### DIFF
--- a/ui/modal_templates.go
+++ b/ui/modal_templates.go
@@ -1,8 +1,11 @@
 package ui
 
 import (
+	"strings"
+
 	"gioui.org/font/gofont"
 	"gioui.org/io/key"
+
 	"gioui.org/layout"
 	"gioui.org/text"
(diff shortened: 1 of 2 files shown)


oshorefueled:
This PR needs an API from dcrlibwallet to set and unset the option to allow spending from an unmixed account and it is currently unavailable.

dreacot:
@oshorefueled this works on mobile without using an API to toggle spending from unmixed account

you can only spend from unmixed account if privacy is not enabled

and that should also work here as well

PR #366, opened 2021-04-11
Fix watch-only settings & rename features

### LINKED ISSUE #362: Fix watch-only settings & rename features
Fix settings & Rename imported watch-only features by directing them to their appropriate pages when their respective options are clicked. Currently, the settings option goes to the "not implemented" page. Same as the rename option.

tuhalang:
Resolves #362

### DIFF
--- a/ui/wallet_settings_page.go
+++ b/ui/wallet_settings_page.go
@@ -67,6 +67,13 @@ func (pg *walletSettingsPage) Layout(gtx layout.Context, common pageCommon) layo
 				common.changePage(PageWallet)
 			},
 			body: func(gtx layout.Context) layout.Dimensions {
+				if common.info.Wallets[*common.selectedWallet].IsWatchingOnly {
+					return layout.Flex{Axis: layout.Vertical}.Layout(gtx,
+						layout.Rigid(pg.notification()),
+						layout.Rigid(pg.debug()),
+						layout.Rigid(pg.dangerZone()),
+					)
+				}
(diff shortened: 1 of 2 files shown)


oshorefueled:
Screenshot 2021-04-12 at 17 43 11

This option should not be displayed, a watch-only wallet doesn't have a spending password.

ui/wallet_settings_page.go
```diff
@@ -67,6 +67,13 @@ func (pg *walletSettingsPage) Layout(gtx layout.Context, common pageCommon) layo
 				common.changePage(PageWallet)
 			},
 			body: func(gtx layout.Context) layout.Dimensions {
+				if common.info.Wallets[*common.selectedWallet].IsWatchingOnly {
```
oshorefueled: This looks like a repetition and it could be handled with an if statement inside the rigid layout. Something like this 
`layout.Rigid(`
						`func(gtx layout.Context) layout.Dimensions {`
							`  if !common.info.Wallets[*common.selectedWallet].IsWatchingOnly {`
								`return pg.changePassphrase()(gtx)`
							`}`
							`return layout.Dimensions{}`
						`}),`
					`layout.Rigid(pg.notification()),`
					`layout.Rigid(pg.debug()),`
					`layout.Rigid(pg.dangerZone()),`

oshorefueled (changes requested):

oshorefueled (approved):

PR #367, opened 2021-04-12
Fix account selector and update send page ui

### LINKED ISSUE #359: fix account selector
Account selector widget is currently showing the overall balance rather than the spendable balance which creates some confusion on the send page. In the screenshot below, the account selector shows a balance of 170 DCR but a balance after send of 59 DCR. They should both display 59 DCR because that's the spendable amount. 
Screenshot 2021-03-31 at 11 39 53

The account selector widget shouldn't extend all the way to the top and bottom as in the screenshot below. 
Screenshot 2021-03-31 at 11 40 15

Instead, If there are too many wallets/accounts: 
1. keep the dialog top/bottom margin at 24DP. 
2. title section is floated at top.
Just as in the screenshot from the mockup below.
Screenshot 2021-03-31 at 11 53 23

JustinBeBoy:
This PR resolve #359

### DIFF
--- a/ui/send_page.go
+++ b/ui/send_page.go
@@ -160,7 +160,7 @@ func (win *Window) SendPage(common pageCommon) layout.Widget {
 		txFeeCollapsible:             common.theme.Collapsible(),
 
 		confirmModal:              common.theme.Modal(),
-		walletAccountModal:        common.theme.Modal(),
+		walletAccountModal:        common.theme.ModalFloatTitle(),
 		isConfirmationModalOpen:   false,
 		isBroadcastingTransaction: false,
 		isWalletAccountModalOpen:  false,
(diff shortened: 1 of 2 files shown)


ui/send_page.go
```diff
@@ -402,7 +402,10 @@ func (pg *sendPage) fromSection(gtx layout.Context, common pageCommon) layout.Di
 							return layout.E.Layout(gtx, func(gtx C) D {
 								return layout.Flex{}.Layout(gtx,
 									layout.Rigid(func(gtx C) D {
-										return pg.theme.Body1((*pg.selectedAccount).TotalBalance).Layout(gtx)
+										spendableBalance := dcrutil.Amount((*pg.selectedAccount).SpendableBalance).String()
+										txt := pg.theme.Body1(spendableBalance)
+										txt.Color = pg.theme.Color.DeepBlue
```
oshorefueled: The default text colour for body1 text is already DeepBlue. There's no need to set the value again in the the code.

oshorefueled (changes requested):

oshorefueled (approved):

PR #380, opened 2021-04-21
add usd equivalent balance to navbar

### LINKED ISSUE #374: add dollar balance to top bar
Dollar balance should be displayed when the dollar options is activated in the settings. It should be displayed as in the screenshot below. 

Screenshot 2021-04-15 at 20 45 44

dreacot:
Resolves #374 

This PR adds the USD equivalent of a user's DCR balance to the overview page.

NB: Currency Conversion needs to be enabled in the settings page for this to work.

Screenshots

Screenshot from 2021-04-23 02-08-37


ui/compontents.go
```diff
@@ -33,6 +34,49 @@ func (page pageCommon) layoutBalance(gtx layout.Context, amount string) layout.D
 		layout.Rigid(func(gtx C) D {
 			return page.theme.Label(values.TextSize14, subText).Layout(gtx)
 		}),
+		layout.Rigid(func(gtx C) D {
+			if topnav {
+				return page.layoutUSDBalance(gtx)
+			}
+			return D{}
+		}),
```
oshorefueled: The app header component is probably going to be the only component that uses that design for currency conversion. I don't think you should add the dollar layout to `layoutBalance` that's used by other parts of the app. It results in the need to pass a "true" or "false" to the method by other parts of the app that would never use the layout. You can simply align the dollar balance vertically beside the DCR balance on the header, rather than adding it to the layoutBalance method.

oshorefueled (changes requested):

oshorefueled (approved):

PR #382, opened 2021-04-22
Move Proposals listeners to Wallet Package

Sirmorrison:
* Move proposals initialization from proposals page to state.go
* Move Proposals sync listeners to wallet package.
* Restored the restore button after sync is completed
* Syncs proposal when no proposal has been fetched (initial sync).

### DIFF
--- a/ui/proposals_page.go
+++ b/ui/proposals_page.go
@@ -23,31 +23,6 @@ import (
 )
 
 const PageProposals = "Proposals"
-const (
-	categoryStateFetching = iota
-	categoryStateFetched
-	categoryStateError
-)
-
-type proposalNotificationListeners struct {
-	page *proposalsPage
-}
-
-func (p proposalNotificationListeners) OnNewProposal(proposal *dcrlibwallet.Proposal) {
-	p.page.addDiscoveredProposal(*proposal)
-}
-
-func (p proposalNotificationListeners) OnProposalVoteStarted(proposal *dcrlibwallet.Proposal) {
-	p.page.updateProposal(*proposal)
-}
-
-func (p proposalNotificationListeners) OnProposalVoteFinished(proposal *dcrlibwallet.Proposal) {
-	p.page.updateProposal(*proposal)
-}
-
-func (p proposalNotificationListeners) OnProposalsSynced() {
-	p.page.isSynced = true
-}
 
 type proposalItem struct {
 	btn               *widget.Clickable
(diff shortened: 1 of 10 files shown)


wallet/commands.go
```diff
@@ -653,17 +653,30 @@ func (wal *Wallet) RenameAccount(walletID int, acct int32, name string, errChan
...
 }
 
+func (wal *Wallet) ClearSavedProposals() {
+	err := wal.multi.Politeia.ClearSavedProposals()
+	if err != nil {
+		fmt.Println(err.Error())
+	}
+
+}
+
```
oshorefueled: This doesn't seem to be used anywhere

ui/proposals_page.go
```diff
@@ -72,27 +47,25 @@ type tabs struct {
...
-	updatedIcon                *widget.Icon
-	updatedLabel               decredmaterial.Label
-	syncButton                 *widget.Clickable
-	startSyncIcon              *widget.Image
-	timerIcon                  *widget.Image
+	theme            *decredmaterial.Theme
+	wallet           *wallet.Wallet
+	selectedProposal **dcrlibwallet.Proposal
+	proposals        **wallet.Proposals
+	proposalSync     *chan *wallet.NewProposal
```
oshorefueled: Is there any reason a pointer to the channel was used here?

ui/proposals_page.go
```diff
@@ -168,81 +142,33 @@ func (pg *proposalsPage) Handle(common pageCommon) {
...
-func (pg *proposalsPage) onFetchError(err error) {
-	pg.state = categoryStateError
-	pg.onFetchComplete()
-	pg.notify(err.Error(), false)
-}
-
-func (pg *proposalsPage) onFetchComplete() {
-	if !pg.hasFetchedInitialProposals {
-		pg.hasFetchedInitialProposals = true
+	default:
```
oshorefueled, line 161: There's no need for a default case if the case is empty

ui/proposals_page.go
```diff
@@ -541,24 +496,45 @@ func (pg *proposalsPage) layoutStartSyncSection(gtx C) D {
...
+			}
+			for k := range pg.tabs.tabs {
+				if pg.tabs.tabs[k].category == (*pg.proposals).Proposals[i].Category {
+					pg.tabs.tabs[k].proposals = append(pg.tabs.tabs[k].proposals, item)
+					break
+				}
+			}
+		}
+	}
+}
```
oshorefueled: This looks like a duplicate of the code in `addDiscoveredProposal` and the method can simply be called here.

ui/proposals_page.go
```diff
@@ -308,23 +234,56 @@ func (pg *proposalsPage) layoutTabs(gtx C) D {
...
+			}
+			pg.tabs.tabs[i].proposals = append([]proposalItem{item}, pg.tabs.tabs[i].proposals...)
+			break
+		}
+	}
 }
 
-func (pg *proposalsPage) layoutErrorState(gtx C) D {
-	str := "Error fetching proposals"
+func (pg *proposalsPage) updateProposal(proposal dcrlibwallet.Proposal) {
```
oshorefueled: Kindly add a comment here or a more descriptive method name. What about the proposal is being updated?

wallet/proposal_listener.go
```diff
@@ -0,0 +1,58 @@
...
+type ProposalStatus int
+
+const (
+	Synced ProposalStatus = iota
+	VoteStarted
+	NewProposalFound
+	VoteFinished
+)
+
+type NewProposal struct {
```
oshorefueled: Proposal is just fine here

oshorefueled (changes requested):

ui/proposals_page.go
Sirmorrison: It is returning the proposal sync status and also the proposal from the proposal listener.

ui/proposals_page.go
Sirmorrison: Removing the default cases causes the app to freeze when proposals page is selected. I spent the last 30 mins trying to figure out this behavior.

ui/proposals_page.go
oshorefueled: I don't think you need a pointer to a channel to do that

ui/proposals_page.go
oshorefueled: oh I see. It's a select case and removing the default case causes the app to wait for a signal from the syncedProposal channel.

oshorefueled (changes requested):

oshorefueled (approved):

Issue #383, opened 2021-04-24
back button on transactions detail always navigates to the transactions page


oshorefueled:
The details of a transaction can be seen from the overview page. However, when I click on the back button on the transaction detail page, I should be navigated back to the overview page not the transactions list page as shown in the screen recording below. The transaction detail page needs a state of the previous page which will be navigated to when the back button is clicked.

![backbutton-bug](https://user-images.githubusercontent.com/15804688/115941976-adf1ea80-a49f-11eb-8ac5-6786dcd4e6a3.gif)

PR #384, opened 2021-04-25
fix return page

tuhalang:
This PR resolve issue #383

### DIFF
--- a/ui/overview_page.go
+++ b/ui/overview_page.go
@@ -679,6 +679,8 @@ func (pg *overviewPage) Handler(eq event.Queue, c pageCommon) {
 			if e.Type == gesture.TypeClick {
 				txn := (*pg.walletTransactions).Recent[index]
 				*pg.walletTransaction = &txn
+
+				c.setReturnPage(PageOverview)
 				c.changePage(PageTransactionDetails)
 				return
 			}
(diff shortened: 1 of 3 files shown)


ui/overview_page.go
```diff
@@ -679,6 +679,8 @@ func (pg *overviewPage) Handler(eq event.Queue, c pageCommon) {
 			if e.Type == gesture.TypeClick {
 				txn := (*pg.walletTransactions).Recent[index]
 				*pg.walletTransaction = &txn
+				*c.returnPage = PageOverview
+				c.setReturnPage(PageOverview)
```
oshorefueled: This looks like the returnPage field is being set twice. It is set directly on line 682 once and a second time with a method 683.

ui/transactions_page.go
```diff
@@ -260,11 +260,14 @@ func (pg *transactionsPage) sortTransactions(common *pageCommon) {
...
 
-func (pg *transactionsPage) goToTxnDetails(events []gesture.ClickEvent, changePage func(string), txn *wallet.Transaction) {
+func (pg *transactionsPage) goToTxnDetails(events []gesture.ClickEvent, common *pageCommon, txn *wallet.Transaction) {
 	for _, e := range events {
 		if e.Type == gesture.TypeClick {
 			*pg.walletTransaction = txn
-			changePage(PageTransactionDetails)
+
+			*common.returnPage = PageTransactions
+			common.setReturnPage(PageTransactions)
```
oshorefueled: This looks like the returnPage field is being set twice. It is set directly on line 268 once and a second time with a method 269.

oshorefueled (changes requested):

oshorefueled (approved):

PR #391, opened 2021-04-28
Fix navigation when click the cancel button on the send and receive page

### LINKED ISSUE #390: cancel button on send and receive page always navigates to the overview page
When I click on the cancel button on the send and receive page, I should be navigated back to the previous page not the overview page. The send and receive pages need a state of the previous page which will be navigated to when the cancel button is clicked.

tuhalang:
Resolves #390 
- Currently, when clicking on the cancel button on the send and receive page, it always navigates to the overview page. This modifies the cancel button navigation to return to the previous page.

### DIFF
--- a/ui/receive_page.go
+++ b/ui/receive_page.go
@@ -331,7 +331,7 @@ func (pg *receivePage) Handle(common pageCommon) {
 	}
 
 	if common.subPageBackButton.Button.Clicked() {
-		*common.page = PageOverview
+		*common.page = *common.returnPage
 	}
 
 	if pg.copy.Button.Clicked() {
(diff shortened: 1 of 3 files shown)


ui/receive_page.go
```diff
@@ -331,7 +331,7 @@ func (pg *receivePage) Handle(common pageCommon) {
 	}
 
 	if common.subPageBackButton.Button.Clicked() {
-		*common.page = PageOverview
+		*common.page = *common.returnPage
```
oshorefueled: the changePage method should be used rather than a direct assignment.

oshorefueled (changes requested):

oshorefueled (approved):

PR #403, opened 2021-05-09
fix bug with font weight and style

Sirmorrison:
* This PR fix issues with making setting the font weight to bold or styling to italics.
* Update font weight and size across multiple pages 
* fix wrong privacy info modal template title 
* use subpage methods for tx details page 
* use modal template for tx details info and fix wrong info title 
* clean up template modal codes. 
* fix bug with yes an no label not displaying on proposals page.


ui/modal_templates.go
```diff
@@ -2,7 +2,7 @@ package ui
 
 import (
 	"gioui.org/layout"
-	"gioui.org/text"
+	w "gioui.org/text"
```
oshorefueled: I'm not sure what "w" denotes here as the import name.

ui/modal_templates.go
```diff
@@ -295,7 +296,8 @@ func (m *ModalTemplate) privacyInfo() []func(gtx C) D {
 			)
 		},
 		func(gtx C) D {
-			text := m.th.Label(values.TextSize18, "Important: keep this app opened while mixer is running.")
+			text := m.th.Body1("Important: keep this app opened while mixer is running.")
+			text.Font.Weight = w.Bold
```
oshorefueled: If the reason you used "w" as the import name for `w "gioui.org/text"` is because you have text as a variable name here then that's unnecessary. You can change "text" to just "t" as your variable and use text.Bold for your font weight.

oshorefueled (changes requested):

oshorefueled (approved):

PR #412, opened 2021-05-15
Replace win.clipboard with gio clipboard api

### LINKED ISSUE #409: use gio clipboard API
godcr/ui.Window uses the raw app.Window.ReadClipboard and WriteClipboard methods for system clipboard access. The 
 gioui.org/op/clipboard API should be used instead. It contains tagged ReadOp and WriteOp operations that can be used from any Gio widget, without access to an app.Window. Gio automatically takes care of multiplexing multiple clipboard readers and writers.

This is a reference to the review made by Elias: https://paste.sr.ht/~eliasnaur/cea1d29d6a5f96668b5e166c2f39ef596974574f

beansgum:
This pr replaces the default `win.clipboard`  with gio's clipboard which can be accessed from any page rather than requiring access to `app.Window`.

Closes #409

### DIFF
--- a/ui/window.go
+++ b/ui/window.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"gioui.org/app"
-	"gioui.org/io/clipboard"
 	"gioui.org/io/key"
 	"gioui.org/io/system"
 	"gioui.org/layout"
@@ -58,7 +57,6 @@ type Window struct {
 	pages                   map[string]layout.Widget
 	walletTabs, accountTabs *decredmaterial.Tabs
 	keyEvents               chan *key.Event
-	clipboard               chan interface{}
 	toast                   chan *toast
 	modal                   chan *modalLoad
 	sysDestroyWithSync      bool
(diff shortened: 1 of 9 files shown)


ui/window.go
```diff
@@ -254,28 +250,11 @@ func (win *Window) Loop(shutdown chan int) {
...
-			case decredmaterial.ReadClipboard:
-				win.window.ReadClipboard()
-			case WriteClipboard:
-				go func() {
-					win.window.WriteClipboard(c.Text)
-					win.toast <- &toast{
-						text:    "copied",
-						success: true,
-					}
-				}()
```
oshorefueled, line 277: As a result of the change, the toast displayed after triggering a copy event doesn't display anymore. I'll suggest using a global list for the toast, similar to how the modal is going to be refactored as described in issue #424.

oshorefueled (changes requested):

oshorefueled (approved):

PR #418, opened 2021-05-17
Replace explicit DPI conversions with standard unit type

### LINKED ISSUE #416: Replace explicit DPI conversions with standard unit type
running `$ grep "\.PxPer`

returns a few instances of explicit use of the pixel-to-dp scale. The pageCommon.layoutNavDrawer converts from Dps to pixels:

`gtx.Constraints.Min.X = int(gtx.Metric.PxPerDp) * width`

However, PxPerDp may not be integral; a better version is

`gtx.Constraints.Min.X = int(gtx.Metric.PxPerDp * width)`

Even better is to change the type of width, navDrawerWidth, navDrawerMinimizedWidth to unit.Value, and use

`gtx.Constraints.Min.X = gtx.Px(width)`

A related issue is the godcr/ui/decredmaterial.pxf function; it is a copy of gioui.org/unit.Metric.Px except that it special-cases the zero scale. I think it is better to never construct zero-valued unit.Metrics and use the standard Px method instead.

This is in reference to Elias' review: https://paste.sr.ht/~eliasnaur/cea1d29d6a5f96668b5e166c2f39ef596974574f

tuhalang:
Resolve #416 
- Changed the type of width to unit.Value

### DIFF
--- a/ui/compontents.go
+++ b/ui/compontents.go
@@ -158,9 +158,9 @@ func (page pageCommon) layoutTopBar(gtx layout.Context) layout.Dimensions {
 	})
 }
 
-const (
-	navDrawerWidth          = 160
-	navDrawerMinimizedWidth = 72
+var (
+	navDrawerWidth          = unit.Value{U: unit.UnitDp, V: 160}
+	navDrawerMinimizedWidth = unit.Value{U: unit.UnitDp, V: 72}
 )
 
 // layoutNavDrawer is the left vertical pane on every page of the app. It vertically lays out buttons used to navigate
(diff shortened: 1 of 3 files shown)


beansgum (changes requested):
There are other places where this change is needed.
https://github.com/planetdecred/godcr/blob/master/ui/tickets_page.go#L531
https://github.com/planetdecred/godcr/blob/master/ui/wallet_page.go#L378

beansgum (approved):

oshorefueled (approved):

PR #426, opened 2021-05-20
Localize strings

### LINKED ISSUE #422: Localize strings
Currently, the strings are statically typed in codes which makes the app impossible to translate and reuse strings. All strings should be kept in a global package for easy access anywhere in the project.

beansgum:
Closes #422 
This pull request adds strings localization feature to the project which will enable strings to be translated to multiple languages. Currently, only a few pages have had their strings moved to strings file and only 2 translations have been added (French & Chinese). The `values.String(key)` function is used the get strings and the key is passed as a parameter, the function also handles falling back to the default language if a string is not available for the current language. There's also a `values.StringF(key, format...)` function used for getting strings that require formatting (e.g `"Step %d/3"`).  

This pr also adds `ListPreference` for the handling of list settings and it's currently being used to implement the exchange rate and language settings. This `ListPreference` helps by taking away the entire UI logic for a list setting out of the settings page and handle showing/dismissing the modal when tapped.

ezgif com-video-to-gif-2


ui/preference/list_preference.go
```diff
@@ -0,0 +1,202 @@
+package preference
```
oshorefueled, line 1: This seems to be a widget that's reused by the settings page. In that case, it would be appropriate to have it in the decredmaterial package.

ui/preference/list_preference.go
```diff
@@ -0,0 +1,202 @@
...
+	"sort"
+
+	"gioui.org/layout"
+	"gioui.org/widget"
+	"github.com/planetdecred/godcr/ui/decredmaterial"
+	"github.com/planetdecred/godcr/ui/values"
+	"github.com/planetdecred/godcr/wallet"
+)
+
+type ListPreference struct {
```
oshorefueled, line 14: ListPreference isn't a self descriptive name. I can't tell what it does from just its name unlike existing widgets.

ui/preference/list_preference.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	for lp.optionsRadioGroup.Changed() {
+		lp.currentValue = lp.optionsRadioGroup.Value
+	}
+}
+
+func (lp *ListPreference) setValue(value string) {
+	lp.wallet.SaveConfigValueForKey(lp.preferenceKey, value)
+	values.SetUserLanguage(value)
```
oshorefueled: your current use of the preference widget is for the currency and language options in the settings page. This seems to set the user language option on every implementation of the widget.

ui/preference/list_preference.go
```diff
@@ -0,0 +1,202 @@
...
+	keys := make([]string, 0)
+	for k := range lp.items {
+		keys = append(keys, k)
+	}
+
+	sort.Slice(keys, func(i int, j int) bool { return keys[i] < keys[j] })
+
+	items := make([]layout.FlexChild, 0)
+	for _, k := range keys {
+		radioItem := layout.Rigid(lp.theme.RadioButton(lp.optionsRadioGroup, k, values.String(lp.items[k])).Layout)
```
oshorefueled, line 165: This seems to be an implementation to selection options using radio buttons. Maybe RadioSelection would be a more descriptive name for the widget.

oshorefueled (changes requested):

ui/preference/list_preference.go
beansgum: It's not a widget, it's specifically for list preferences and it saves the select value to the config database.

oshorefueled (approved):

PR #430, opened 2021-05-22
use a global list for toasts

### LINKED ISSUE #425: use a global list for toasts
Use a global list for toasts, similar to how the modal is going to be refactored as described in issue #424. Rather than use a channel to trigger toasts as currently done, a method can be added that adds toasts to a global list of toasts. The toast is displayed on the UI anytime there's a toast added to the list. It can then delete the toast from the list after a couple of seconds.

song50119:
Resolves #425


ui/page.go
```diff
@@ -112,8 +112,8 @@ type pageCommon struct {
 	walletTabs      *decredmaterial.Tabs
 	accountTabs     *decredmaterial.Tabs
 	keyEvents       chan *key.Event
-	toast           chan *toast
-	toastLoad       *toast
+	toasts          *[]*toast
```
oshorefueled: Rather than using a pointer to a slice, make the method on the pageCommon struct to use a pointer receiver. Change the other methods to pointer receivers as well to maintain consistency.

ui/page.go
```diff
@@ -299,8 +297,11 @@ func (win *Window) addPages(decredIcons map[string]image.Image) {
 		changePage:              win.changePage,
 		setReturnPage:           win.setReturnPage,
 		refreshWindow:           win.refresh,
+		toastList:               layout.List{Axis: layout.Vertical},
```
oshorefueled: I don't think godcr has a use case where there will be more than one notification at once. That means there's no need to create a list layout.

ui/toast.go
```diff
@@ -37,14 +37,19 @@ func displayToast(th *decredmaterial.Theme, gtx layout.Context, n *toast) layout
 	})
 }
 
-// Timer implements the standard time.AfterFunc. It only creates a new time.Timer when toast.timer is nil.
+// Timer is only creates a new time.Timer when toast.timer is nil.
 // Gio re-renders its UI recursively and Timer prevents multiple time.Timer from being created before the duration is
 // exceeded.
 func (n *toast) Timer(d time.Duration, f func()) {
```
oshorefueled: This method can be removed. Using time.AfterFunc would be sufficient. The entire toast.go file can be deleted and `displayToast` method can be added to components.go

ui/page.go
```diff
@@ -473,37 +472,27 @@ func (page pageCommon) Layout(gtx layout.Context, body layout.Widget) layout.Dim
...
-				default:
-					break outer
-				}
-			}
-
-			if page.toastLoad.text != "" {
-				page.toastLoad.Timer(time.Second*3, func() {
-					page.toastLoad.text = ""
+					}
+					return t((*page.toasts)[index])
```
oshorefueled: You can simply loop through the list of toasts, call the time.AfterFunc method to remove that toast from the list in the specified duration. You should add a mutex lock for appending and removing data from the toast list so a data race issue doesn't occur when trying to access data in the list.

oshorefueled (changes requested):

ui/page.go
song50119: If only display toast at once app only need single struct of toast, I don't think app need slice of toasts

ui/compontents.go
```diff
@@ -940,7 +948,45 @@ func ticketCard(gtx layout.Context, c pageCommon, t *wallet.Ticket) layout.Dimen
...
+	}
+
+	if (*page.toast).timer == nil {
+		(*page.toast).timer = time.NewTimer(time.Second * 3)
+	}
+
+	select {
+	case <-(*page.toast).timer.C:
+		*page.toast = nil
+	default:
```
oshorefueled, line 983: This can be replaced with time.AfterFunc

oshorefueled (changes requested):

oshorefueled (approved):

PR #438, opened 2021-05-26
fix crash that occured due to invalid memory address

dreacot:
This PR fixes the crashes that occurs on both the Receive Page and Transaction Details Page


oshorefueled (dismissed):

Sirmorrison (changes requested):
This issue also happens on the sign message page as well.

Sirmorrison (approved):

PR #439, opened 2021-05-26
Implement ticket activity page

song50119:
PR implement for ticket page activity
Screen Shot 2021-05-27 at 01 00 04

### DIFF
--- a/ui/tickets_activity_page.go
+++ b/ui/tickets_activity_page.go
@@ -34,7 +34,7 @@ func (win *Window) TicketActivityPage(c pageCommon) Page {
 		tickets:     &win.walletTickets,
 		ticketsList: layout.List{Axis: layout.Vertical},
 	}
-	pg.orderDropDown = c.theme.DropDown([]decredmaterial.DropDownItem{{Text: "Newest"}, {Text: "Oldest"}}, 1)
+	pg.orderDropDown = createOrderDropDown(&c)
 	pg.ticketTypeDropDown = c.theme.DropDown([]decredmaterial.DropDownItem{
 		{Text: "All"},
 		{Text: "Unmined"},
@@ -79,7 +79,7 @@ func (pg *ticketsActivityPage) Layout(gtx layout.Context) layout.Dimensions {
 								}
 								return layout.UniformInset(values.MarginPadding16).Layout(gtx, func(gtx C) D {
 									return pg.ticketsList.Layout(gtx, len(tickets), func(gtx C, index int) D {
-										return ticketRowActivity(gtx, &c, tickets[index], index)
+										return ticketActivityRow(gtx, &c, tickets[index], index)
 									})
 								})
 							})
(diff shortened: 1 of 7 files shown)


ui/tickets_activity_page.go
```diff
@@ -0,0 +1,162 @@
...
+			},
+			body: func(gtx C) D {
+				walletID := c.info.Wallets[pg.walletDropDown.SelectedIndex()].ID
+				tickets := (*pg.tickets).Confirmed[walletID]
+				return layout.Stack{Alignment: layout.N}.Layout(gtx,
+					layout.Expanded(func(gtx C) D {
+						return layout.Inset{Top: values.MarginPadding60}.Layout(gtx, func(gtx C) D {
+							return c.theme.Card().Layout(gtx, func(gtx C) D {
+								gtx.Constraints.Min = gtx.Constraints.Max
+								if pg.ticketTypeDropDown.SelectedIndex()-1 != -1 {
```
oshorefueled, line 68: -1 in the if check is a magic number. Why is pg.ticketTypeDropDown being checked with "-1"

ui/tickets_activity_page.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return t
+}
+
+func (pg *ticketsActivityPage) handle() {
+	c := pg.common
+	initWalletDropDown(c, &pg.walletDropDown)
+
+	sortSelection := pg.orderDropDown.SelectedIndex()
+	if pg.filterSorter != sortSelection {
```
oshorefueled, line 144: Looks like this filterSorter check is for the purpose of only executing the ticket filter code when the dropdown is clicked. A better implementation is to have a Clicked() event emitted by the dropdown widget when the dropdown widget is clicked. You can look at how the Clicked() method is implemented in the clickable widget.

ui/tickets_activity_page.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+		return c.SubPageLayout(gtx, page)
+	}
+
+	return c.Layout(gtx, func(gtx C) D {
+		return c.UniformPadding(gtx, body)
+	})
+}
+
+func (pg *ticketsActivityPage) dropDowns(gtx layout.Context) layout.Dimensions {
```
oshorefueled, line 102: The wallet dropdown, sort dropdown and a page specific dropdown seems to be used in several pages. They could be implemented as "components" that can be called from each of those pages. This eliminates repetition.

ui/compontents.go
```diff
@@ -960,6 +960,92 @@ func ticketCard(gtx layout.Context, c pageCommon, t *wallet.Ticket) layout.Dimen
 	})
 }
 
+// ticketRowActivity layouts out ticket info, display ticket activities on the tickets_page and tickets_activity_page
+func ticketRowActivity(gtx layout.Context, c *pageCommon, t wallet.Ticket, index int) layout.Dimensions {
```
oshorefueled: This looks like it should be named `ticketActivityRow`

oshorefueled (changes requested):

ui/tickets_activity_page.go
song50119: Dropdowns is init with item + 1
`c := &DropDown{
		isOpen:         false,
		items:          make([]DropDownItem, len(items)+1), 
...`

oshorefueled (approved):

oshorefueled (approved):

PR #440, opened 2021-05-28
fix loader display bug on password modal when error occurs

### LINKED ISSUE #431: loader doesn't stop displaying on confirm modal
The loader doesn't stop displaying on the send modal when a wrong password is submitted. It should revert back to the send button when there's an error on the modal. The same thing occurs when the send button is clicked and godcr is connected to the decred network.

Screenshot 2021-05-23 at 23 51 22

Sirmorrison:
This pr fix #431
fix loader display bug on password modal when error occurs
* clean up sign message page code, prevent password
 modal from closing when there is a password error.

### DIFF
--- a/wallet/responses.go
+++ b/wallet/responses.go
@@ -157,7 +157,6 @@ type SyncStatus struct {
 // Signature is sent in response to Wallet.SignMessage
 type Signature struct {
 	Signature string
-	Err       error
 }
 
 // TxHash is sent when the Wallet successfully broadcasts a transaction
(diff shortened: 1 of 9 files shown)


wallet/responses.go
```diff
@@ -157,7 +157,7 @@ type SyncStatus struct {
 // Signature is sent in response to Wallet.SignMessage
 type Signature struct {
 	Signature string
-	Err       error
+	// Err       error
```
oshorefueled: Removed commented line if it's not going to be used

oshorefueled (changes requested):

oshorefueled (approved):

PR #443, opened 2021-05-29
Separate Modals

### LINKED ISSUE #424: replace current modal trigger with a global list
This is the structure of how how modals are currently being triggered

```
go func() {
	common.modalReceiver <- &modalLoad{
		...
	}
}()
```

Displaying the modal is a single threaded concept and the use of goroutines to trigger them is unnecessary. It is non-deterministic and racy. A function should be used for adding a modal to a global list of modals. The modal at the top of the list would be displayed and when the cancel button is clicked, it gets popped from the list.  If a modal must be displayed as a result of an external event, use a separate channel for those events and construct the modals in the same goroutine as the rest of the UI code.

This is a reference to Elias' review https://paste.sr.ht/~eliasnaur/cea1d29d6a5f96668b5e166c2f39ef596974574f

hydra0210:
This separates modals into their self-containing files. 

Will fix #424  by displaying the modals in the same UI thread as the app other than what we have now.

### DIFF
--- a/ui/modals/change_password.go
+++ b/ui/modals/change_password.go
@@ -1,0 +1,37 @@
+package modals
+
+import (
+	"gioui.org/layout"
+)
+
+type ChangePassword struct {
+	title     string
+	*common
+}
+
+const ChangePasswordModal = "ChangePassword"
+
+func (m *Modals) registerChangePasswordModal() {
+	m.modals[ChangePasswordModal] = &ChangePassword{
+		common: m.common,
+		title:     "Change Password",
+	}
+}
+
+func (m *ChangePassword) getTitle() string {
+	return m.title
+}
+
+func (m *ChangePassword) onCancel()  {}
+func (m *ChangePassword) onConfirm() {}
+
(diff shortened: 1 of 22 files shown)


oshorefueled:
* Registering Modals
All modals seem to be initialized or registered when the loadPage method is called and assigned to a global variable `modals` on the common struct. https://github.com/planetdecred/godcr/blob/3b45a76debe747dd77da27f2409809843f2e2b6c/ui/page.go#L315

I don't think it needs to be initialized at the start of the app. The modal can be initialized from each page, similar to the editor or preferences widget. They both handle their events in the widget but are initialized from pages that use them. The modal can be implemented the same way. The only difference with the modals though is that there'll be a dedicated container created at the root of the app where all modals are displayed. When a modal needs to be displayed, it gets pushed/added to the global list of modals and it can be laid out in the dedicated container. 

* common struct
From the code, it looks like it is trying to group widgets that are shared among modals that use similar widgets. I think the code can be cleaner without the common struct.

oshorefueled:
#460 resolves the issue this PR references

PR #423, opened 2021-05-19
Statistics page

oluwandabira:
Adds a statistics page

PR #445, opened 2021-05-29
Statistics Page

song50119:
This PR continue process #423

### DIFF
--- a/ui/statistics_page.go
+++ b/ui/statistics_page.go
@@ -3,9 +3,11 @@ package ui
 import (
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"gioui.org/layout"
+	"gioui.org/op"
 	"github.com/planetdecred/godcr/ui/decredmaterial"
 	"github.com/planetdecred/godcr/ui/values"
 	"github.com/planetdecred/godcr/wallet"
@@ -18,8 +20,9 @@ type statPage struct {
 	txs         **wallet.Transactions
 	theme       *decredmaterial.Theme
 	l           layout.List
-	startupTime string
+	startupTime time.Time
 	syncStatus  *wallet.SyncStatus
+	netType     string
 }
 
 func (win *Window) StatPage(common pageCommon) Page {
@@ -31,16 +34,15 @@ func (win *Window) StatPage(common pageCommon) Page {
 			Axis: layout.Vertical,
 		},
 	}
-	pg.startupTime = time.Now().String()
+	pg.startupTime = time.Now()
 	pg.syncStatus = win.walletSyncStatus
+	if common.wallet.Net == "testnet3" {
+		pg.netType = "Testnet"
+	} else {
+		pg.netType = strings.Title(common.wallet.Net)
+	}
 
 	return pg
-}
-
-func (pg *statPage) lineSeparator() layout.Widget {
-	return func(gtx C) D {
-		return pg.theme.Separator().Layout(gtx)
-	}
 }
 
 func (pg *statPage) layoutStats(gtx C) D {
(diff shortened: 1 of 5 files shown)


ui/statistics_page.go
```diff
@@ -0,0 +1,120 @@
...
+	pg.startupTime = time.Now().String()
+	pg.syncStatus = win.walletSyncStatus
+
+	return pg
+}
+
+func (pg *statPage) lineSeparator() layout.Widget {
+	return func(gtx C) D {
+		return pg.theme.Separator().Layout(gtx)
+	}
```
Sirmorrison:
```suggestion
		return pg.theme.Separator().Layout
```

ui/statistics_page.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+const PageStat = "Stat"
+
+type statPage struct {
+	common      pageCommon
+	txs         **wallet.Transactions
+	theme       *decredmaterial.Theme
+	l           layout.List
+	startupTime string
```
Sirmorrison: Not needed and can be removed.
```suggestion
	
```

ui/statistics_page.go
```diff
@@ -0,0 +1,120 @@
...
+func (win *Window) StatPage(common pageCommon) Page {
+	pg := &statPage{
+		txs:    &win.walletTransactions,
+		common: common,
+		theme:  common.theme,
+		l: layout.List{
+			Axis: layout.Vertical,
+		},
+	}
+	pg.startupTime = time.Now().String()
```
Sirmorrison: Not needed and can be removed.
```suggestion

```

ui/statistics_page.go
```diff
@@ -0,0 +1,120 @@
...
+				return endToEndRow(gtx, l.Layout, r.Layout)
+			})
+		}
+	}
+	items := []layout.Widget{
+		item("Build", pg.common.wallet.Net+", "+time.Now().Format("2006-01-02")),
+		pg.lineSeparator(),
+		item("Peers connected", strconv.Itoa(int(pg.syncStatus.ConnectedPeers))),
+		pg.lineSeparator(),
+		item("Uptime", pg.startupTime),
```
Sirmorrison:
```suggestion
		item("Uptime", time.Now().String()),
```

ui/statistics_page.go
```diff
@@ -0,0 +1,120 @@
...
+func (pg *statPage) layoutStats(gtx C) D {
+	background := pg.common.theme.Color.Surface
+	card := pg.common.theme.Card()
+	card.Color = background
+	inset := layout.Inset{
+		Top:    values.MarginPadding12,
+		Bottom: values.MarginPadding12,
+		Right:  values.MarginPadding16,
+		Left:   values.MarginPadding16,
+	}
```
Sirmorrison, line 57: should the values need be updated at any point, you have have one point of edit.
```suggestion
       mtb := values.MarginPadding12,
       mlr := values.MarginPadding16,
	inset := layout.Inset{
		Top:    mtb,
		Bottom: mtb,
		Right:  mrl,
		Left:   mrl,
	}
```

ui/statistics_page.go
```diff
@@ -0,0 +1,120 @@
...
+
+	item := func(t, v string) layout.Widget {
+		return func(gtx C) D {
+			l := pg.theme.Label(values.TextSize14, t)
+			r := pg.theme.Label(values.TextSize14, v)
```
Sirmorrison:
```suggestion
			l := pg.theme.Body2(t)
			r := pg.theme.Body2(v)
```

ui/statistics_page.go
```diff
@@ -0,0 +1,120 @@
...
+		item("Wallet data", pg.common.wallet.DataSize()),
+		pg.lineSeparator(),
+		item("Transactions", fmt.Sprintf("%d", (*pg.txs).Total)),
+		pg.lineSeparator(),
+		item("Wallets", fmt.Sprintf("%d", len(pg.common.info.Wallets))),
+	}
+
+	return card.Layout(gtx, func(gtx C) D {
+		m16 := values.MarginPadding16
+		return layout.Inset{Left: m16}.Layout(gtx, func(gtx C) D {
```
Sirmorrison:
```suggestion
		return layout.Inset{Left: values.MarginPadding16}.Layout(gtx, func(gtx C) D {
```

No need for the extra variable as it can fit into the Inset.

ui/statistics_page.go
```diff
@@ -0,0 +1,120 @@
...
+func (pg *statPage) Layout(gtx layout.Context) layout.Dimensions {
+	container := func(gtx C) D {
+		page := SubPage{
+			title: "Statistics",
+			back: func() {
+				pg.common.changePage(PageDebug)
+			},
+			body: func(gtx C) D {
+				return pg.layoutStats(gtx)
+			},
```
Sirmorrison:
```suggestion
			body: pg.layoutStats,
```

Sirmorrison (dismissed):
The uptime on the mockup is different from what we have implemented. From the mockup, it looks like the up time is a timer, to determine how long the app has been running.
image

Also, you currently have `testnet3` on the `Network` it should read either `testnet` or `mainnet`
image

ui/statistics_page.go
oshorefueled: I don't think this change is necessary. The reason for the values package is already to be able to manage values from a single point if there's need for it. This is a local scope and the variable declaration to its use is pretty close, so it's not necessary. Besides the former has shorter lines of code than the suggestion.

ui/statistics_page.go
oshorefueled: time.Now() returns the current time, Uptime should be how long the app has been running.

oshorefueled (changes requested):

oshorefueled (dismissed):

oshorefueled (approved):

PR #450, opened 2021-05-31
General code improvements

beansgum:
This pr brings a bunch of changes to the code, mostly in the aspect of the code structure, how pages interact, and reusing codes. I noticed the use of global variables everywhere which can cause unintended nil pointer errors as the modification of these variables are not done in one source nor protected by a mutex. The pages aren't properly managed, in terms of the page lifecycle, state variables, and passing data between pages. The dcrlibwallet implementation is wrong in some places and there are also some instances of reimplementation of features already handled by dcrlibwallet.

This pr adds a new page, `MainPage`, and its job is to manage the app toolbar, navigation bar, and back stack history of other pages. This moves the independent management of the page navigation from the pages to `MainPage` which then provides functionality like `popPage` & `popToPage` that helps pages move back in history. A `showModal` function is also added to `MainPage` intended to replace the current `modalReceiver` channel. The `showModal` function allows displaying multiple modals at once which are stacked on each other and ordered with the latest modal on top. 

The account selector was restructured, the previous implementation made use of global variables and a shared instance for every single account selector use in the app. The new account selector allows an instance to be created, it has a configurable account filter that helps filter out accounts/wallets for specific usage, and it also handles displaying the modal when tapped with the help of the `showModal` function.   

The use of multi-wallet proxy functions(`commands.go`) is removed from most parts of the app as it makes use of global variables & channels to deliver results that can be gotten by directly calling the multi-wallet function. This also caused some page changes especially to the ones that required the results globally from multi-wallet. Local variables
were added for required pages and their values are refreshed when needed.


dreacot:
1. Clicking the 'see all' button causes the app to crash
Screenshot 2021-05-31 at 5 38 02 PM

2. Clicking confirm on the modal when trying to view seed phrase causes the app to crash
Screenshot 2021-05-31 at 5 43 44 PM

3. While trying to setup mixer, after inputting your password and clicking confirm in the modal displayed below, the app crashes
Screenshot 2021-05-31 at 5 45 31 PM

Sirmorrison:
This PR no longer applies as it is now redundant. Closing.

PR #459, opened 2021-06-04
Add main page 

beansgum:
This pr adds a new page, `MainPage`, and its job is to manage the app toolbar, navigation bar, and back stack history of other pages and it also moves the independent management of the page navigation from the pages to `MainPage`.

### DIFF
--- a/ui/wallet_page.go
+++ b/ui/wallet_page.go
@@ -25,7 +25,7 @@ const PageWallet = "Wallets"
 type menuItem struct {
 	text     string
 	button   *widget.Clickable
-	action   func(pageCommon)
+	action   func(*pageCommon)
 	separate bool
 }
 
@@ -36,7 +36,7 @@ type collapsible struct {
 }
 
 type walletPage struct {
-	common                                     pageCommon
+	common                                     *pageCommon
 	walletInfo                                 *wallet.MultiWalletInfo
 	wallet                                     *wallet.Wallet
 	walletAccount                              **wallet.Account
@@ -68,7 +68,7 @@ type walletPage struct {
 	refreshPage                                *bool
 }
 
-func (win *Window) WalletPage(common pageCommon) Page {
+func (win *Window) WalletPage(common *pageCommon) Page {
 	pg := &walletPage{
 		common:                   common,
 		walletInfo:               win.walletInfo,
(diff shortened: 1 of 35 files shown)


dreacot:
1. Navigating to the tickets page crashes the app

2. Clicking on the wallet selector doesn't work on both the send and receive page
Screenshot 2021-06-04 at 11 27 19 AM

3. clicking on the send page, then the receive page right after, the x button closes the receive page but not the send page
Screenshot 2021-06-04 at 11 31 56 AM

4. clicking clear all fields on the send page only clears the amount and not the address
Screenshot 2021-06-04 at 11 34 00 AM

5. clicking "got it" on the modal that displays on the security tools page does nothing
Screenshot 2021-06-04 at 11 37 48 AM

6. clicking the edit button on the account details page causes the app to crash
Screenshot 2021-06-04 at 11 40 57 AM

7. clicking account[1] for my second wallet, shows the the account details of account[1] for my first wallet

8. The toast message doesn't disappear
Screenshot 2021-06-04 at 11 45 20 AM

ui/window.go
```diff
@@ -99,20 +99,18 @@ func CreateWindow(wal *wallet.Wallet, decredIcons map[string]image.Image, collec
...
 
 	return win, nil
 }
 
 func (win *Window) changePage(page string) {
-	win.pages[win.current].onClose()
-	win.current = page
+	_ = page
 	win.refresh()
 }
```
oshorefueled: This method doesn't seem to be doing anything, it can be removed. The win.refresh() method isn't needed as well. Each page can handle invalidating the page by calling op.InvalidateOp. The changePageAndRefresh doesn't seem to be doing anything either.

ui/wallet_page.go
```diff
@@ -280,16 +278,14 @@ func (pg *walletPage) showImportWatchOnlyWalletModal(common pageCommon) {
...
 func (pg *walletPage) Layout(gtx layout.Context) layout.Dimensions {
 	common := pg.common
-	if *pg.refreshPage {
-		common.refreshWindow()
-		*pg.refreshPage = false
-	}
+	// if *pg.refreshPage {
+	// 	common.refreshWindow()
+	// 	*pg.refreshPage = false
+	// }
```
oshorefueled, line 284: This can be deleted if it's not needed.

oshorefueled (changes requested):

oshorefueled:
The top bar doesn't seem to be updated until I restart the app. I changed the currency conversion option in the settings page and it doesn't get displayed on the top bar until I restart the app.

oshorefueled (approved):

PR #460, opened 2021-06-05
Add modal interface and migrate existing modals

beansgum:
This pr adds a `Modal` interface, an interface for creating and managing custom modals made to replace the current modal manager which uses a global instance for all modals.  A `showModal` & `dismissModal` function is also added to MainPage and made available to all pages via `pageCommon`l. The `showModal` function allows displaying multiple modals at once which are stacked on each other and ordered with the latest modal on top and `dismissModal` helps with dismissing a specific modal.


dreacot (dismissed):

ui/create_watch_only_modal.go
```diff
@@ -0,0 +1,146 @@
...
+
+func (cm *createWatchOnlyModal) setError(err string) {
+	if err == "" {
+		cm.extendedPubKey.ClearError()
+	} else {
+		cm.extendedPubKey.SetError(err)
+	}
+}
+
+func (cm *createWatchOnlyModal) callbackFunc(callback func(walletName, extPubKey string, m *createWatchOnlyModal) bool) *createWatchOnlyModal {
```
oshorefueled: a better name for this would be `watchOnlyCreated` or `walletCreated`. You did something similar here https://github.com/planetdecred/godcr/pull/460/files#diff-a1dc465a2f07bd3ed4f26a53ae45bf1edd88004d5af4b2353493cb1b8e0cf14dR106

ui/modal.go
```diff
@@ -0,0 +1,13 @@
+package ui
+
+import (
+	"gioui.org/layout"
+)
+
+type Modal interface {
```
oshorefueled, line 7: dismiss and show seems to be a common method on every modal and should be defined on the interface

ui/text_input_modal.go
```diff
@@ -0,0 +1,98 @@
...
+
+const ModalTextInput = "text_input_modal"
+
+type textInputModal struct {
+	*infoModal
+
+	isLoading bool
+
+	textInput decredmaterial.Editor
+	callback  func(string, *textInputModal) bool
```
oshorefueled, line 19: a comment should probably be added that the callback here and in other modals where they appear would be optional on some modals and the main aim of the modal is to have it handle their intended functions within the modal.

oshorefueled (changes requested):

oshorefueled (approved):

PR #467, opened 2021-06-13
Fix display the account name in the Inputs and the Outputs of the transaction

tuhalang:
The brackets () before the wallet name(Wallet-Main) should display the account name that owns the address or 'external' if the address does not belong to the wallet. But right now, it is empty as shown
image

So this PR will display the account name or 'external' if the address does not belong to the wallet
image

### DIFF
--- a/ui/transaction_details_page.go
+++ b/ui/transaction_details_page.go
@@ -91,7 +91,7 @@ func (pg *transactionDetailsPage) Layout(gtx layout.Context) layout.Dimensions {
 						return pg.separator(gtx)
 					},
 					func(gtx C) D {
-						return pg.txnInputs(gtx)
+						return pg.txnInputs(gtx, common)
 					},
 					func(gtx C) D {
 						return pg.separator(gtx)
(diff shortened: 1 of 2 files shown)


ui/transaction_details_page.go
```diff
@@ -300,7 +300,10 @@ func (pg *transactionDetailsPage) txnInputs(gtx layout.Context) layout.Dimension
 	collapsibleBody := func(gtx C) D {
 		return pg.transactionInputsContainer.Layout(gtx, len(transaction.Txn.Inputs), func(gtx C, i int) D {
 			amount := dcrutil.Amount(transaction.Txn.Inputs[i].Amount).String()
-			acctName := fmt.Sprintf("(%s)", transaction.AccountName)
+			walletID := transaction.Txn.WalletID
+			accountNumber := transaction.Txn.Inputs[i].AccountNumber
+			accountName := common.wallet.GetAccountName(walletID, accountNumber)
+			acctName := fmt.Sprintf("(%s)", accountName)
 			walName := transaction.WalletName
 			hashAcct := transaction.Txn.Inputs[i].PreviousOutpoint
```
oshorefueled: This looks like a duplicate of what's in txOutputs. Those values look like they could be used directly in `txIORow method by just passing it the transaction as an argument.

oshorefueled (changes requested):

oshorefueled (approved):

PR #475, opened 2021-06-17
Use dcrlibwallet directly in overview and transactions page

beansgum:
The change removes the use of global transactions data to populate the layout of overview, transactions and transaction details page. The transactions are now fetched from dcrlibwallet when the page is initially loaded and the list is refreshed whenever a transaction notification is received.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,4 +25,7 @@ require (
 	golang.org/x/text v0.3.3
 )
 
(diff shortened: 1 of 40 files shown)


oshorefueled:
Clicking on one clickable text on the transaction details page triggers a click on another

transactions bug

There's also a slight navigation delay when navigating from any page to the overview page

oshorefueled (approved):

PR #484, opened 2021-06-20
Make account selector instance reusable

beansgum:
This pr changes the how the account selector is used and managed by pages. The previous implementation had all pages share one instance of account selector that is managed globally and is configured by tweaking global variables. This fixes that by allowing pages to create instances of account selector, allows configurable account filter helpful for filtering out privacy account and also triggers a callback function when the selected account is changed.


ui/account_selector.go
```diff
@@ -0,0 +1,486 @@
...
+	selectedAccount    *dcrlibwallet.Account
+	selectedWalletName string
+	totalBalance       string
+}
+
+func newAccountSelector(common *pageCommon) *accountSelector {
+
+	return &accountSelector{
+		pageCommon:         common,
+		accountIsValid:     func(a *dcrlibwallet.Account) bool { return true },
```
oshorefueled: There's no need to declare a variable name for the parameter of type *dcrlibwallet.Account in the function. It is not used anywhere.

oshorefueled (changes requested):

oshorefueled (approved):

PR #487, opened 2021-06-24
add custom switch widget

### LINKED ISSUE #478: custom material switch widget
we're currently using the default material design from gio material for our switch/toggle widgets. This isn't in conformity with our mockup designs and so a custom switch widget in decred material should be created. 

#### Current Implementation
Screenshot 2021-06-18 at 15 14 19

#### Mockup Designs
Screenshot 2021-06-18 at 15 15 12

Screenshot 2021-06-18 at 15 15 21

It should be noted that the colours used should be the same as specified in the mockups.

Sirmorrison:
Fix #478 
This PR implements custom switch widget.

### DIFF
--- a/ui/decredmaterial/switch.go
+++ b/ui/decredmaterial/switch.go
@@ -17,11 +17,13 @@ import (
 )
 
 type Switch struct {
-	active       color.NRGBA
-	inactive     color.NRGBA
-	thumbColor   color.NRGBA
-	disabled     bool
-	switchWidget *widget.Bool
+	active     color.NRGBA
+	inactive   color.NRGBA
+	thumbColor color.NRGBA
+	disabled   bool
+	value      bool
+	changed    bool
+	clk        *widget.Clickable
 }
 
 type SwitchItem struct {
@@ -39,7 +41,7 @@ type SwitchButtonText struct {
 
 func (t *Theme) Switch() *Switch {
 	sw := &Switch{
-		switchWidget: new(widget.Bool),
+		clk: new(widget.Clickable),
 	}
 	sw.active, sw.inactive, sw.thumbColor = t.Color.Primary, t.Color.InactiveGray, t.Color.Surface
 	return sw
@@ -82,10 +84,6 @@ func (s *Switch) Layout(gtx layout.Context) layout.Dimensions {
 	}}
 
 	col := s.inactive
-	if s.disabled {
-		col = Disabled(s.inactive)
-		s.thumbColor = Disabled(s.thumbColor)
-	}
 	if s.IsChecked() {
 		col = s.active
 	}
@@ -133,7 +131,7 @@ func (s *Switch) Layout(gtx layout.Context) layout.Dimensions {
 	sz := image.Pt(clickSize, clickSize)
 	pointer.Ellipse(image.Rectangle{Max: sz}).Add(gtx.Ops)
 	gtx.Constraints.Min = sz
-	s.switchWidget.Layout(gtx)
+	s.clk.Layout(gtx)
 	stack.Load()
 
 	dims := image.Point{X: trackWidth, Y: thumbSize}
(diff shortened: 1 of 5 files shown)


ui/decredmaterial/switch.go
```diff
@@ -3,16 +3,26 @@
 package decredmaterial
 
 import (
+	"image"
 	"image/color"
 
+	"gioui.org/f32"
+	// "gioui.org/internal/f32color"
```
beansgum: Cleanup imports

ui/decredmaterial/switch.go
```diff
@@ -3,16 +3,26 @@
...
-	"gioui.org/widget/material"
+	// "gioui.org/widget/material"
 )
 
 type Switch struct {
-	material.SwitchStyle
+	Enabled    color.NRGBA
+	Disabled   color.NRGBA
+	thumbColor color.NRGBA
+	Switch     *widget.Bool
```
beansgum: I guess this holds value for the active state of the switch, a better variable name would be `isChecked.

ui/decredmaterial/switch.go
```diff
@@ -3,16 +3,26 @@
...
+	"gioui.org/op"
+	"gioui.org/op/clip"
+	"gioui.org/op/paint"
 	"gioui.org/unit"
 	"gioui.org/widget"
-	"gioui.org/widget/material"
+	// "gioui.org/widget/material"
 )
 
 type Switch struct {
```
beansgum, line 19: Make the variables in this struct private and add functions to set them.

beansgum, line 19: This widget needs a `Changed()` event for when the enabled state of the switch is changed.

ui/decredmaterial/switch.go
```diff
@@ -28,8 +38,12 @@ type SwitchButtonText struct {
 	selected                           int
 }
 
-func (t *Theme) Switch(swtch *widget.Bool) Switch {
-	return Switch{material.Switch(t.Base, swtch)}
+func (t *Theme) Switch(swtch *widget.Bool) *Switch {
```
beansgum: Make this a primitive bool variable, handle the `*widget.Bool` here and provide a `Changed()` function.
```suggestion
func (t *Theme) Switch(isChecked bool) *Switch {
```

ui/decredmaterial/switch.go
```diff
@@ -28,8 +38,12 @@ type SwitchButtonText struct {
...
+	sw := &Switch{
+		Switch: swtch,
+	}
+	sw.Enabled, sw.Disabled, sw.thumbColor = t.Color.Primary, t.Color.InactiveGray, t.Color.Surface
```
beansgum: Disabled and inactive states are mixed up here. Add an inactive state for when the switch is inactive and a disabled state for when user interaction is disabled on the switch.

beansgum (changes requested):
Can a disabled state be added to this widget where it'll ignore all user-triggered events?

ui/decredmaterial/switch.go
oshorefueled: The Switch widget should really just be a skin that accepts an instance of the bool type. This means it shouldn't return an object. Similar to the Button widget that accepts an instance of the Clickable type, the Switch widget should only need to accept an instance of the Bool type.  

To use the switch widget in this current state, we'll need to keep an instance of both the Switch type and the bool type which is quite redundant.

oshorefueled (changes requested):

ui/decredmaterial/switch.go
```diff
@@ -54,6 +67,106 @@ func (t *Theme) SwitchButtonText(i []SwitchItem) *SwitchButtonText {
...
+	return layout.Dimensions{Size: dims}
+}
+
+func (s *Switch) Changed() bool {
+	return s.switchWidget.Changed()
+}
+
+func (s *Switch) IsChecked() bool {
+	if s.disabled {
+		s.SetChecked(false)
```
beansgum: Why is this value set in a read-only function? Also it's possible for a switch to be checked while in a disabled state. The disabled state simply means no user interaction is allowed.

beansgum (changes requested):

beansgum (approved):

oshorefueled (approved):

PR #491, opened 2021-06-25
window: maintain pages history and add function to go to previous page

### LINKED ISSUE #489: Return button not active on create restore page
On the create restore page, if the back button is pressed, nothing happens. Looking through the codes, it was observed that this has not yet been implemented.

IMG_20210625_023105

beansgum:
This adds a `popPage` function that changes the previous window page when called and returns true if the page was changed. A `keepBackStack` parameter is also added to `changePage` function to configure that pages should be in the back stack history.
Closes #489

### DIFF
--- a/ui/create_restore_page.go
+++ b/ui/create_restore_page.go
@@ -639,6 +639,8 @@ func (pg *createRestore) handle() {
 			pg.resetSeeds()
 			pg.resetPasswords()
 
+			// Go back to wallets page if there's more than one wallet
+			// or launch main page.
 			if pg.common.multiWallet.LoadedWalletsCount() > 1 {
 				pg.common.popWindowPage()
 			} else {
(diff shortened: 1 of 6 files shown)


ui/create_restore_page.go
```diff
@@ -642,8 +640,13 @@ func (pg *createRestore) handle() {
...
 
-			pg.common.wallet.SetupListeners()
-			pg.common.changeWindowPage(newMainPage(pg.common))
+			if pg.common.multiWallet.LoadedWalletsCount() > 1 {
+				pg.common.popWindowPage()
+			} else {
+				pg.common.wallet.SetupListeners()
+				pg.common.changeWindowPage(newMainPage(pg.common), false)
+			}
+
```
Sirmorrison: remove space here and line 624.

Sirmorrison (changes requested):

Sirmorrison (dismissed):

oshorefueled (approved):

PR #492, opened 2021-06-25
Remove global subpage buttons and declare them locally

beansgum:
This fixes a bug where the sub-page buttons of a page can be overwritten from another page when the page is changed.


Sirmorrison (dismissed):
Tested and PR looks good.

oshorefueled (dismissed):

oshorefueled (approved):

dreacot (approved):

PR #514, opened 2021-07-04
Make seed backup notification row clickable

### LINKED ISSUE #505: make entire rows clickable
1. the entire backup warning row should be clickable and not just the arrow icon
Screenshot 2021-06-30 at 4 26 42 PM

2. The entire "Add new account" row should be clickable and not just the "+" icon
Screenshot 2021-06-30 at 4 37 14 PM

JustinBeBoy:
Resolve #505 
This PR allows:
 - Make backup warning row clickable
 - Make add account row clickable

### DIFF
--- a/ui/page/wallet_page.go
+++ b/ui/page/wallet_page.go
@@ -30,9 +30,9 @@ type walletListItem struct {
 	accountsList *decredmaterial.ClickableList
 
 	// normal wallets
-	collapsible   *decredmaterial.CollapsibleWithOption
-	addAcctBtn    decredmaterial.IconButton
-	backupAcctBtn decredmaterial.IconButton
+	collapsible         *decredmaterial.CollapsibleWithOption
+	addAcctClickable    *widget.Clickable
+	backupAcctClickable *widget.Clickable
 
 	// watch only
 	moreButton decredmaterial.IconButton
@@ -70,6 +70,8 @@ type WalletPage struct {
 	watchOnlyWalletIcon      *widget.Image
 	shadowBox                *decredmaterial.Shadow
 	separator                decredmaterial.Line
+	addAcctIcon              *widget.Icon
+	backupAcctIcon           *widget.Icon
 }
 
 func NewWalletPage(l *load.Load) *WalletPage {
@@ -85,9 +87,12 @@ func NewWalletPage(l *load.Load) *WalletPage {
 		openPopupIndex:           -1,
 		shadowBox:                l.Theme.Shadow(),
 		separator:                l.Theme.Separator(),
+		addAcctIcon:              l.Icons.ContentAdd,
+		backupAcctIcon:           l.Icons.NavigationArrowForward,
 	}
 
 	pg.separator.Color = l.Theme.Color.Background
+	pg.addAcctIcon.Color = l.Theme.Color.Text
 
 	pg.watchOnlyWalletLabel = pg.Theme.Body1(values.String(values.StrWatchOnlyWallets))
 	pg.watchOnlyWalletLabel.Color = pg.Theme.Color.Gray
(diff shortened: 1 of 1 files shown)


Sirmorrison:
* Testing out this PR, i noticed that i have tried creating 3 different account in my 3 different wallets, and only one account showed up (in the last wallet). and that happened after i exited and restarted the application. I have fully synced my wallets, yet the other two are still not displayed.

* Tested this on master and the accounts where created successfully, however they only displayed when i restart the application. **You should fix the issue with the wallet account creation and also fix the issue with the account displaying on the UI only after the application is restarted instead of having them display on the UI after they are created**.

* Not so sure about the approach you used to implement the clickable row though, however, i think a better approach would have been to use the approach used in the settings page.
https://github.com/planetdecred/godcr/blob/c31798a81ccd2943430f24e7e491edc4a3cd8f1a/ui/page/settings_page.go#L337

* You may need to extend the clickable row method so it can be used across multiple pages.

Sirmorrison (changes requested):
See comment above

oshorefueled:
> * Testing out this PR, i noticed that i have tried creating 3 different account is my 3 different wallets, and only one account showed up (the last wallet). and that happened after i exited and restarted the application. I have fully synced my wallets, yet the other two are still not displayed.
> * Tested this on master and the accounts where created, however only displayed when i restart the application. What it should be is to have the accounts displayed in the UI after they are created.

Kindly list out the steps to reproduce the error you're experiencing. If the error isn't peculiar to this PR, kindly create an issue for it also listing out the steps to reproduce the error and it would be fixed on another PR. The same thing applies to your other comments, thank you.

ui/page/wallet_page.go
```diff
@@ -694,9 +698,13 @@ func (pg *WalletPage) walletAccountsLayout(gtx layout.Context, account *dcrlibwa
...
 }
 
-func (pg *WalletPage) backupSeedNotification(gtx layout.Context, listItem *walletListItem) layout.Dimensions {
+func (pg *WalletPage) backupSeedNotification(gtx layout.Context, index int) layout.Dimensions {
 	gtx.Constraints.Min.X = gtx.Constraints.Max.X
 	textColour := pg.Theme.Color.InvText
+	click := pg.backupAcct[index]
+	pointer.Rect(image.Rectangle{Max: gtx.Constraints.Max}).Add(gtx.Ops)
+	click.Add(gtx.Ops)
+	pg.openAcctBackup(gtx, click)
```
oshorefueled: rather than handling click operations from the layout code, you can create a clickableRow widget that handles the click operations and returns a Clicked(). A clickableRow widget would also be useful in the settings page.

oshorefueled (changes requested):

ui/page/wallet_page.go
```diff
@@ -694,42 +694,47 @@ func (pg *WalletPage) walletAccountsLayout(gtx layout.Context, account *dcrlibwa
 	)
 }
 
-func (pg *WalletPage) backupSeedNotification(gtx layout.Context, listItem *walletListItem) layout.Dimensions {
+func (pg *WalletPage) backupSeedNotification(gtx layout.Context, index int) layout.Dimensions {
```
beansgum: Pass the walletListItem not index.
```suggestion
func (pg *WalletPage) backupSeedNotification(gtx layout.Context, listItem *walletListItem) layout.Dimensions {
```

ui/page/wallet_page.go
```diff
@@ -775,6 +780,12 @@ func (pg *WalletPage) layoutAddWalletSection(gtx layout.Context) layout.Dimensio
 	})
 }
 
+func (pg *WalletPage) clickableRow(gtx layout.Context, clickable *widget.Clickable, body layout.Widget) layout.Dimensions {
```
beansgum: Nothing in this function makes it a 'clickable row', it's just a wrapper for decredmaterial clickable. It should be removed and decredmaterial clickable should be used directly.

ui/page/wallet_page.go
```diff
@@ -694,42 +694,47 @@ func (pg *WalletPage) walletAccountsLayout(gtx layout.Context, account *dcrlibwa
...
-			layout.Flexed(1, func(gtx C) D {
-				inset := layout.Inset{
-					Top: values.MarginPadding5,
-				}
-				return inset.Layout(gtx, func(gtx C) D {
-					return layout.E.Layout(gtx, listItem.backupAcctBtn.Layout)
-				})
-			}),
-		)
+	return pg.clickableRow(gtx, pg.listItems[index].backupAcctClickable, func(gtx C) D {
```
beansgum: Remove redundant wrapper function
```suggestion
	return decredmaterial.Clickable(gtx, pg.listItems[index].backupAcctClickable, func(gtx C) D {
```

ui/page/wallet_page.go
```diff
@@ -433,21 +427,27 @@ func (pg *WalletPage) walletSection(gtx layout.Context) layout.Dimensions {
...
-										Left:  values.MarginPadding38,
-									}.Layout(gtx, listItem.addAcctBtn.Layout)
-								}),
-								layout.Rigid(func(gtx C) D {
-									txt := pg.Theme.H6(values.String(values.StrAddNewAccount))
-									txt.Color = pg.Theme.Color.Gray
-									return txt.Layout(gtx)
-								}),
-							)
+						return pg.clickableRow(gtx, pg.listItems[i].addAcctClickable, func(gtx C) D {
```
beansgum: Remove redundant wrapper function
```suggestion
						return decredmaterial.Clickable(gtx, pg.listItems[i].addAcctClickable, func(gtx C) D {
```

beansgum (changes requested):

beansgum (approved):

oshorefueled (approved):

Sirmorrison (approved):

Sirmorrison:
When an account is created, they are only displayed after the application is restarted. 
I will be creating an issue for this.

PR #525, opened 2021-07-14
Resolve data races

### LINKED ISSUE #455: logging data races
Enabling the race detector shows there are a lot of races happening around the logging, and this is the cause:

```
// Write writes the data in p to standard out and the log rotator.
func (l logWriter) Write(p []byte) (n int, err error) {
        os.Stdout.Write(p)
        go func() {
                internalLog <- string(p)
        }()
        return logRotator.Write(p)
}
```

The byte slice passed to Write is being modified later on, because the slog package will recycle and reuse the byte slice to minimize allocations.  Asynchronously reading this slice in another goroutine spawned by Write is incorrect and unsafe.

oluwandabira:
Resolves #455

This resolves the issue of a data race in logging (by copying the bytes) but there are still other data races from wallet functionality.

### DIFF
--- a/log.go
+++ b/log.go
@@ -20,14 +20,18 @@ import (
 // the write-end pipe of an initialized log rotator.
 type logWriter struct{}
 
-var internalLog = make(chan string, 10)
+var (
+	internalLog = make(chan string, 10)
+)
 
 // Write writes the data in p to standard out and the log rotator.
 func (l logWriter) Write(p []byte) (n int, err error) {
+	str := make([]byte, len(p))
+	copy(str, p)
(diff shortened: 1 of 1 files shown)


jrick:
This fixes the data race, but the writes to the internalLog channel can still be made in a different order than how the log messages were originally logged, because there is no synchronization between each of these writes (they just happen in a newly spawned goroutine).  I can't tell you if that matters or not, but I suspect it does for someone, somewhere.

Sirmorrison:
This issue is resolved by PR #574.

PR #540, opened 2021-07-26
add in-app and system notifications 

### LINKED ISSUE #527: Implement toasts notifications on pages in the page package
Some pages have been moved to the page package and don't have access to the notification implementation made on the common struct. Notification implementation should be added for pages in the page package.

oshorefueled:
This PR resolves #527 

It implements and in-app notification (toast) which is used in displaying success or error messages to a user within the app in response to an action. It also implements a system notification which displays os specific system-wide applications. Some of the notifications displayed with the system notification could be new block notifications, receive transaction notifications, transaction confirmation notification e.t.c..

Calls to the CreateToast method used by pages in the page package are replaced with the Notify method created in this PR.


ui/notification/toast.go
```diff
@@ -0,0 +1,77 @@
...
+	return &Toast{
+		theme: th,
+	}
+}
+
+func (t *Toast) Notify(message string, success bool) {
+	t.message = message
+	t.success = success
+	t.timer = time.NewTimer(time.Second * 3)
+	t.show = true
```
beansgum: Protect these variables with a mutex.

ui/notification/toast.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (t *Toast) Layout(gtx layout.Context) layout.Dimensions {
+	if !t.show {
```
beansgum: Remove the show variable and replace this condition with `t.timer == nil`.

ui/notification/toast.go
```diff
@@ -0,0 +1,77 @@
...
+		return layout.Dimensions{}
+	}
+
+	t.handleToastDisplay(gtx)
```
beansgum: This check should be done at the top of the function.

ui/notification/log.go
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2017, The dcrdata developers
```
beansgum: This package does not need its own logger.

ui/notification/system.go
```diff
@@ -0,0 +1,35 @@
...
+
+	return &SystemNotification{
+		iconPath: filepath.Join(absolutePath, icon),
+	}
+}
+
+func (s *SystemNotification) Notify(message string) {
+	err := beeep.Notify(title, message, s.iconPath)
+	if err != nil {
+		log.Info("could not initiate desktop notification, reason:", err.Error())
```
beansgum: This error should be returned not logged.

ui/notification/utils.go
```diff
@@ -0,0 +1,22 @@
+package notification
+
+import (
+	"fmt"
+	"os"
+	"path"
+	"path/filepath"
+)
+
+func getAbsolutePath() (string, error) {
```
beansgum: I don't think any utils.go file is needed in this package since it's just this function and it's used only in system.go

beansgum (changes requested):

ui/notification/system.go
oshorefueled: The log avoids handling notification errors in the ui code. All that needs to be called is the Notify method, just like toast notifications.

ui/notification/toast.go
oshorefueled: moving the handle function to the top of the Layout method would lead to a nil dereference error.

beansgum (approved):

Sirmorrison (approved):
Tested and looks good.

PR #545, opened 2021-08-02
Sign message bug

### LINKED ISSUE #543: Sign message button bug
When the sign message page is opened without moving the mouse, the `sign message` button is shown as active. However, when the mouse is moved, the page colors tend to update.

Also, when the sign message button is clicked with all inputs empty, there is a flashing error that does not stay. see screen shots below.
ezgif com-gif-maker(4)

Also, when there is an error from the address editor, the sign message button is active, which should not be so.
image

Sirmorrison:
Close #543 

This pr resolve issue with 
* fix active sign message button color when page is opened
* fix active sign message button color when error exist
* fix issue with flashing error message

### DIFF
--- a/ui/page/sign_message_page.go
+++ b/ui/page/sign_message_page.go
@@ -20,7 +20,6 @@ const SignMessagePageID = "SignMessage"
 
 type SignMessagePage struct {
 	*load.Load
-	theme     *decredmaterial.Theme
 	container layout.List
 	wallet    *dcrlibwallet.Wallet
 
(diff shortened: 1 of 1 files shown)


ui/page/sign_message_page.go
```diff
@@ -252,54 +271,69 @@ func (pg *SignMessagePage) Handle() {
...
+	pg.messageEditor.SetError("")
 	pg.signedMessageLabel.Text = ""
 	pg.errorLabel.Text = ""
+	pg.isValidatedMsg = false
+	pg.isValidatedAddress = false
 }
 
-func (pg *SignMessagePage) OnClose() {}
+func (pg *SignMessagePage) OnClose() {
+	pg.clearForm()
```
beansgum: This isn't required since the page isn't data persisted when closed.

ui/page/sign_message_page.go
```diff
@@ -20,11 +21,12 @@ const SignMessagePageID = "SignMessage"
 
 type SignMessagePage struct {
 	*load.Load
-	theme     *decredmaterial.Theme
 	container layout.List
 	wallet    *dcrlibwallet.Wallet
 
 	isSigningMessage                           bool
+	isValidatedMsg                             bool
+	isValidatedAddress                         bool
```
beansgum: Proper name should be `hasValidatedMsg` or `addressIsValid`

ui/page/sign_message_page.go
```diff
@@ -252,54 +271,69 @@ func (pg *SignMessagePage) Handle() {
 	}
 }
 
-func (pg *SignMessagePage) validate(ignoreEmpty bool) bool {
-	isAddressValid := pg.validateAddress(ignoreEmpty)
-	isMessageValid := pg.validateMessage(ignoreEmpty)
+func (pg *SignMessagePage) validate() bool {
+	isAddressValid := pg.validateAddress()
+	isMessageValid := pg.validateMessage()
```
beansgum:
```suggestion
	addressIsValid := pg.validateAddress()
	messageIsValid := pg.validateMessage()
```

ui/page/sign_message_page.go
beansgum: These two variables are modified in several references but only used once and they were used together. Instead of two variables, make it one and it should be `true` if message and address is valid.

beansgum (changes requested):

beansgum (approved):

oshorefueled (approved):

PR #550, opened 2021-08-06
fix mixer toggle button bug

### LINKED ISSUE #549: StakeShuffle Page bug.
On the StakeShuffle page, when the start mixer toggle is clicked, the toggle becomes active... However, if you exit and return to the StakeShuffle page, the toggles goes back to an inactive state while mixer is still running.
For some reason, the toggle state is not saved.
Screenshot 2021-08-03 at 07 56 40

Screenshot 2021-08-03 at 07 58 55

Lastly, at the danger zone section, if the allow spending from unmixed account is toggled and the page is exited, the toggle state returns to an inactive state. 
This is an issue because the user would not know if this is active or not.

Sirmorrison:
Close #549 

* Set the mixer toggle to active when mixer is running.
* Disabled spend from unmixed account toggle as it is not yet implemented.

### DIFF
--- a/ui/page/privacy_page.go
+++ b/ui/page/privacy_page.go
@@ -44,16 +44,15 @@ func NewPrivacyPage(l *load.Load, wallet *dcrlibwallet.Wallet) *PrivacyPage {
 	}
 	pg.toPrivacySetup.Background = l.Theme.Color.Primary
 	pg.backButton, pg.infoButton = components.SubpageHeaderButtons(l)
-	pg.allowUnspendUnmixedAcct.Disabled()
 	return pg
 }
 
 func (pg *PrivacyPage) OnResume() {
-
+	pg.toggleMixer.SetChecked(pg.wallet.IsAccountMixerActive())
+	pg.allowUnspendUnmixedAcct.Disabled()
 }
 
 func (pg *PrivacyPage) Layout(gtx layout.Context) layout.Dimensions {
-	pg.updateMixerOptions()
 	d := func(gtx C) D {
 		sp := components.SubPage{
 			Load:       pg.Load,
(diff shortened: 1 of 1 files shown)


ui/page/privacy_page.go
```diff
@@ -436,4 +436,8 @@ func (pg *PrivacyPage) showModalPasswordStartAccountMixer() {
 		}).Show()
 }
 
+func (pg *PrivacyPage) updateMixerOptions() {
+	pg.toggleMixer.SetChecked(pg.wallet.IsAccountMixerActive())
```
oshorefueled: does this need to be wrapped in a method?

ui/page/privacy_page.go
```diff
@@ -54,6 +53,7 @@ func (pg *PrivacyPage) OnResume() {
 }
 
 func (pg *PrivacyPage) Layout(gtx layout.Context) layout.Dimensions {
+	pg.updateMixerOptions()
```
oshorefueled: This should be in the OnResume method

ui/page/privacy_page.go
```diff
@@ -17,35 +14,37 @@ import (
...
-		allowUnspendUnmixedAcct: new(widget.Bool),
+		toggleMixer:             l.Theme.Switch(),
+		allowUnspendUnmixedAcct: l.Theme.Switch(),
 		toPrivacySetup:          l.Theme.Button(new(widget.Clickable), "Set up mixer for this wallet"),
 		dangerZoneCollapsible:   l.Theme.Collapsible(),
 	}
 	pg.toPrivacySetup.Background = l.Theme.Color.Primary
 	pg.backButton, pg.infoButton = components.SubpageHeaderButtons(l)
-
+	pg.allowUnspendUnmixedAcct.Disabled()
```
oshorefueled: This would probably be in the OnResume method

oshorefueled (changes requested):

beansgum (approved):

oshorefueled (approved):

PR #559, opened 2021-08-16
Fix transaction row highlight when pressed

beansgum:
The transaction row does not highlight the proper areas when pressed, this pr fixes the bug.
Before changes:

After after:

### DIFF
--- a/ui/decredmaterial/linearlayout.go
+++ b/ui/decredmaterial/linearlayout.go
@@ -25,6 +25,7 @@ type LinearLayout struct {
 	Direction   layout.Direction
 }
 
+// Layout2 displays a linear layout with a single child.
 func (ll LinearLayout) Layout2(gtx C, wdg layout.Widget) D {
 	return ll.Layout(gtx, layout.Rigid(wdg))
 }
--- a/ui/decredmaterial/linearlayout.go
+++ b/ui/decredmaterial/linearlayout.go
@@ -25,6 +25,10 @@ type LinearLayout struct {
 	Direction   layout.Direction
 }
 
+func (ll LinearLayout) Layout2(gtx C, wdg layout.Widget) D {
+	return ll.Layout(gtx, layout.Rigid(wdg))
+}
+
 func (ll LinearLayout) Layout(gtx C, children ...layout.FlexChild) D {
 	// draw margin
 	return ll.Margin.Layout(gtx, func(gtx C) D {
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"gioui.org/layout"
-	"gioui.org/text"
 	"gioui.org/unit"
 
 	"github.com/ararog/timeago"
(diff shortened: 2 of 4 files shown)


Sirmorrison:
image

image

image

The dimensions on the recent transactions table is off.

Also, there is a gray line immediately after the side Nav. (second screenshot... red section)

ui/page/components/components.go
```diff
@@ -117,110 +116,68 @@ func LayoutBalance(gtx layout.Context, l *load.Load, amount string) layout.Dimen
...
-	} else if row.Index == 0 {
-		// todo: remove top margin from container
-		directionIconTopMargin = values.MarginPadding0
-	}
 
 	wal := l.WL.MultiWallet.WalletWithID(row.Transaction.WalletID)
 
-	return layout.Inset{Top: values.MarginPadding16}.Layout(gtx, func(gtx C) D {
-		return layout.Flex{Axis: layout.Horizontal}.Layout(gtx,
+	return layout.Inset{Top: values.MarginPadding16, Bottom: values.MarginPadding16}.Layout(gtx, func(gtx C) D {
```
Sirmorrison: The padding used on the recent transaction page is not `values.MarginPadding16` its `values.MarginPadding5`
image

this is the reason we currently have excess padding on the recent transactions page.

Sirmorrison (changes requested):

Sirmorrison (dismissed):

ui/page/components/components.go
```diff
@@ -117,113 +116,80 @@ func LayoutBalance(gtx layout.Context, l *load.Load, amount string) layout.Dimen
...
-				icon.Scale = 1.0
+	return decredmaterial.LinearLayout{
+		Orientation: layout.Horizontal,
+		Width:       decredmaterial.MatchParent,
+		Height:      gtx.Px(values.MarginPadding56),
+		Direction:   layout.W,
+		Padding:     layout.Inset{Left: values.MarginPadding16, Right: values.MarginPadding16},
+	}.Layout(gtx,
+		layout.Rigid(func(gtx C) D {
+			gtx.Constraints.Min.Y = gtx.Constraints.Max.Y
```
oshorefueled, line 130: why not add this as a field in LinearLayout?

ui/decredmaterial/linearlayout.go
```diff
@@ -25,6 +25,10 @@ type LinearLayout struct {
 	Direction   layout.Direction
 }
 
+func (ll LinearLayout) Layout2(gtx C, wdg layout.Widget) D {
```
oshorefueled, line 29: Can we find a better name for this? If we can't how's it different from the Layout method, can we add a comment to it?

ui/page/components/components.go
```diff
@@ -117,113 +116,80 @@ func LayoutBalance(gtx layout.Context, l *load.Load, amount string) layout.Dimen
...
+		layout.Flexed(1, func(gtx C) D {
+			return layout.Flex{
+				Axis:      layout.Horizontal,
+				Spacing:   layout.SpaceStart,
+				Alignment: layout.Middle,
+			}.Layout(gtx,
+				layout.Rigid(func(gtx C) D {
+					return layout.Inset{Right: values.MarginPadding8}.Layout(gtx,
+						func(gtx C) D {
+							gtx.Constraints.Min.Y = gtx.Constraints.Max.Y
```
oshorefueled, line 171: Can this be part of the options in LinearLayout?

oshorefueled (changes requested):

ui/page/components/components.go
beansgum: I noticed that `layout.Rigid` doesn't inherit the constraint property of the parent layout.

oshorefueled (approved):

Sirmorrison (approved):

PR #560, opened 2021-08-16
Add different notification messages for stake transactions

beansgum:
- Place wallet name before notification message and hide if only one wallet exists.

Screenshot 2021-08-16 at 22 43 18

Screenshot 2021-08-16 at 22 28 43


Sirmorrison (approved):
Looks good

dreacot (approved):

PR #562, opened 2021-08-18
Fix top nav highlight bug when send or receive button is selected

### LINKED ISSUE #555: Receive page address card ui update
On the Receive page, the receive address card has rounded edges at the right hand side.
image

from the mock up, it should by square edged.
image

Also, when the receive page is selected, the top navigation is supposed to be highlighted.
image

But it is not in the app.
image

Sirmorrison:
Resolves #555 
The first issue with the address card has been resolved.
This PR resolves bug with Highlighting selected top Nav page button.

* Old UI
image

*New UI
image

### DIFF
--- a/ui/page/main_page.go
+++ b/ui/page/main_page.go
@@ -1,6 +1,7 @@
 package page
 
 import (
+	"image/color"
 	"strconv"
 
 	"gioui.org/layout"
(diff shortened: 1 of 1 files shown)


beansgum (changes requested):
The highlight ripple animation doesn't show when the buttons are pressed, unlike other clickables.

Sirmorrison:
> The highlight ripple animation doesn't show when the buttons are pressed, unlike other clickables.

After troubleshooting this request, it was observed that the reason for this was that the card widget used to highlight the currently selected tab was overlaying the button effect.
We will create a task for the button effect and fix this from another PR after #563 is merged.

beansgum (approved):

oshorefueled (approved):

Issue #571, opened 2021-08-20
Log page bug


Sirmorrison:
Currently on the log page, when the page is opened for the first time, logs shows up however, when you exit the page and return to it, all the logs missing.

<img width="797" alt="image" src="https://user-images.githubusercontent.com/27733432/130196710-5a8dc816-058b-41f4-96db-80ed07c5319f.png">

from my observation, it seems all the initial logs in the console are wiped out of the page and the page starts listening for new console logs.
<img width="798" alt="image" src="https://user-images.githubusercontent.com/27733432/130197045-4ab30606-e521-4533-93ec-4c3ceb474e44.png">

The expected behavior is to always have logs on the page both new and existing.

PR #574, opened 2021-08-20
Resolve Log Race 

### LINKED ISSUE #455: logging data races

devchoplife:
Ths PR resolves #455
This PR fixes the Log race issue and also resolves the disappeareance of logs on log page in #571

### DIFF
--- a/log.go
+++ b/log.go
@@ -21,13 +21,11 @@ import (
 // the write-end pipe of an initialized log rotator.
 type logWriter struct{}
 
-var internalLog = make(chan string, 10)
-
 // Write writes the data in p to standard out and the log rotator.
 func (l logWriter) Write(p []byte) (n int, err error) {
 	os.Stdout.Write(p)
 	go func() {
-		internalLog <- string(p)
+		//internalLog <- string(p)
 	}()
 	return logRotator.Write(p)
 }
--- a/log.go
+++ b/log.go
@@ -23,11 +23,7 @@ type logWriter struct{}
 
 // Write writes the data in p to standard out and the log rotator.
 func (l logWriter) Write(p []byte) (n int, err error) {
-	os.Stdout.Write(p)
-	go func() {
-		//internalLog <- string(p)
-	}()
-	return logRotator.Write(p)
+	return os.Stdout.Write(p)
 }
 
 // Loggers per subsystem.  A single backend logger is created and all subsytem
--- a/log.go
+++ b/log.go
@@ -23,7 +23,8 @@ type logWriter struct{}
 
 // Write writes the data in p to standard out and the log rotator.
 func (l logWriter) Write(p []byte) (n int, err error) {
-	return os.Stdout.Write(p)
+	os.Stdout.Write(p)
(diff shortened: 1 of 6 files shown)


log.go
```diff
@@ -21,13 +21,11 @@ import (
...
 type logWriter struct{}
 
-var internalLog = make(chan string, 10)
-
 // Write writes the data in p to standard out and the log rotator.
 func (l logWriter) Write(p []byte) (n int, err error) {
 	os.Stdout.Write(p)
 	go func() {
-		internalLog <- string(p)
+		//internalLog <- string(p)
```
Sirmorrison: if this is no longer needed, kindly remove the go routine.

ui/page/log_page.go
```diff
@@ -67,12 +66,17 @@ func (pg *LogPage) copyLogEntries(gtx C) {
 	}()
 }
 
-func (pg *LogPage) watchLogs(internalLog chan string) {
-	for l := range internalLog {
-		entry := l[:len(l)-1]
+func (pg *LogPage) watchLogs() {
+	//TODO
+	//add function to get log directory
+	logPath := "/home/devchoplife/.godcr/logs/godcr.log"
```
Sirmorrison: i am not sure i will be able to test with this path.

Sirmorrison (changes requested):
This is the display if i open the log page.
image

no logs are available as the `/home/devchoplife/.godcr/logs/godcr.log` does not exist in my machine.

image

The log path should not the be hard coded.

log.go
devchoplife: I added a //TODO there .
A function will be added later to get the log directory

ui/page/log_page.go
devchoplife: I added a //TODO there .
A function will be added later to get the log directory

Sirmorrison: When is the function going to be added and by who? With out this, this PR is not ready for review as it cannot be tested for functionality. Marking it as work in progress till it is ready.

Clariceco:
@devchoplife what is the status of this? When will this be completed so that we can merge it?

devchoplife:
> @devchoplife what is the status of this? When will this be completed so that we can merge it?

Waiting on https://github.com/planetdecred/dcrlibwallet/pull/206. It includes a function that gets log file path.

devchoplife:
Works fine 
@C-ollins

Sirmorrison (approved):

beansgum (approved):

PR #578, opened 2021-08-24
Fix entire license row not clickable on about page

### LINKED ISSUE #575: Entire license row on about page not clickable
On about page, the entire license row is not clickable. Only the text row is clickable. 
As seen below, the dark gray overlay is only present on the text row when clicked and other parts of the row are not clickable.
Screenshot from 2021-08-21 19-00-50

devchoplife:
This PR fixes #575 

It makes the whole license row on about page clickable:

Before:
Screenshot from 2021-08-21 19-00-50

After:
Screenshot from 2021-08-25 08-14-28

### DIFF
--- a/ui/page/about_page.go
+++ b/ui/page/about_page.go
@@ -85,37 +85,27 @@ func (pg *AboutPage) Layout(gtx layout.Context) layout.Dimensions {
 	return components.UniformPadding(gtx, body)
 }
 
-var inset = layout.Inset{
+var in = layout.Inset{
 	Top:    values.MarginPadding20,
 	Bottom: values.MarginPadding20,
 	Left:   values.MarginPadding16,
 	Right:  values.MarginPadding16,
 }
 
-func containerPadding() components.Container {
-	return components.Container{
-		Padding: inset,
-	}
-}
-
-func padding() layout.Inset {
-	return inset
-}
-
 func (pg *AboutPage) layoutRows(gtx layout.Context) layout.Dimensions {
 	w := []func(gtx C) D{
 		func(gtx C) D {
-			return containerPadding().Layout(gtx, func(gtx C) D {
+			return components.Container{Padding: in}.Layout(gtx, func(gtx C) D {
 				return components.EndToEndRow(gtx, pg.version.Layout, pg.versionValue.Layout)
 			})
 		},
 		func(gtx C) D {
-			return containerPadding().Layout(gtx, func(gtx C) D {
+			return components.Container{Padding: in}.Layout(gtx, func(gtx C) D {
 				return components.EndToEndRow(gtx, pg.buildDate.Layout, pg.buildDateValue.Layout)
 			})
 		},
 		func(gtx C) D {
-			return containerPadding().Layout(gtx, func(gtx C) D {
+			return components.Container{Padding: in}.Layout(gtx, func(gtx C) D {
 				return components.EndToEndRow(gtx, pg.network.Layout, pg.networkValue.Layout)
 			})
 		},
@@ -123,11 +113,11 @@ func (pg *AboutPage) layoutRows(gtx layout.Context) layout.Dimensions {
 			return decredmaterial.Clickable(gtx, pg.licenseRow, func(gtx C) D {
 				return layout.Flex{}.Layout(gtx,
 					layout.Rigid(func(gtx C) D {
-						return padding().Layout(gtx, pg.license.Layout)
+						return in.Layout(gtx, pg.license.Layout)
 					}),
 					layout.Flexed(1, func(gtx C) D {
 						return layout.E.Layout(gtx, func(gtx C) D {
-							return padding().Layout(gtx, func(gtx C) D {
+							return in.Layout(gtx, func(gtx C) D {
 								return pg.chevronRightIcon.Layout(gtx, values.MarginPadding20)
 							})
 						})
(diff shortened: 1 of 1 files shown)


ui/page/about_page.go
```diff
@@ -88,21 +88,62 @@ func (pg *AboutPage) Layout(gtx layout.Context) layout.Dimensions {
...
 		func(gtx C) D {
-			return components.EndToEndRow(gtx, pg.version.Layout, pg.versionValue.Layout)
+			return components.Container{
+				Padding: layout.Inset{
+					Top:    values.MarginPadding20,
+					Bottom: values.MarginPadding20,
+					Left:   values.MarginPadding16,
+					Right:  values.MarginPadding16,
+				},
+			}.Layout(gtx, func(gtx C) D {
```
Sirmorrison: This can be added into a method and called for every instance where it is used/needed.

Sirmorrison (changes requested):

ui/page/about_page.go
```diff
@@ -85,24 +85,54 @@ func (pg *AboutPage) Layout(gtx layout.Context) layout.Dimensions {
...
+}
+
+func clickablePadding() layout.Inset {
+	return layout.Inset{
+		Top:    values.MarginPadding20,
+		Bottom: values.MarginPadding20,
+		Left:   values.MarginPadding16,
+		Right:  values.MarginPadding16,
+	}
+}
```
Sirmorrison:
```suggestion
var inset = layout.Inset{
	Top:    values.MarginPadding20,
	Bottom: values.MarginPadding20,
	Left:   values.MarginPadding16,
	Right:  values.MarginPadding16,
}

func containerPadding() components.Container {
	return components.Container{
		Padding: inset,
	}
}

func padding() layout.Inset {
	return inset
}
```

Sirmorrison (changes requested):

Sirmorrison (dismissed):

ui/page/about_page.go
```diff
@@ -85,24 +85,51 @@ func (pg *AboutPage) Layout(gtx layout.Context) layout.Dimensions {
...
+
+func containerPadding() components.Container {
+	return components.Container{
+		Padding: inset,
+	}
+}
+
+func padding() layout.Inset {
+	return inset
+}
```
oshorefueled: Do you think we this? The Container type is like a syntactic sugar to differentiate the use of an inset for margin or padding. When using an Inset as a padding, use Container type and when using it as a margin, use the Inset type as it is. In this case, they do the same thing.

ui/page/about_page.go
```diff
@@ -85,24 +85,51 @@ func (pg *AboutPage) Layout(gtx layout.Context) layout.Dimensions {
...
+
 func (pg *AboutPage) layoutRows(gtx layout.Context) layout.Dimensions {
 	w := []func(gtx C) D{
 		func(gtx C) D {
-			return components.EndToEndRow(gtx, pg.version.Layout, pg.versionValue.Layout)
+			return containerPadding().Layout(gtx, func(gtx C) D {
+				return components.EndToEndRow(gtx, pg.version.Layout, pg.versionValue.Layout)
+			})
```
oshorefueled: Rather than create a method for container, you could use Container from the components package like so

```suggestion
			return components.Container{Padding: in}.Layout(gtx, func(gtx C) D {
				return components.EndToEndRow(gtx, pg.version.Layout, pg.versionValue.Layout)
			})
```

Where the variable `in` is a variable to the inset defined within the method `layoutRows()` as in
```go 
       in := layout.Inset{
		Top:    values.MarginPadding20,
		Bottom: values.MarginPadding20,
		Left:   values.MarginPadding16,
		Right:  values.MarginPadding16,
	}
```

ui/page/about_page.go
```diff
@@ -85,24 +85,51 @@ func (pg *AboutPage) Layout(gtx layout.Context) layout.Dimensions {
...
+			return containerPadding().Layout(gtx, func(gtx C) D {
+				return components.EndToEndRow(gtx, pg.network.Layout, pg.networkValue.Layout)
+			})
 		},
 		func(gtx C) D {
 			return decredmaterial.Clickable(gtx, pg.licenseRow, func(gtx C) D {
 				return layout.Flex{}.Layout(gtx,
-					layout.Rigid(pg.license.Layout),
+					layout.Rigid(func(gtx C) D {
+						return padding().Layout(gtx, pg.license.Layout)
```
oshorefueled: You could do the same thing here

oshorefueled (changes requested):

Sirmorrison (dismissed):

oshorefueled (approved):

ui/page/about_page.go
```diff
@@ -85,24 +85,41 @@ func (pg *AboutPage) Layout(gtx layout.Context) layout.Dimensions {
 	return components.UniformPadding(gtx, body)
 }
 
+var in = layout.Inset{
+	Top:    values.MarginPadding20,
+	Bottom: values.MarginPadding20,
+	Left:   values.MarginPadding16,
+	Right:  values.MarginPadding16,
+}
+
```
oshorefueled: The inset seems to only be used in the layoutRows method. You could make in a local variable in the method.

oshorefueled (changes requested):

oshorefueled (approved):

Sirmorrison (approved):

PR #584, opened 2021-08-27
Fix Clickable widget effect highlight.

Sirmorrison:
Currently when the clickableList button is pressed, the button effect extends pass the card constraint.
image

This PR is to have this resolved so it fix within the card and also to set custom highlight colors.


Sirmorrison:
PR #630 resolves this.

PR #618, opened 2021-09-07
Use high resolution icons and scale icon widgets to correct sizes

### LINKED ISSUE #556: blurry icons on higher resolution display
Some Icons tend to be blurry on screens with higher resolution. The solution was to use larger icon sizes but scale it down to the preferred size. It works fine on screens with high screen resolution but part of the icon gets clipped off on a 1080p display. This forces the use of icons that are suited for 1080p displays and maintaining a scale of 1.0

An issue has been created [upstream](https://todo.sr.ht/~eliasnaur/gio/250) to that effect.

beansgum:
Closes #556 
This pr fixes a bug that causes icons to appear blurry on higher resolution display by using icons with higher resolutions and scaling the icons widgets to the correct size.


dreacot (dismissed):

Sirmorrison (approved):
Tested on Mac and windows machines and i looks good.

dreacot (approved):

PR #625, opened 2021-09-12
Allow empty new password for wallet in serveral cases

### LINKED ISSUE #596: Allow empty password input.
Currently, on godcr, we are preventing the use of empty passwords/strings.
This should be removed to allow the use of empty strings/passwords.

The idea of this issue is to enable the user to try to execute a wallet command that requires a password, without entering a password, and getting an error.

At the moment, when there is no password entered, the button is grayed out or you cannot proceed without entering the passwords.

The idea is for them to be able click on the confirm button and get a server error that a password is needed.

JustinBeBoy:
This will resolve #596 

Currently, when we send empty password, the server resolves in this:
image

I have already work out a way to notify the user about the response.
image

Whether or not this behavior is expected and/or desired is debateable.

### DIFF
--- a/ui/decredmaterial/theme.go
+++ b/ui/decredmaterial/theme.go
@@ -362,3 +362,18 @@ func approxLuminance(c color.NRGBA) byte {
 	)
 	return byte((r*int(c.R) + g*int(c.G) + b*int(c.B)) / t)
 }
+
+func HandleEditorEvents(editors ...*widget.Editor) (bool, bool) {
+	var submit, changed bool
+	for _, editor := range editors {
+		for _, evt := range editor.Events() {
+			switch evt.(type) {
+			case widget.ChangeEvent:
+				changed = true
+			case widget.SubmitEvent:
+				submit = true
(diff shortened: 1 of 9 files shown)


Sirmorrison:
The idea of this PR is to enable the user to try to execute a wallet command that requires a password, without entering a password, and getting an error.

The server error that a password is needed is the desired behavior.

At the moment, when there is no password entered, the button is grayed out or you cannot proceed without entering the passwords.

The idea is for them to be able click on the confirm button and get a server error that a password is needed.

Sirmorrison:
On the send page/password modal, remove the password validation.

image

image

on the create start up password in the settings page, when only the startup password is entered, and the confirm button is pressed, the form submits and sets the password.
* The behavior should allow for password confirmation when setting up the password.
if only the startup password is entered and the enter button is pressed, you should display an error to confirm the main password.
image

dreacot (approved):

Sirmorrison (approved):

PR #628, opened 2021-09-14
Add makefile

beansgum:
Using a makefile would enable building a binary with build flags & variables while targeting multiple os.


Sirmorrison (commented):
It would be nice to update the readme doc so we know exactly how to build using the makefile changes.

Sirmorrison (commented):
image
 i get the error when i try to run `make` on windows.

beansgum:
@Sirmorrison  it  errors because you’re trying to build for freebsd and cross compilation isn’t possible yet.

Sirmorrison (dismissed):
Tested and works fine.

dreacot (approved):

Sirmorrison (approved):

PR #644, opened 2021-09-25
Fix top nav USD balance bugs.

### LINKED ISSUE #641: USD equivalent balance is missing from the top nav
The USD equivalent balance is missing from the top nav even when it's enabled in settings.

This was previously implemented here https://github.com/planetdecred/godcr/pull/380

I believe it might have been removed by accident, it should be re-added back if that's the case

Screenshot 2021-09-23 at 7 57 36 AM

What it should look like

Screenshot 2021-06-30 at 3 36 50 PM

amass01:
This diff fixes two issues with the USD balance in the top nav where:

 - [x] Main page's USD balance is not showing once you switch the
   setting from none to USD.

 - [x] USD conversion setting doesn't persist.

---

Closes #641.

### DIFF
--- a/ui/page/send/layout.go
+++ b/ui/page/send/layout.go
@@ -6,7 +6,6 @@ import (
 	"gioui.org/layout"
 	"gioui.org/op"
 
-	"github.com/planetdecred/dcrlibwallet"
 	"github.com/planetdecred/godcr/ui/decredmaterial"
 	"github.com/planetdecred/godcr/ui/page/components"
 	"github.com/planetdecred/godcr/ui/values"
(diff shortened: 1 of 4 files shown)


Sirmorrison (changes requested):
Turning on the USD balance from the settings page, adds the additional editor on the send page as it should be. 
image

However, when the USD balance is set to none from settings page, the send page is not updated correctly until the app is restarted.
image

The send page should revert to screenshot below, when the USD toggle is off.
image

amass01:
@Sirmorrison addressed.

dreacot (dismissed):
LGTM

ui/page/send/layout.go
```diff
@@ -223,7 +224,9 @@ func (pg *Page) pageSections(gtx layout.Context, title string, showAccountSwitch
 }
 
 func (pg *Page) toSection(gtx layout.Context) layout.Dimensions {
-
+	usdExchangeSet := components.USDExchangeValue ==
+		pg.WL.Wallet.
+			ReadStringConfigValueForKey(dcrlibwallet.CurrencyConversionConfigKey)
```
Sirmorrison: There is some form of delay in updating this setting when the page is opened. 
I believe moving this to the `handle()` should improve the performance.

Sirmorrison:
image
What the send page looks before the USD setting is set.

image
After setting the USD values

When the setting is turned off, the tx fee details are not updated.
image

the confirmation modal as well is not updated correctly after the USD values are turned off.
image

What it should look like when the USD value is not set.
image

amass01:
@Sirmorrison Addressed.

Sirmorrison (approved):

dreacot (approved):

PR #662, opened 2021-10-07
Show total time app has been open on statistics page

### LINKED ISSUE #653: Uptime bug
On the statistics page, the uptime displayed starts from when he statistics page is opened.
image

The time to be displayed is the total time the application has been opened and not when the page is opened as exiting the page and opening it again, starts the time from 0.

devchoplife:
This PR fixes #653 
This PR fixes uptime bug and allows you see the total time the app has been open on statistics page

### DIFF
--- a/ui/page/statistics_page.go
+++ b/ui/page/statistics_page.go
@@ -22,7 +22,7 @@ type StatPage struct {
 	*load.Load
 	txs         *wallet.Transactions
 	l           layout.List
-	startupTime time.Time
+	startupTime string
 	syncStatus  *wallet.SyncStatus
 	netType     string
 
@@ -38,7 +38,6 @@ func NewStatPage(l *load.Load) *StatPage {
 		},
 		netType: l.WL.Wallet.Net,
 	}
-	pg.startupTime = l.WL.Wallet.StartupTime()
 	pg.syncStatus = l.WL.SyncStatus
 	if pg.netType == "testnet3" {
 		pg.netType = "Testnet"
@@ -56,7 +55,7 @@ func (pg *StatPage) ID() string {
 }
 
 func (pg *StatPage) OnResume() {
-
+	pg.appStartTime()
 }
 
 func (pg *StatPage) layoutStats(gtx C) D {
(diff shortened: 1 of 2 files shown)


wallet/wallet.go
```diff
@@ -28,6 +28,7 @@ type Wallet struct {
 	Send               chan Response
 	Sync               chan SyncStatusUpdate
 	OverallBlockHeight int32
+	startUpTime        time.Time
```
Sirmorrison, line 31:
```suggestion
	startUpTime       string
```

ui/page/statistics_page.go
```diff
@@ -38,7 +38,7 @@ func NewStatPage(l *load.Load) *StatPage {
 		},
 		netType: l.WL.Wallet.Net,
 	}
-	pg.startupTime = time.Now()
+	pg.startupTime = l.WL.Wallet.StartupTime()
```
Sirmorrison:
```suggestion
```

Sirmorrison (changes requested):
When the statistics page button is clicked there is this delay before the page is opened.
Troubleshooting from my end, i find out that the time method is the issue.

For this to be handled correctly the start up time needs to be in the handle method and on resume method.

This method  https://github.com/planetdecred/godcr/blob/8b4e82023feefdf5cc6181ba0750f9e19cf46ee9/ui/page/statistics_page.go#L83
 should be in a separate method.

```
func (pg *StatPage) appStartTime()  {
	pg.startupTime = func(t time.Time) string {
		v := int(time.Since(t).Seconds())
		h := v / 3600
		m := (v - h*3600) / 60
		s := v - h*3600 - m*60
		return fmt.Sprintf("%02d:%02d:%02d", h, m, s)
	}(pg.WL.Wallet.StartupTime())
}
```

`startupTime` should be a string variable.

```
func (pg *StatPage) Handle()  {
	pg.appStartTime()
}
```

```
func (pg *StatPage) OnResume() {
	pg.appStartTime()
}
```

Sirmorrison (approved):

PR #663, opened 2021-10-08
Add wallet seed backup notification modal

Sirmorrison:
This PR adds wallet seed backup notification when pages loads for the first time and there is a wallet that needs its seed words to be backed up. 

image

image

### DIFF
--- a/ui/modal/info_modal_layouts.go
+++ b/ui/modal/info_modal_layouts.go
@@ -104,21 +104,20 @@ func transactionDetailsInfo(th *decredmaterial.Theme) []layout.Widget {
 }
 
 func backupInfo(th *decredmaterial.Theme) []layout.Widget {
-	textGray := "Please backup your seed words and keep them in a safe place in order to recover your funds " +
-		"if your device gets lost or broken."
-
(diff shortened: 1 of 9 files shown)


ui/modal/info_modal_layouts.go
```diff
@@ -101,3 +102,23 @@ func transactionDetailsInfo(th *decredmaterial.Theme) []layout.Widget {
 		renderers.RenderHTML(text, th).Layout,
 	}
 }
+
+func backupInfo(th *decredmaterial.Theme) []layout.Widget {
+	textGray := "Please backup your seed words and keep them in a safe place in order to recover your funds " +
```
dreacot: why does this string need to be concatenated?

dreacot (changes requested):
Currently, each time the user selects backup later and returns to the overview page, the modal pops back up, the user experience for that isn't good, a boolean should be created to check if backup later has be selected and set to true, so the dialog doesn't display for the current user session, until the app is restarted

dreacot:
This seems to be the correct mockup as indicated by @fegor_clinton

Screenshot 2021-10-11 at 9 40 17 PM
Screenshot 2021-10-11 at 9 40 38 PM

dreacot (approved):
LGTM

Issue #619, opened 2021-09-08
Integrate Dex into Godcr


song50119:
Integrate Dex into Godcr, will use modules from https://github.com/decred/dcrdex

- [x] On the UI create new menu item to navigate to dex page
- [x] Add modal to add new server
- [x] Add fee selector modal to choose DCR/BTC to pay the fee
- [ ]  Add notifications modal
- [ ] Create mini trade form to swap dcr/btc coins(immediately sell/buy market book orders)
- [ ] Add Dex settings page, orders history page, wallet settings page

<img width="912" alt="Screen Shot 2022-06-04 at 00 27 59" src="https://user-images.githubusercontent.com/9800579/171918825-1d100b01-f380-432c-91be-902fe2a3936e.png">
<img width="912" alt="Screen Shot 2022-06-04 at 00 28 04" src="https://user-images.githubusercontent.com/9800579/171918834-4bf435d5-24d4-4247-8368-8e1116480d81.png">
<img width="912" alt="Screen Shot 2022-06-04 at 00 28 12" src="https://user-images.githubusercontent.com/9800579/171918846-2d393746-953f-4f1f-b647-f99b6e599952.png">
<img width="912" alt="Screen Shot 2022-06-04 at 00 28 32" src="https://user-images.githubusercontent.com/9800579/171918856-9ea474a9-4c40-4dc4-8983-5854b974fedf.png">
<img width="912" alt="Screen Shot 2022-06-04 at 00 47 24" src="https://user-images.githubusercontent.com/9800579/171918953-d7ae1e5a-086a-4ffc-b621-66c29f0c73c3.png">
<img width="912" alt="Screen Shot 2022-06-04 at 00 48 13" src="https://user-images.githubusercontent.com/9800579/171919039-26752274-1435-432b-930f-ad2f84049710.png">

Notes
1. All the tasks have done at #722, but the code implemented need separate into smaller PRs.
2. Godcr will create BTC wallet and use it for DEX instead of DEX create and use BTC as it's own wallet, the rest of tasks will merge after this

PR #676, opened 2021-10-21
DEX create password, DCR wallet and add server

song50119:
Work towards #619

- Initialize app password
- Add wallet for dcr to use to pay the dex server fee
- Add dex server this come before adding any wallet
- Show the fee confirmations

Test server: dex-test.ssgen.io:7232


ui/page/main_page.go
```diff
@@ -160,12 +163,45 @@ func (mp *MainPage) initNavItems() {
 				Title:         values.String(values.StrMore),
 				PageID:        MorePageID,
 			},
+			{
+				Clickable:     mp.Theme.NewClickable(true),
+				Image:         mp.Icons.DexLogo,
+				ImageInactive: mp.Icons.DexLogo,
+				Title:         values.String(values.StrDex),
+				PageID:        dexclient.MarketPageID,
+			},
```
itswisdomagain: I think the Dex tab should come before the More tab, the More tab should be last.

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,212 @@
...
+	return pg
+}
+
+func (pg *Page) ID() string {
+	return MarketPageID
+}
+
+func (pg *Page) OnResume() {
+	pg.ctx, pg.ctxCancel = context.WithCancel(context.TODO())
+	pg.refreshUser()
```
itswisdomagain: Can just do `pg.initializeModal = false` directly here, although I doubt it is even necessary. It appears this method is only called once, after the page is created with `NewMarketPage`. When the page is created, `initializeModal` is set to false.

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,212 @@
...
+func (pg *Page) pageSections(gtx layout.Context, body layout.Widget) layout.Dimensions {
+	return layout.Inset{
+		Bottom: values.MarginPadding8,
+	}.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
+		return pg.Theme.Card().Layout(gtx, func(gtx C) D {
+			gtx.Constraints.Min.X = gtx.Constraints.Max.X
+			return layout.UniformInset(values.MarginPadding16).Layout(gtx, body)
+		})
+	})
+}
```
itswisdomagain: This isn't used (yet). If it is useful for some follow up work, can remove it from this PR and restore it later when needed.

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+}
+
+func (pg *Page) Layout(gtx C) D {
+	return pg.registrationStatusLayout(gtx)
+}
+
+func (pg *Page) registrationStatusLayout(gtx C) D {
+	dex := pg.Dexc.Core.User().Exchanges[pg.selectedMaket.host]
+	if dex == nil || !dex.Connected {
```
itswisdomagain: Instead of showing status for the dextest host only, this can work with any dex the user connects, in case a user connects a different dex.

For e.g., the code below can pick the first dex on the exchanges map to use, if the map is not empty.

```suggestion
	var dex *core.Exchange
	for _, dex = range pg.Dexc.Core.User().Exchanges {
		break
	}
	if dex == nil || !dex.Connected {
```

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,212 @@
...
+		Bottom: values.MarginPadding8,
+	}.Layout(gtx, func(gtx layout.Context) layout.Dimensions {
+		return pg.Theme.Card().Layout(gtx, func(gtx C) D {
+			gtx.Constraints.Min.X = gtx.Constraints.Max.X
+			return layout.UniformInset(values.MarginPadding16).Layout(gtx, body)
+		})
+	})
+}
+
+func (pg *Page) Handle() {
```
itswisdomagain: Please doc this and other exported methods and fields. May not need to do so for all files in the project having exported methods but at least for the exported methods added by the PR.

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,212 @@
...
+		md.Show()
+		return
+	}
+
+	// The dcr wallet must be connected before registering with
+	// a DEX.
+	// TODO: Since other assets can now be used to pay the fee,
+	// this shouldn't be a pre-requirement. Instead, attempt to
+	// connect a DEX first and determine what wallet is required
+	// for fee payment.
```
itswisdomagain: My recommendation is to connect the DEX first since it's possible that the DEX a user chooses to connect to does not use dcr for fee payment. After connecting the DEX, check the assets that are supported for fee payment and connect any of them. I have a diff for this that I'll share later.

ui/page/dexclient/add_dex_modal.go
```diff
@@ -0,0 +1,108 @@
...
+func (md *addDexModal) Handle() {
+	if md.addDexServer.Button.Clicked() {
+		if md.dexServerAddress.Editor.Text() == "" || md.isSending {
+			return
+		}
+
+		md.isSending = true
+		go func() {
+			c := []byte(md.cert.Editor.Text())
+			ce, err := md.Dexc.GetDEXConfig(md.dexServerAddress.Editor.Text(), c)
```
itswisdomagain: Should use `DiscoverAccount` now instead of `GetDEXConfig`. This will require modifying this modal to show a `App Password` field. Not really important now anyways, since `DiscoverAccount` only works with dex clients restored from seed, which isn't yet supported by this PR. A todo here will be nice.

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,212 @@
...
+			cfModal := newConfirmRegisterModal(md.Load)
+			cfModal.updateCertAndExchange(cert, ce)
+			cfModal.confirmed = func(password []byte) {
+				pg.refreshUser()
+				pg.connectDex(ce.Host, password)
+			}
+			cfModal.Show()
+		}
+		md.Show()
+		return
```
itswisdomagain: Instead of 2 modals, why not have the `addDexModal` also display the confirmation prompt and complete the registration?

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,212 @@
...
+		u.Assets[dexc.DefaultAssetID].Wallet != nil &&
+		!u.Assets[dexc.DefaultAssetID].Wallet.Open {
+		md := newUnlockWalletModal(pg.Load)
+		md.unlocked = func(password []byte) {
+			pg.refreshUser()
+			pg.connectDex(pg.selectedMaket.host, password)
+		}
+		md.Show()
+		return
+	}
```
itswisdomagain: This wallet unlock isn't necessary. If a wallet needs to be unlocked for pending trades, it is unlocked when the user logs in. If required for a new trade, it is unlocked when/before the trade is placed. If required for fee payment, it is unlocked when `Core.Register` is called with the app password.

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,212 @@
...
+	return MarketPageID
+}
+
+func (pg *Page) OnResume() {
+	pg.ctx, pg.ctxCancel = context.WithCancel(context.TODO())
+	pg.refreshUser()
+
+	if pg.Dexc.Core.User().Initialized && pg.Dexc.IsLoggedIn {
+		go pg.readNotifications()
+	}
```
itswisdomagain: Don't need to be logged in to receive core notifications, not sure it's a requirement that the app be initialized as well.

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// Initialized client must be logged in.
+	if !pg.Dexc.IsLoggedIn {
+		md := newloginModal(pg.Load)
+		md.loggedIn = func(password []byte) {
+			pg.refreshUser()
+			pg.Dexc.IsLoggedIn = true
+			if u.Assets[dexc.DefaultAssetID] != nil &&
+				u.Assets[dexc.DefaultAssetID].Wallet != nil {
+				pg.connectDex(pg.selectedMaket.host, password)
```
itswisdomagain: This really only calls `go pg.readNotifications`, which can be done much earlier in `OnResume()` rather than on login or other places. The `pg.connectDex` function can be removed, I believe.

ui/window.go
```diff
@@ -74,7 +79,7 @@ type WriteClipboard struct {
 // Should never be called more than once as it calls
 // app.NewWindow() which does not support being called more
 // than once.
-func CreateWindow(wal *wallet.Wallet) (*Window, *app.Window, error) {
+func CreateWindow(appCtx context.Context, wal *wallet.Wallet, dc *dexc.Dexc) (*Window, *app.Window, error) {
```
itswisdomagain:
```suggestion
func CreateWindow(appCtx context.Context, wal *wallet.Wallet, dexc *dexc.Dexc) (*Window, *app.Window, error) {
```

ui/window.go
```diff
@@ -59,6 +62,8 @@ type Window struct {
 	sysDestroyWithSync    bool
 	walletAcctMixerStatus chan *wallet.AccountMixer
 	internalLog           chan string
+
+	dexc *dexc.Dexc
```
itswisdomagain: `win.load` has a `dexc` field as well, this can go.

dexc/dexc.go
```diff
@@ -0,0 +1,198 @@
...
+
+	coreConfig *core.Config
+	IsRunning  bool
+	IsLoggedIn bool // Keep user logged in state
+}
+
+const (
+	DefaultAsset            = "dcr"
+	DefaultAssetID   uint32 = 42
+	ConversionFactor        = 1e8
```
itswisdomagain: `ConversionFactor` is unused. I don't think there is still a concept of default asset, instead of these, use `dcr.BipID` and `dex.BipIDSymbol(dcr.BipID)` where necessary.

dexc/dexc.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	n, err := dex.NetFromString(net)
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	}
+
+	// Prepare the Core.
+	coreConfig := &core.Config{
+		DBPath: dbPath, // global set in config.go
```
itswisdomagain: Not sure that this comment refers to.

dexc/dexc.go
```diff
@@ -0,0 +1,198 @@
...
+
+// Shutdown causes the dex client to shutdown. The shutdown attempt will be
+// prevented if there are active orders or if some other error occurs. The
+// bool return indicates if shutdown was successful. If successful, dexc will
+// need to be restarted before it can be used again.
+func (d *Dexc) Shutdown() bool {
+	return d.shutdown(false)
+}
+
+// Shutdown causes the dex client to shutdown regardless of whether or not there
```
itswisdomagain:
```suggestion
// ForceShutdown causes the dex client to shutdown regardless of whether or not there
```

dexc/dexc.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// ConnectedDcrWalletIDConfigKey is used as the key in a simple key-value
+	// database to store the ID of the dcr wallet that is connected to the DEX
+	// client, to facilitate reconnecting the wallet when godcr is restarted.
+	ConnectedDcrWalletIDConfigKey = "dex_dcr_wallet_id"
+)
+
+func NewDex(debugLevel string, dbPath, net string, w io.Writer) (*Dexc, error) {
+	logMaker := initLogging(debugLevel, true, w)
+	log = logMaker.Logger("DEXC")
```
itswisdomagain: Can have `main.go` pass in a logger instead of creating one here. That should also eliminate the need for a `log.go` file.

dexc/types.go
```diff
@@ -0,0 +1,38 @@
+package dexc
+
+import (
+	"decred.org/dcrdex/client/core"
+	"decred.org/dcrdex/dex/encode"
+)
+
+// Response represents a discriminated union for wallet responses.
+// Either Resp or Err must be nil.
+type Response struct {
```
itswisdomagain: Non of the types defined in this file are used. If they're needed for upcoming work, can add this file then and remove from this PR.

main.go
```diff
@@ -56,14 +58,25 @@ func main() {
...
+		cancel()
 		wal.Shutdown()
 		os.Exit(0)
 	}()
 
-	win, appWindow, err := ui.CreateWindow(wal)
+	dbPath := filepath.Join(cfg.HomeDir, cfg.Network, "dexc.db")
+	dc, err := dexc.NewDex(cfg.DebugLevel, dbPath, cfg.Network, logWriter{})
+	if err != nil {
+		fmt.Printf("error creating Dex: %s", err)
```
itswisdomagain: Seems you can use `log.Errorf` here.

ui/load/load.go
```diff
@@ -37,7 +39,7 @@ type Receiver struct {
 type Icons struct {
 	ContentAdd, NavigationCheck, NavigationMore, ActionCheckCircle, ActionInfo, NavigationArrowBack,
 	NavigationArrowForward, ActionCheck, ChevronRight, NavigationCancel, NavMoreIcon,
-	ImageBrightness1, ContentClear, DropDownIcon, Cached, ContentRemove *widget.Icon
+	ImageBrightness1, ContentClear, DropDownIcon, Cached, ContentRemove, ActionSwapHoriz *widget.Icon
```
itswisdomagain: `ActionSwapHoriz` is assigned but not used.

ui/page/dexclient/confirm_register_modal.go
```diff
@@ -0,0 +1,116 @@
...
+		},
+		func(gtx C) D {
+			return md.register.Layout(gtx)
+		},
+	}
+
+	return md.modal.Layout(gtx, w, 900)
+}
+
+func (md *confirmRegisterModal) updateCertAndExchange(cert []byte, ce *core.Exchange) {
```
itswisdomagain: Can pass these to the `newConfirmRegisterModal` constructor rather than via an additional method.

ui/page/dexclient/create_wallet_modal.go
```diff
@@ -0,0 +1,253 @@
...
+	sourceAccountSelector *components.AccountSelector
+	modal                 *decredmaterial.Modal
+	createNewWallet       decredmaterial.Button
+	walletPassword        decredmaterial.Editor
+	appPassword           decredmaterial.Editor
+	walletName            decredmaterial.Editor
+	RPCAddress            decredmaterial.Editor
+	RPCPort               decredmaterial.Editor
+	RPCUsername           decredmaterial.Editor
+	RPCPassword           decredmaterial.Editor
```
itswisdomagain: Not used and can be removed.

ui/page/dexclient/create_wallet_modal.go
```diff
@@ -0,0 +1,253 @@
...
+		RPCPassword:      l.Theme.EditorPassword(new(widget.Editor), "RPC Password"),
+	}
+
+	md.createNewWallet.TextSize = values.TextSize12
+	md.createNewWallet.Background = l.Theme.Color.Primary
+	md.appPassword.Editor.SingleLine = true
+	md.appPassword.Editor.SetText("")
+
+	md.sourceAccountSelector = components.NewAccountSelector(md.Load).
+		Title("Sellect DCR account to pay fee").
```
itswisdomagain: The selected account isn't only used for fee payment, it may not even be used at all for fee payment.
```suggestion
		Title("Select DCR account to use with DEX").
```

ui/page/dexclient/create_wallet_modal.go
```diff
@@ -0,0 +1,253 @@
...
+			}
+
+			settings := make(map[string]string)
+			var walletType string
+			appPass := []byte(md.appPassword.Editor.Text())
+			walletPass := []byte(md.walletPassword.Editor.Text())
+
+			switch coinID {
+			case dcr.BipID:
+				settings["account"] = md.sourceAccountSelector.SelectedAccount().Name
```
itswisdomagain: Looks like we can do away with that if statement above and set the dcrwallet here:

```suggestion
			case dcr.BipID:
				selectedAccount := md.sourceAccountSelector.SelectedAccount()
				selectedDcrWallet = md.Load.WL.MultiWallet.WalletWithID(selectedAccount.WalletID)
				md.Dexc.SetWalletForDcrAsset(selectedDcrWallet)
				settings["account"] = selectedAccount.Name
```

ui/page/dexclient/create_wallet_modal.go
```diff
@@ -0,0 +1,253 @@
...
+							}),
+							layout.Rigid(func(gtx C) D {
+								return layout.Inset{Top: values.MarginPadding15}.Layout(gtx, func(gtx C) D {
+									return md.appPassword.Layout(gtx)
+								})
+							}),
+						)
+					}
+
+					return layout.Flex{Axis: layout.Vertical}.Layout(gtx,
```
itswisdomagain: Can delete?

ui/page/dexclient/create_wallet_modal.go
```diff
@@ -0,0 +1,253 @@
...
+		func(gtx C) D {
+			return layout.Flex{Axis: layout.Vertical}.Layout(gtx,
+				layout.Rigid(func(gtx C) D {
+					if md.walletInfoWidget.coinID == dexc.DefaultAssetID {
+						return md.Load.Theme.Label(values.TextSize14, "Your Decred wallet is required to pay registration fees.").Layout(gtx)
+					}
+					return layout.Dimensions{}
+				}),
+				layout.Rigid(func(gtx C) D {
+					// TODO: will create select widget to select DCR wallet
```
itswisdomagain: This is done now.

ui/page/dexclient/create_wallet_modal.go
```diff
@@ -0,0 +1,253 @@
...
+				layout.Rigid(func(gtx C) D {
+					return md.Load.Theme.Label(values.TextSize20, fmt.Sprintf("%s Wallet", md.walletInfoWidget.coinName)).Layout(gtx)
+				}),
+			)
+		},
+		func(gtx C) D {
+			return layout.Flex{Axis: layout.Vertical}.Layout(gtx,
+				layout.Rigid(func(gtx C) D {
+					if md.walletInfoWidget.coinID == dexc.DefaultAssetID {
+						return md.Load.Theme.Label(values.TextSize14, "Your Decred wallet is required to pay registration fees.").Layout(gtx)
```
itswisdomagain: This message is not necessary since registration fees can be paid in any asset.

ui/page/dexclient/listeners.go
```diff
@@ -0,0 +1,36 @@
...
+	"fmt"
+
+	"decred.org/dcrdex/client/core"
+)
+
+func (pg *Page) connectDex(h string, password []byte) {
+	go pg.readNotifications()
+	// TODO: connect to dex server to listen messages
+	fmt.Println(h, password[0])
+}
```
itswisdomagain: This really only does `go pg.readNotifications()`, doesn't need to be a method.

ui/page/dexclient/listeners.go
```diff
@@ -0,0 +1,36 @@
...
+
+// readNotifications reads from the Core notification channel.
+func (pg *Page) readNotifications() {
+	ch := pg.Dexc.NotificationFeed()
+	for {
+		select {
+		case n := <-ch:
+			fmt.Println("Recv notification", n)
```
itswisdomagain: Would this printing alone suffice to avoid littering?

ui/page/dexclient/login_modal.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/planetdecred/godcr/ui/load"
+	"github.com/planetdecred/godcr/ui/values"
+)
+
+const dexLoginModalID = "dex_login_modal"
+
+type loginModal struct {
+	*load.Load
+	modal    *decredmaterial.Modal
+	loggedIn func([]byte)
```
itswisdomagain: Should not pass passwords around unless absolutely necessary. This isn't needed or used currently.

ui/page/dexclient/password_modal.go
```diff
@@ -0,0 +1,115 @@
...
+
+const dexPasswordModalID = "dex_password_modal"
+
+type passwordModal struct {
+	*load.Load
+	modal *decredmaterial.Modal
+
+	createPassword                decredmaterial.Button
+	appPassword, appPasswordAgain decredmaterial.Editor
+	marketBaseID, marketQuoteID   uint32
```
itswisdomagain: Unused.

ui/page/dexclient/unlock_wallet_modal.go
```diff
@@ -0,0 +1,155 @@
...
+	*load.Load
+	modal            *decredmaterial.Modal
+	unlockWallet     decredmaterial.Button
+	appPassword      decredmaterial.Editor
+	isSending        bool
+	walletInfoWidget *walletInfoWidget
+	unlocked         func([]byte)
+}
+
+func newUnlockWalletModal(l *load.Load) *unlockWalletModal {
```
itswisdomagain: This is not necessary (yet). Can remove this file from this PR but keep it around somewhere in case it's needed again. I believe the DEX client currently performs auto-wallet unlock and does not need a special form.

itswisdomagain (changes requested):
Looks good and works well. Major change required is connecting a DEX before connecting a wallet to pay fees. Also, if there are no plans to add assets aside dcr and btc for now, no need to initialize them or add their icons.

While reviewing, I tinkered with some of the requested changes and made a commit here for your perusal: https://github.com/itswisdomagain/godcr-1/commit/5bcef82eff640a09ef8ebc2af1a466adf960dbb1

ui/page/dexclient/confirm_register_modal.go
```diff
@@ -0,0 +1,116 @@
...
+func (md *confirmRegisterModal) Layout(gtx layout.Context) D {
+	w := []layout.Widget{
+		func(gtx C) D {
+			return md.Load.Theme.Label(values.TextSize20, "Confirm Registration").Layout(gtx)
+		},
+		func(gtx C) D {
+			return md.Load.Theme.Label(values.TextSize14, "Enter your app password to confirm DEX registration. When you submit this form, 1.000 DCR will be spent from your Decred wallet to pay registration fees.").Layout(gtx)
+		},
+		func(gtx C) D {
+			return md.Load.Theme.Label(values.TextSize14, "The DCR lot size for this DEX is 1.000 DCR. All trades are in multiples of this lot size. This is the minimum possible trade amount in DCR.").Layout(gtx)
```
itswisdomagain: Replace these default values with values from the dex config. Can modify the `newConfirmRegisterModal` to also accept the feeAmt and lotSize as variables for use here.

ui/page/dexclient/markets.go
song50119: Any page implement `load.Page` interface

itswisdomagain: Yeah, I noticed that, but the `load.Page` interface methods aren't commented either; and even after they're commented, exported fields and methods in go files should also be commented/doc'ed. Since this is a wide-reaching change, can set it aside for some future work, perhaps make an issue for it in the meantime.

dexc/dexc.go
```diff
@@ -0,0 +1,190 @@
...
+
+	"decred.org/dcrdex/client/asset"
+	"github.com/planetdecred/dcrlibwallet"
+	"github.com/planetdecred/dcrlibwallet/dexdcr"
+
+	"decred.org/dcrdex/client/core"
+	"decred.org/dcrdex/dex"
+)
+
+// Dexc represents of the Decred exchange client.
```
itswisdomagain: This what you mean maybe?
```suggestion
// Dexc represents the Decred Dex client.
```

dexc/dexc.go
```diff
@@ -0,0 +1,190 @@
...
+
+// Shutdown causes the dex client to shutdown. The shutdown attempt will be
+// prevented if there are active orders or if some other error occurs. The
+// bool return indicates if shutdown was successful. If successful, dexc will
+// need to be restarted before it can be used again.
+func (d *Dexc) Shutdown() bool {
+	return d.shutdown(false)
+}
+
+// Shutdown causes the dex client to shutdown regardless of whether or not there
```
itswisdomagain:
```suggestion
// ForceShutdown causes the dex client to shutdown regardless of whether or not there
```

dexc/log.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"github.com/decred/slog"
+)
+
+// log is a logger that is initialized with no output filters.  This
+// means the package will not perform any logging by default until the caller
+// requests it.
+var apiLog = slog.Disabled
+var log = slog.Disabled
```
itswisdomagain: `apiLog` isn't used. Since `NewDex` receives a `logMaker` argument, there may not be a need to store a global `log` variable, but I see global variables used elsewhere in the godcr codebase, so no issues there I guess.

log.go
```diff
@@ -114,3 +120,16 @@ func setLogLevels(logLevel string) {
 		setLogLevel(subsystemID, logLevel)
 	}
 }
+
+// initLogging initializes the logging rotater to write logs to logFile and
+// create roll files in the same directory. initLogging must be called before
+// the package-global log rotator variables are used.
+func initLogging(lvl string, utc bool, w io.Writer) *dex.LoggerMaker {
```
itswisdomagain:
```suggestion
func initDexLogging(lvl string, utc bool, w io.Writer) *dex.LoggerMaker {
```

ui/page/main_page.go
```diff
@@ -28,6 +30,7 @@ const (
 	TicketsNavID
 	ProposalsNavID
 	MoreNavID
+	DexNavID
```
itswisdomagain: This and the other page IDs aren't used anymore.

ui/page/dexclient/login_modal.go
```diff
@@ -0,0 +1,89 @@
...
+const dexLoginModalID = "dex_login_modal"
+
+type loginModal struct {
+	*load.Load
+	modal    *decredmaterial.Modal
+	loggedIn func()
+
+	submit                      decredmaterial.Button
+	appPassword                 decredmaterial.Editor
+	marketBaseID, marketQuoteID uint32
```
itswisdomagain: Not used.

ui/page/dexclient/add_dex_modal.go
```diff
@@ -0,0 +1,178 @@
...
+	addDexServer     decredmaterial.Button
+	dexServerAddress decredmaterial.Editor
+	isSending        bool
+	cert             decredmaterial.Editor
+	done             func()
+}
+
+type walletInfoWidget struct {
+	image    *decredmaterial.Image
+	coin     string
```
itswisdomagain: `coin` is set but unused. I also don't see how `coin` differs from `coinName`.

ui/page/dexclient/create_wallet_modal.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (md *createWalletModal) OnDismiss() {
+}
+
+func (md *createWalletModal) OnResume() {
+}
+
+func (md *createWalletModal) Handle() {
+	if md.createNewWallet.Button.Clicked() {
+		if strings.Trim(md.appPassword.Editor.Text(), " ") == "" || md.isSending {
```
itswisdomagain: Shouldn't trim passwords, `' '` might be a valid password.

itswisdomagain (changes requested):
Almost there. Just a few nits from me now.

ui/page/dexclient/confirm_register_modal.go
```diff
@@ -0,0 +1,120 @@
...
+		},
+		func(gtx C) D {
+			amount := strconv.FormatFloat(float64(md.dex.Fee.Amt)/1e8, 'f', -1, 64)
+			txt := fmt.Sprintf("Enter your app password to confirm DEX registration. When you submit this form, %s DCR will be spent from your Decred wallet to pay registration fees.", amount)
+			return md.Load.Theme.Label(values.TextSize14, txt).Layout(gtx)
+		},
+		func(gtx C) D {
+			// lotSize := md.dex.Markets[md.dex.Host].LotSize
+			txt := fmt.Sprintf("The DCR lot size for this DEX is %d DCR. All trades are in multiples of this lot size. This is the minimum possible trade amount in DCR.", 1)
+			return md.Load.Theme.Label(values.TextSize14, txt).Layout(gtx)
```
itswisdomagain: I have an idea at https://github.com/itswisdomagain/godcr-1/commit/a414c7a44187f9a11c294e0cc11fb41eb44270e4 for showing the correct lot size.
Display output sample:
Screenshot 2021-10-29 at 09 17 56

itswisdomagain (commented):
Looks good to go from my end. Just one more possible improvement.

itswisdomagain:
I have this issue with fee payments sometimes not getting broadcasted until godcr restart. Occasionally, I see the following error logs after paying the dex fee but the fee tx may still not be broadcasted even if there is no error log:
```
2021-10-29 09:33:43.025 [INF] DLWL: Inserting unconfirmed transaction b90cdf32172c7e699be17ac9728d77671dbbd33cdaf58f515c315063f772e751
2021-10-29 09:33:43.105 [INF] DLWL: [1] New Transaction b90cdf32172c7e699be17ac9728d77671dbbd33cdaf58f515c315063f772e751
2021-10-29 08:33:43.267 [INF] CORE: notify: |SUCCESS| (feepayment) Fee payment in progress - Waiting for 1 confirmations before trading at dex-test.ssgen.io:7232
2021-10-29 09:33:43.827 [WRN] DLWL: 47.242.243.123:19108 reject(tx, REJECT_INSUFFICIENTFEE, b90cdf32172c7e699be17ac9728d77671dbbd33cdaf58f515c315063f772e751): transaction b90cdf32172c7e699be17ac9728d77671dbbd33cdaf58f515c315063f772e751 has insufficient priority (6.32261967816092e+06 <= 5.76e+07)
2021-10-29 09:33:43.925 [WRN] DLWL: 116.62.229.19:19108 reject(tx, REJECT_INSUFFICIENTFEE, b90cdf32172c7e699be17ac9728d77671dbbd33cdaf58f515c315063f772e751): transaction b90cdf32172c7e699be17ac9728d77671dbbd33cdaf58f515c315063f772e751 has insufficient priority (6.32261967816092e+06 <= 5.76e+07)
```

In all cases though, I see
```
2021-10-29 09:33:43.025 [INF] DLWL: Inserting unconfirmed transaction b90cdf32172c7e699be17ac9728d77671dbbd33cdaf58f515c315063f772e751
2021-10-29 09:33:43.105 [INF] DLWL: [1] New Transaction b90cdf32172c7e699be17ac9728d77671dbbd33cdaf58f515c315063f772e751
```
but the tx may not show up on dcrdata until godcr is restarted.

My guess is the dex is using a too little miners fee for the dex fee tx which causes it to be rejected by spv peers until godcr restart where dcrlibwallet/dcrwallet resends pending transactions. Confirming this theory will require constructing and sending raw transactions with small fees using dcrlibwallet, to see if the sent transaction gets indexed by dcrlibwallet but isn't found on dcrdata.

UPDATE:
A couple godcr restarts later, fee tx isn't on dcrdata still. Reset and re-connected the dex a couple times, same issue, restarting godcr that had fixed this for me isn't working now. I think this is the issue you had one time @song50119, where your dex fee remained at 0 confs for quite some time.

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,137 @@
...
+func (pg *Page) ID() string {
+	return MarketPageID
+}
+
+func (pg *Page) OnResume() {
+	pg.ctx, pg.ctxCancel = context.WithCancel(context.TODO())
+
+	if pg.Dexc.Core.User().Initialized && pg.Dexc.IsLoggedIn {
+		go pg.readNotifications()
+	}
```
itswisdomagain: The if statement surrounding `go pg.readNotifications()` is unnecessary and prevents catching fee payment notifications after the user registers. The result is that the page continues to display `Waiting for confirmations` even after the fee is confirmed because the notification listener is not enabled. It's safe to just call `go pg.readNotifications()` directly, without an if surrounding it.

itswisdomagain (changes requested):
Just one more change requirement from me. I had tried to test what happens after the fee is confirmed but was unable to because multiple fee payment attempts had their transactions rejected and thus never confirmed. I've tested it now and it doesn't work quite well.

One additional thing, once any modal is displayed on the dex page, the rest of the app cannot be used until the user completes what's required by the modal. For a new dex user, that would require initializing the app, adding a DEX, adding a wallet and paying the fee before the modals disappear and the user can use the rest of the godcr app. Consider adding a close [x] button to each modal or allowing the modals to be closed by clicking outside them.

ui/page/dexclient/markets.go
song50119: right!

itswisdomagain (dismissed):

dexc/dexc.go
```diff
@@ -0,0 +1,190 @@
...
+		log.Error(err)
+		return nil, err
+	}
+
+	coreConfig := &core.Config{
+		DBPath: dbPath,
+		Net:    n,
+		Logger: logMaker.Logger("CORE"),
+		// TorProxy:     TorProxy,
+		// TorIsolation: TorIsolation,
```
Sirmorrison: Are these commented codes still needed? I no, kindly remove.

ui/page/dexclient/add_dex_modal.go
```diff
@@ -0,0 +1,176 @@
...
+		modal:            l.Theme.ModalFloatTitle(),
+		dexServerAddress: l.Theme.Editor(new(widget.Editor), "DEX Address"),
+		addDexServer:     l.Theme.Button("Submit"),
+		cert:             l.Theme.Editor(new(widget.Editor), "Cert content"),
+	}
+
+	md.addDexServer.TextSize = values.TextSize12
+	md.addDexServer.Background = l.Theme.Color.Primary
+
+	md.dexServerAddress.Editor.SetText(testDexHost) // TODO: Debug feature only, remove before merging.
```
Sirmorrison: From the comment, i believe this should be removed.

main.go
```diff
@@ -56,16 +58,28 @@ func main() {
...
 		<-shutdown
+		cancel()
 		wal.Shutdown()
 		os.Exit(0)
 	}()
 
-	win, appWindow, err := ui.CreateWindow(wal)
+	dbPath := filepath.Join(cfg.HomeDir, cfg.Network, "dexc.db")
+	logMaker := initDexLogging(cfg.DebugLevel, true, logWriter{})
+	dexc, err := dexc.NewDex(dbPath, cfg.Network, logMaker)
```
Sirmorrison: Kindly confirm why we are initializing an app dex db. Why isn't this handled from dcrlibwallet?.
Same with the logger.

ui/page/dexclient/confirm_register_modal.go
```diff
@@ -0,0 +1,139 @@
...
+			if err != nil {
+				md.Toast.NotifyError(err.Error())
+				return
+			}
+
+			md.confirmed()
+			md.Dismiss()
+		}()
+	}
+
```
Sirmorrison: remove white space.

ui/page/dexclient/password_modal.go
```diff
@@ -0,0 +1,114 @@
...
+	md.createPassword.Background = l.Theme.Color.Primary
+	md.appPassword.Editor.SingleLine = true
+	md.appPasswordAgain.Editor.SingleLine = true
+
+	return md
+}
+
+func (md *passwordModal) ModalID() string {
+	return dexPasswordModalID
+}
```
Sirmorrison: Why do we have a new password modal?.. We currently have password modals in the modal package.

Sirmorrison (changes requested):
I believe we need more detailed description on this PR... Reason is i see a few unrelated changes to the issue addressed here.

Like `appCTX` dex initializing a db on godcr etc.

Also, details on how this PR can be properly tested.

main.go
itswisdomagain: Do you mean like a `dcrlibwallet.InitDexClient` function/method? `dexc.NewDex` ultimately calls `decred.org/dcrdex/client/core.New()`, which requires a dbPath and logger though, so regardless of where `core.New` is called from, a separate db will be created by `core` and a dbPath will need to be provided.

Sirmorrison: So basically, when adding dex to mobile, another db will have to be initialized. I believe we can make this work from dcrlibwallet so we dont have to repeat this on all the platforms we will be integrating dex on. That is why we have dcrlibwallet.

We did same for VSPD. although i understand the functionalities are totally different but it will be nice to have all dex initialization from dcrlibwallet, and all we do on the platform is call the need methods.

itswisdomagain:
> i see a few unrelated changes to the issue addressed here.
> Like appCTX dex initializing a db on godcr etc.

Both those changes re `appCTX` and additional db are part of the DEX client initialization process. I think that initial setup is missing in the PR description, although it might also be too granular to include.

This PR uses a couple dcrdex packages to provide DEX client functionality. The primary package for this purpose is `decred.org/dcrdex/client/core` which provides methods for initializing the client, adding wallets, placing and canceling orders and executing trades. Once `decred.org/dcrdex/client/core` is started, a number of background processes may run simultaneously. When starting `core`, a `ctx` is required that will be used to signal to core when all processes should shutdown. This context cannot be a random `context.Background()` but a context that actually gets canceled when the GoDCR app is shutting down; this signals `core` to also shutdown its activities. That's why the `appCtx` was introduced. In all honesty though, there should have been an `appCtx` running around somewhere to ensure pages are closed, background processes stopped when the godcr app is shut down; instead of using `context.Background()` as I've seen in a couple portions of this codebase.

As explained in a comment above, `core` needs to maintain its own database to function effectively. The DEX client is a pretty big app on its own and most of the work being done on the GoDCR end will be providing the interface to perform DEX functionality using `core` behind the scene.

main.go
itswisdomagain: I understand now. That's possible. Rule of thumb will be moving all code bits that don't concern the UI to dcrlibwallet. I'll discuss this with @song50119.

Sirmorrison:
> > i see a few unrelated changes to the issue addressed here.
> > Like appCTX dex initializing a db on godcr etc.
> 
> Both those changes re `appCTX` and additional db are part of the DEX client initialization process. I think that initial setup is missing in the PR description, although it might also be too granular to include.
> 
> This PR uses a couple dcrdex packages to provide DEX client functionality. The primary package for this purpose is `decred.org/dcrdex/client/core` which provides methods for initializing the client, adding wallets, placing and canceling orders and executing trades. Once `decred.org/dcrdex/client/core` is started, a number of background processes may run simultaneously. When starting `core`, a `ctx` is required that will be used to signal to core when all processes should shutdown. This context cannot be a random `context.Background()` but a context that actually gets canceled when the GoDCR app is shutting down; this signals `core` to also shutdown its activities. That's why the `appCtx` was introduced. In all honesty though, there should have been an `appCtx` running around somewhere to ensure pages are closed, background processes stopped when the godcr app is shut down; instead of using `context.Background()` as I've seen in a couple portions of this codebase.
> 
> As explained in a comment above, `core` needs to maintain its own database to function effectively. The DEX client is a pretty big app on its own and most of the work being done on the GoDCR end will be providing the interface to perform DEX functionality using `core` behind the scene.

Valid point.. however, i believe dcrlibwallet is designed for this. We can pass any required parameter to the calling function from the lib.

dcrwallet, VSPD etc operate almost like dex.

ui/page/dexclient/password_modal.go
song50119: This modal have two inputs for password and re-password

Sirmorrison: We dont need this modal here we already have the `NewCreatePasswordModal`. You can see example use case on settings page... Start up password section.
image

Sirmorrison: I believe it was what was used here... Issue with modal is that i cannot cancel the action as there is no cancel button.
image

Also, starting a page with a modal is weird as i will like to understand what happens when the cancel button is clicked.

ui/load/load.go
```diff
@@ -130,6 +135,9 @@ func (l *Load) RefreshTheme() {
 		l.Theme.SwitchDarkMode(isDarkModeOn)
 	}
 }
+func (l *Load) Dexc() *dcrlibwallet.DexClient {
+	return l.WL.MultiWallet.DexClient()
+}
```
Sirmorrison, line 142: I believe this is not needed as it is used only once on the debug page. Further comments on where it is used on the debug page Line 149.

ui/page/debug_page.go
```diff
@@ -128,3 +138,19 @@ func (pg *DebugPage) Layout(gtx C) D {
...
 }
+
+func (pg *DebugPage) resetDexData() {
+	// Show confirm modal before resetting dex client data.
+	confirmModal := modal.NewInfoModal(pg.Load).
+		Title("Confirm DEX Client Reset").
+		Body("You may need to restart godcr before you can use the DEX again. Proceed?").
+		NegativeButton(values.String(values.StrCancel), func() {}).
+		PositiveButton("Reset DEX Client", func() {
+			if pg.Dexc().Reset() {
```
Sirmorrison, line 150:
```suggestion
			if pg.WL.MultiWallet.DexClient.Reset() {
```

This should follow same approach used here https://github.com/planetdecred/godcr/blob/989a13fead1759eae8a28f51f4c0b8e6b7eaf76c/ui/page/proposal/proposals_page.go#L121

ui/page/dexclient/add_dex_modal.go
```diff
@@ -0,0 +1,172 @@
...
+		"ltc": icons.LTC,
+	}
+	coin, ok := m[strings.ToLower(coinName)]
+
+	if !ok {
+		return nil
+	}
+
+	return coin
+}
```
Sirmorrison: This can be moved to the `components` package

ui/page/dexclient/confirm_register_modal.go
```diff
@@ -0,0 +1,136 @@
...
+	md.DismissModal(md)
+}
+
+func (md *confirmRegisterModal) OnDismiss() {
+	md.appPassword.Editor.SetText("")
+}
+
+func (md *confirmRegisterModal) OnResume() {
+	md.appPassword.Editor.Focus()
+}
```
Sirmorrison: I believe we can use the `NewPasswordModal()` for this. 
A functions can be added to collect the ` dex *core.Exchange` and  `cert []byte` parameters.

Kindly have this modal removed.

ui/page/dexclient/confirm_register_modal.go
```diff
@@ -0,0 +1,136 @@
...
+
+func formatAmount(assetID uint32, assetName string, amount uint64) string {
+	assetInfo, err := asset.Info(assetID)
+	if err != nil {
+		return fmt.Sprintf("%d [%s units]", amount, assetName)
+	}
+	unitInfo := assetInfo.UnitInfo
+	convertedLotSize := float64(amount) / float64(unitInfo.Conventional.ConversionFactor)
+	return fmt.Sprintf("%s %s", strconv.FormatFloat(convertedLotSize, 'f', -1, 64), unitInfo.Conventional.Unit)
+}
```
Sirmorrison: This can be moved to the components package.

ui/page/dexclient/confirm_register_modal.go
```diff
@@ -0,0 +1,136 @@
...
+		},
+		func(gtx C) D {
+			markets := make([]string, 0, len(md.dex.Markets))
+			for _, mkt := range md.dex.Markets {
+				lotSize := formatAmount(mkt.BaseID, mkt.BaseSymbol, mkt.LotSize)
+				markets = append(markets, fmt.Sprintf("Base: %s\tQuote: %s\tLot Size: %s", strings.ToUpper(mkt.BaseSymbol), strings.ToUpper(mkt.QuoteSymbol), lotSize))
+			}
+			txt := fmt.Sprintf("This DEX supports the following markets. All trades are in multiples of each market's lot size.\n\n%s", strings.Join(markets, "\n"))
+			return md.Load.Theme.Label(values.TextSize14, txt).Layout(gtx)
+		},
```
Sirmorrison: You can use a condition to have this displayed on the `NewPasswordModal()` modal.

ui/page/dexclient/listeners.go
```diff
@@ -0,0 +1,20 @@
+package dexclient
+
+import (
+	"decred.org/dcrdex/client/core"
+)
```
Sirmorrison:
```suggestion
import "decred.org/dcrdex/client/core"
```

ui/page/dexclient/login_modal.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (md *loginModal) Dismiss() {
+	md.DismissModal(md)
+}
+
+func (md *loginModal) OnDismiss() {
+	md.appPassword.Editor.SetText("")
+}
+
```
Sirmorrison: The password modal can be used in this scenario `NewPasswordModal()`.

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,130 @@
...
+func (pg *Page) registrationStatusLayout(gtx C) D {
+	dex := pg.dex()
+	if dex == nil || dex.PendingFee == nil {
+		return D{}
+	}
+	reqConfirms, currentConfs := dex.Fee.Confs, dex.PendingFee.Confs
+	return layout.Flex{Axis: layout.Vertical, Alignment: layout.Middle}.Layout(gtx,
+		layout.Rigid(func(gtx C) D {
+			return pg.Theme.Label(values.TextSize14, "Waiting for confirmations...").Layout(gtx)
+		}),
```
Sirmorrison:
```suggestion
		layout.Rigid(pg.Theme.Label(values.TextSize14, "Waiting for confirmations...").Layout),
```

Sirmorrison (changes requested):
The dex button is always active... You need to use the inactive icon until the dex tab is selected.
image

ui/page/dexclient/confirm_register_modal.go
itswisdomagain: It may not need those additional parameters, perhaps just a text to display on the modal and a callback function that is invoked when the password is provided; the callback function then completes the registration.

ui/page/dexclient/confirm_register_modal.go
itswisdomagain: I don't think this is used widely enough to be a public member of a different package, can still be an unexported function in the file where it is used. But then, I'm not familiar with the design and use of the components package.

ui/page/dexclient/confirm_register_modal.go
Sirmorrison: This is already sorted in the `NewPasswordModal()` modal so we don't need this modal. It should be removed.

ui/page/dexclient/password_modal.go
song50119: Should we change it to a page?

ui/page/debug_page.go
itswisdomagain: `Dexc().Reset()` does not return a fixed, stored value. It tries to delete the dexc database and do other stuff, then returns true if it succeeds. That's why the method name is a verb `Reset` rather than a noun `IsReset`. The return value is also only used here to determine if to show a success message or not, nothing more, don't see a reason to store the returned value in a struct property.

ui/page/components/dex.go
```diff
@@ -0,0 +1,25 @@
...
+		"ltc": icons.LTC,
+	}
+	coin, ok := m[strings.ToLower(coinName)]
+
+	if !ok {
+		return nil
+	}
+
+	return coin
+}
```
Sirmorrison: Move this to component.go.

Sirmorrison (changes requested):
image
* When the Dex page is selected, the button is still the inactive button. not Active. (I have asked the UI/UX guys to get us the logo).
* Also, When the dex page is opened, is shows password form. My recommendation will be to have a dex welcome describing what dex is about, and then a button for customer proceed. When the button is pressed, the create password modal is shown and the set up process follow. if the modal cancel button is pressed, we remain on the dex welcome page.

This just a proof of concept, however, if this is merged, it will be part of the next prerelease and as such, should look presentable on the app.

Sirmorrison:
Link to the DEX icons https://www.figma.com/file/e0cYaAzkXyXXPY00v23cQY/GoDCR?node-id=452%3A90904

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,186 @@
...
+		return dex
+	}
+	return nil
+}
+
+func (pg *Page) registrationStatusLayout(gtx C) D {
+	dex := pg.dex()
+	if dex == nil || dex.PendingFee == nil {
+		return pg.Theme.Label(values.TextSize14, "Ready to trade").Layout(gtx)
+	}
```
itswisdomagain: If `dex` is nil, how's it ready to trade? If `pg.dex()` returns nil, it should mean a dex isn't yet connected. The add dex server page/modal should be displayed instead.

I do see that this method is only called after checking that at least 1 dex is connected, so you can just ignore the `dex == nil` part of that if statement and assume that there is a dex. Otherwise, if it is necessary to check `dex == nil`, then that condition should show the add dex modal/page.

itswisdomagain: While testing, I noticed that it is possible for `dex.PendingFee` to be nil even though the fee payment isn't known to be fully confirmed yet. This happens when the dex client is unable to reconnect the wallet that was used to pay the fee and so is unable to fetch the current fee confirms.

To reproduce, restart godcr after adding a server and paying the fee but before the fee is confirmed. After restarting, disconnect the wallet syncer from the home page BEFORE clicking the dex tab. You should see an `unable to connect wallet` error. Click the dex tab now and you'll see from the logs that the fee isn't known to be confirmed yet but the GUI shows "Ready to trade". Of course, if the trading GUI was ready, the user would be unable to trade, which leads me to a concern that we've missed before now:

The dex client reports errors via the notification channel. We need a way to display ALL errors to the user, actually almost all messages, except maybe the POKE messages. Regular users will likely not be looking at logs when running a GUI app, and this issue of not being able to connect a wallet would be easier to miss without a way of showing the user errors reported by the dex client.

itswisdomagain: Aside showing errors, I think it's a good idea to NOT show the "Ready to trade" page or any trading GUI unless trading is really ready. This means the necessary wallets should be connected and at least a DEX server should be connected with the fee paid and confirmed. This can be a TODO or a separate issue to be addressed later though.

itswisdomagain: I should also note that simply fixing the sync issue on the home page does not cause the dex client to reconnect the wallet.

ui/load/load.go
```diff
@@ -58,11 +59,14 @@ type Icons struct {
...
 	TicketRevokedIcon,
 	TicketUnminedIcon *decredmaterial.Image
+
+	DexIcon, DexIconInactive, BTC, DCR, LTC, BCH *decredmaterial.Image
 }
 
 type Load struct {
 	Theme *decredmaterial.Theme
 
+	AppCtx   context.Context
```
itswisdomagain: This still used? Believe it can go now.

ui/page/components/components.go
```diff
@@ -664,3 +664,20 @@ func TranslateErr(err error) string {
...
+		"bch": icons.BCH,
+		"ltc": icons.LTC,
+	}
+	coin, ok := m[strings.ToLower(coinName)]
+
+	if !ok {
+		return nil
+	}
+
+	return coin
```
itswisdomagain: Wonder if a switch might be better here rather than a map of several icons where only one is needed, or maybe even none if an unexpected coin name is passed:

```suggestion
	switch strings.ToLower(coinName) {
	case "btc":
		return icons.BTC
	case "dcr":
		return icons.DCR
	case "bch":
		return icons.BCH
	case "ltc":
		return icons.LTC
	}
	return nil
```

ui/page/debug_page.go
itswisdomagain, line 150: This method had a bug on dcrlibwallet which is fixed in a new commit. Will share the commit ref to use in go.mod shortly.

ui/window.go
```diff
@@ -21,6 +22,7 @@ import (
 // Window represents the app window (and UI in general). There should only be one.
 // Window uses an internal state of booleans to determine what the window is currently displaying.
 type Window struct {
+	appCtx     context.Context
```
itswisdomagain: Doubt this is used anymore. Can remove.

ui/window.go
```diff
@@ -96,6 +98,7 @@ func CreateWindow(wal *wallet.Wallet) (*Window, *app.Window, error) {
 	win.invalidate = make(chan struct{}, 2)
 
 	win.wallet = wal
+
```
itswisdomagain: Can also undo this change so this PR doesn't register any change to this file as it is not required.

ui/page/dexclient/add_dex_widget.go
```diff
@@ -0,0 +1,182 @@
...
+		dexServerAddress: l.Theme.Editor(new(widget.Editor), "DEX Address"),
+		addDexServer:     l.Theme.Button("Submit"),
+		cert:             l.Theme.Editor(new(widget.Editor), "Cert content"),
+		createWalletWdg:  newDexCreateWalletWidget(l),
+	}
+
+	dwg.addDexServer.TextSize = values.TextSize12
+	dwg.addDexServer.Background = l.Theme.Color.Primary
+
+	dwg.dexServerAddress.Editor.SetText(testDexHost)
```
itswisdomagain: Only applicable for testnet instances:

```suggestion
	if l.WL.MultiWallet.NetType() == dcrlibwallet.Testnet3 {
		dwg.dexServerAddress.Editor.SetText(testDexHost)
	}
```

ui/page/dexclient/add_dex_widget.go
```diff
@@ -0,0 +1,182 @@
...
+	showCreateWallet bool
+}
+
+func newAddDexWidget(l *load.Load) *addDexWidget {
+	dwg := &addDexWidget{
+		Load:             l,
+		dexServerAddress: l.Theme.Editor(new(widget.Editor), "DEX Address"),
+		addDexServer:     l.Theme.Button("Submit"),
+		cert:             l.Theme.Editor(new(widget.Editor), "Cert content"),
+		createWalletWdg:  newDexCreateWalletWidget(l),
```
itswisdomagain: This modal may never be needed, why not create it only when it is needed, below just before line 84.

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,186 @@
...
+	if dex == nil {
+		return nil
+	}
+	return dex.Markets[mktName]
+}
+
+func NewMarketPage(l *load.Load) *Page {
+	pg := &Page{
+		Load:         l,
+		addDexWidget: newAddDexWidget(l),
```
itswisdomagain: Should only create this widget when necessary. Don't want to load this widget into memory when a DEX has already been added and it won't ever be used.

ui/page/main_page.go
```diff
@@ -22,15 +24,9 @@ import (
...
-
 const (
-	OverviewNavID = iota
-	TransactionsNavID
-	WalletsNavID
-	StakingNavID
-	ProposalsNavID
-	MoreNavID
+	MainPageID           = "Main"
+	HideBalanceConfigKey = "hide_balance"
```
itswisdomagain: Where'd this come from? Is it used?

itswisdomagain (changes requested):

ui/page/dexclient/markets.go
song50119: Changed it for now, but in the future can add more dex. So we'll need to init in page struct

go.mod
```diff
@@ -21,12 +22,15 @@ require (
...
 	golang.org/x/image v0.0.0-20210628002857-a66eb6448b8d
 	golang.org/x/sys v0.0.0-20210910150752-751e447fb3d0 // indirect
 	golang.org/x/text v0.3.7
 )
 
-replace github.com/decred/dcrdata/txhelpers/v4 => github.com/decred/dcrdata/txhelpers/v4 v4.0.0-20200108145420-f82113e7e212
+replace (
+	github.com/planetdecred/dcrlibwallet => github.com/itswisdomagain/dcrlibwallet v1.0.0-rc1.0.20211112050550-d0810081daec
+	github.com/planetdecred/dcrlibwallet/dexdcr => github.com/itswisdomagain/dcrlibwallet/dexdcr v0.0.0-20211112050550-d0810081daec
+)
```
itswisdomagain: New commit to fix the `Reset DEX Client` bug, and also eliminates the need to include a replace directive for the dexdcr package, a single dcrlibwallet replace is sufficient now.
```suggestion
replace github.com/planetdecred/dcrlibwallet => github.com/itswisdomagain/dcrlibwallet v1.0.0-rc1.0.20211203085346-c04212c93a13
```

ui/page/dexclient/listeners.go
```diff
@@ -0,0 +1,18 @@
...
+		select {
+		case n := <-ch:
+			if n.Type() == core.NoteTypeFeePayment {
+				pg.RefreshWindow()
+			}
+		case <-pg.ctx.Done():
+			return
+		}
+	}
+}
```
Sirmorrison: This was only used in the market.go file... Kindly move this to the market to go file.

ui/page/dexclient/add_dex_widget.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	dwg.addDexServer.TextSize = values.TextSize12
+	dwg.addDexServer.Background = l.Theme.Color.Primary
+
+	dwg.dexServerAddress.Editor.SingleLine = true
+	if l.WL.MultiWallet.NetType() == dcrlibwallet.Testnet3 {
+		dwg.dexServerAddress.Editor.SetText(testDexHost)
+	}
+
```
Sirmorrison: Not sure why we are using this here.
image
I dont see a cancel button. What if i change my mind?
Do we really need this? I believe the text_input_modal.go can be modified to use 2 input fields and used.

or you can create a better widget modal for this in the dex package just like it was done for staking purchace modal.
https://github.com/planetdecred/godcr/blob/master/ui/page/staking/purchase_modal.go

ui/page/dexclient/create_wallet_widget.go
```diff
@@ -0,0 +1,177 @@
...
+	dcw := &dexCreateWalletWidget{
+		Load:            l,
+		walletPassword:  l.Theme.EditorPassword(new(widget.Editor), "Wallet Password"),
+		appPassword:     l.Theme.EditorPassword(new(widget.Editor), "App Password"),
+		createNewWallet: l.Theme.Button("Add"),
+		walletInfoWdg:   walInfoWdg,
+		walletCreated:   walletCreated,
+	}
+
+	dcw.createNewWallet.TextSize = values.TextSize12
```
Sirmorrison: image

This widget should be a modal widget see the way it was implemented in staking page https://github.com/planetdecred/godcr/blob/master/ui/page/staking/purchase_modal.go

Sirmorrison: I also dont see a cancel button or a way to go back.

Sirmorrison (changes requested):

go.mod
```diff
@@ -21,12 +22,12 @@ require (
...
+	github.com/planetdecred/dcrlibwallet v1.6.2-0.20211102194150-7acc77e00431
 	github.com/yeqown/go-qrcode v1.5.1
 	golang.org/x/exp v0.0.0-20210722180016-6781d3edade3
 	golang.org/x/image v0.0.0-20210628002857-a66eb6448b8d
 	golang.org/x/sys v0.0.0-20210910150752-751e447fb3d0 // indirect
 	golang.org/x/text v0.3.7
 )
 
-replace github.com/decred/dcrdata/txhelpers/v4 => github.com/decred/dcrdata/txhelpers/v4 v4.0.0-20200108145420-f82113e7e212
+replace github.com/planetdecred/dcrlibwallet => github.com/itswisdomagain/dcrlibwallet v1.0.0-rc1.0.20211203085346-c04212c93a13
```
itswisdomagain: Please update to `github.com/itswisdomagain/dcrlibwallet v1.0.0-rc1.0.20211208075343-e6e3d219e32c` to fix the dex reset panic.

ui/load/load.go
```diff
@@ -59,6 +59,8 @@ type Icons struct {
 	TicketExpiredIcon,
 	TicketRevokedIcon,
 	TicketUnminedIcon *decredmaterial.Image
+
+	DexIcon, DexIconInactive, BTC, DCR, LTC, BCH *decredmaterial.Image
```
itswisdomagain: What's the plan for the LTC and BCH icons? Is there any plan to use them currently? I don't see it, we can remove these icons and add them later if and when they become useful.

itswisdomagain (changes requested):
Just a few more nits for me.

itswisdomagain (approved):
LGTM and tests well.

Sirmorrison (dismissed):
This PR looks good and merge ready, However, we are holding up on merging it for now.

ui/page/dexclient/markets.go
```diff
@@ -0,0 +1,247 @@
...
+
+type Page struct {
+	*load.Load
+	ctx                     context.Context
+	ctxCancel               context.CancelFunc
+	login                   decredmaterial.Button
+	initialize              decredmaterial.Button
+	addDex                  decredmaterial.Button
+	sync                    decredmaterial.Button
+	isConnectedToDcrNetwork bool
```
itswisdomagain: Shouldn't need this. Use `pg.WL.MultiWallet.IsConnectedToDecredNetwork()` instead where it's needed.

itswisdomagain (changes requested):
While we're here, a minor improvement.

Sirmorrison (approved):

PR #684, opened 2021-11-06
update wallet settings page to mock up

Sirmorrison:
Update wallet settings page the to updated UI.

### DIFF
--- a/ui/page/settings_page.go
+++ b/ui/page/settings_page.go
@@ -298,8 +298,7 @@ func (pg *SettingsPage) agent() layout.Widget {
 func (pg *SettingsPage) mainSection(gtx layout.Context, title string, body layout.Widget) layout.Dimensions {
 	return layout.Inset{Bottom: values.MarginPadding10}.Layout(gtx, func(gtx C) D {
 		return pg.Theme.Card().Layout(gtx, func(gtx C) D {
-			m15 := values.MarginPadding15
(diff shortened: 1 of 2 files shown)


devchoplife:
I notticed this and I am not sure if that is the expected behaviour:
The click effect only affects the text rows on wallet settings page. See image below:
Screenshot from 2021-11-08 07-33-13

Sirmorrison:
From my end, there is no hover effect.
image

dreacot:
What @devchoplife is referring to is the clickable area for the menu, it should be expanded to be same with that of currency conversion or language settings on the settings page

Sirmorrison:
> What @devchoplife is referring to is the clickable area for the menu, it should be expanded to be same with that of currency conversion or language settings on the settings page

Not exactly. He was referring to the highlight of the app labels when the button is clicked.

Also, i dont see the need for increasing the width of the click area as it has no impact on the features functionality.

devchoplife:
> > What @devchoplife is referring to is the clickable area for the menu, it should be expanded to be same with that of currency conversion or language settings on the settings page
> 
> Not exactly. He was referring to the highlight of the app labels when the button is clicked.
> 
> Also, i dont see the need for increasing the width of the click area as it has no impact on the features functionality.

I was referring to the clickable area as @dreacot  mentioned.

dreacot (approved):

PR #695, opened 2021-11-20
Fixes wallet dropdown bug

### LINKED ISSUE #689: dropdown selected account bug
When a wallet is selected from the dropdown menu, and a transaction is selected to view the tx details, returning to the main transaction page, the wallet selection is reverted back to the default.

What is expected that is that the wallet dropdown should retain the initially selected wallet before the transaction was selected to view the details.

ezgif com-gif-maker(13)

AdimekweEbuka:
Fixes #689
godcr (testnet) 2021-11-20 13-44-17

### DIFF
--- a/ui/load/wallet.go
+++ b/ui/load/wallet.go
@@ -31,7 +31,7 @@ type WalletLoad struct {
 
 	SelectedWallet  *int
 	SelectedAccount *int
-} //
+}
 
 func (wl *WalletLoad) SortedWalletList() []*dcrlibwallet.Wallet {
 	wallets := wl.MultiWallet.AllWallets()
--- a/ui/load/wallet.go
+++ b/ui/load/wallet.go
@@ -31,7 +31,7 @@ type WalletLoad struct {
 
 	SelectedWallet  *int
 	SelectedAccount *int
-}
+} //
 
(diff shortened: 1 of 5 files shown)


devchoplife (changes requested):
This fix is not implemented on tickets list page

devchoplife (dismissed):
Looks good

ui/load/wallet.go
```diff
@@ -31,7 +31,7 @@ type WalletLoad struct {
 
 	SelectedWallet  *int
 	SelectedAccount *int
-}
+} //
```
Sirmorrison: Why is the `//`?

ui/page/transaction/transaction_details_page.go
```diff
@@ -104,7 +104,7 @@ func NewTransactionDetailsPage(l *load.Load, transaction *dcrlibwallet.Transacti
 			}
 		}
 	}
-
+	//
```
Sirmorrison: What is this for?

Sirmorrison (changes requested):

ui/page/transaction/transaction_details_page.go
AdimekweEbuka: Sorry, forgot to take it out.

Sirmorrison (approved):

PR #726, opened 2021-12-21
Fix sync-related app slowness

### LINKED ISSUE #377: godcr slows down while syncing
when syncing a new wallet, the user interface seems to lag. It is understood that syncing a new wallet can be CPU intensive, however, the UI shouldn't lag when this is going on.

Machine Specifications
OS: MacOs 
Processor: 2.4 GHz Quad-Cor Intel Core i5
Memory:  8gb

itswisdomagain:
Fix slowness observed during sync, which is primarily due to
the window display being refreshed too often during sync.
Resolves #377.


ui/load/interface.go
```diff
@@ -2,12 +2,31 @@ package load
...
 import "gioui.org/layout"
 
+// Page defines methods that control the appearance and functionality of
+// UI components displayed on a window.
 type Page interface {
+	// ID is a unique string that identifies the page and may be used
+	// to differentiate this page from other pages.
 	ID() string
-	OnResume() // called when a page is starting or resuming from a paused state.
+	// WillAppear is called when the page is about to displayed and may
```
Sirmorrison: Typo on this line.
```suggestion
	// WillAppear is called when the page is about to be displayed and may
```

ui/page/about_page.go
```diff
@@ -66,14 +66,24 @@ func NewAboutPage(l *load.Load) *AboutPage {
...
 
+// ID is a unique string that identifies the page and may be used
+// to differentiate this page from other pages.
+// Part of the load.Page interface.
 func (pg *AboutPage) ID() string {
 	return AboutPageID
 }
 
-func (pg *AboutPage) OnResume() {
+// WillAppear is called when the page is about to displayed and may
```
Sirmorrison:
```suggestion
// WillAppear is called when the page is about to be displayed and may
```
i think this reoccurs everywhere.. so we can just modify it.

ui/page/debug_page.go
```diff
@@ -54,21 +54,40 @@ func NewDebugPage(l *load.Load) *DebugPage {
...
 
+// ID is a unique string that identifies the page and may be used
+// to differentiate this page from other pages.
+// Part of the load.Page interface.
 func (pg *DebugPage) ID() string {
 	return DebugPageID
 }
 
-func (pg *DebugPage) OnResume() {
+// WillAppear is called when the page is about to displayed and may
```
Sirmorrison:
```suggestion
// WillAppear is called when the page is about to be displayed and may
```

Sirmorrison (commented):
There is really no much change with the speed especially when syncing both proposals and wallet.

main.go
```diff
@@ -50,27 +50,25 @@ func main() {
...
-	go func() {
-		<-shutdown
-		wal.Shutdown()
-		os.Exit(0)
-	}()
-
-	win, appWindow, err := ui.CreateWindow(wal)
+	win, err := ui.CreateWindow(wal)
 	if err != nil {
+		wal.Shutdown() // TODO: Confirm that the wallet hasn't even been opened and thus doesn't need to be closed
```
itswisdomagain: Will revisit this and update.

ui/page/main_page.go
```diff
@@ -59,17 +59,19 @@ type MainPage struct {
 
 	hideBalanceItem HideBalanceItem
 
+	// rename to currentFragment (android term) or currentSubPage
```
itswisdomagain: Will address this.

ui/page/main_page.go
```diff
@@ -90,6 +92,7 @@ func NewMainPage(l *load.Load) *MainPage {
 			mp.StartSyncing()
 		}
 	}
+	// evaluate these
```
itswisdomagain: Will remove this.

ui/page/main_page.go
```diff
@@ -356,9 +371,16 @@ func (mp *MainPage) Handle() {
...
+// the displayed window. This method should ideally be used to disable
+// features that are irrelevant when the page is NOT displayed.
+// NOTE: The page may be re-displayed on the app's window, in which case
+// WillAppear() will be called again. This method should not destroy UI
+// components unless they'll be recreated in the WillAppear() method.
+// Part of the load.Page interface.
+func (mp *MainPage) WillDisappear() {
 	if mp.currentPage != nil {
-		mp.currentPage.OnClose()
+		mp.currentPage.WillDisappear()
```
itswisdomagain: Should also do same for `mp.sendPage` if not nil.

ui/page/main_page.go
```diff
@@ -393,17 +415,19 @@ func (mp *MainPage) popFragment() {
...
 
-		mp.currentPage.OnClose()
-		previousPage.OnResume()
+		mp.currentPage.WillDisappear()
+		previousPage.WillAppear()
 		mp.currentPage = previousPage
 	}
 }
 
+// revisit
```
itswisdomagain: Will revisit.

ui/page/wallets/restore_page.go
```diff
@@ -63,10 +64,10 @@ type Restore struct {
 	keyEvent    chan *key.Event
 }
 
-// Loading lays out the loading widget with a faded background
-func NewRestorePage(l *load.Load) *Restore {
+func NewRestorePage(l *load.Load, onRestoreComplete func()) *Restore {
```
itswisdomagain, line 67: This allows callers to determine which page should be loaded next, rather than using a channel. Need to retest to confirm that the issue that previously necessitated using a channel isn't resurrected.

Sirmorrison:
So testing this PR on my local computer it works fine, but on my VM it is slows down when navigating to another page and a new block is being fetched while syncing.

When rescanning blocks the issues is not noticed.

PR also needs a rebase.

ui/page/wallets/wallet_settings_page.go
```diff
@@ -246,15 +261,13 @@ func (pg *WalletSettingsPage) Handle() {
...
 							}
 
-							pg.RefreshWindow()
 							if pg.WL.MultiWallet.LoadedWalletsCount() > 0 {
 								pg.Toast.Notify("Wallet removed")
 								pg.PopFragment()
 							} else {
-								pg.WL.Wallet.ClearListeners()
-								pg.Load.Receiver.AllWalletsDeleted <- struct{}{}
+								pg.Load.ReloadApp()
```
itswisdomagain, line 268: `win.Loop` is now simplified to only handle actual window events, not auxiliary events like this. The method approach should achieve the same effect.

Sirmorrison (approved):

Issue #385, opened 2021-04-26
can't disconnect from the decred network


oshorefueled:
sometimes clicking the disconnect button doesn't close the connection with the decred network. It seems to disconnect and then connect back almost immediately.

![connect-bug](https://user-images.githubusercontent.com/15804688/116125546-8bd5b380-a6bd-11eb-8acb-0d11b7f78d60.gif)

oshorefueled:
The bug occurs randomly and the steps to reproduce it hasn't be defined.

dreacot:
resolved by https://github.com/planetdecred/godcr/pull/731

PR #731, opened 2021-12-27
Disable disconnect button when clicked.

### LINKED ISSUE #724: Disable disconnect button/update status during disconnection
Sometimes it takes few seconds for godcr to disconnect from the network. Without any visual update that godcr is disconnecting from the network and with the disconnect button remaining active. If the user isn't looking at the log, the user is likely to click disconnect again with the assumption that the software did not register the previous click. The button is active so each click sends a new command to dcrlibwallet. Fast successive clicks will create a condition where:

- dcrlibwallet receives multiple disconnect command, but disconnects successfully.
- It disconnects and immediately reconnect to the network, because between clicks disconnect button state changed but the user did not realize that.  Possibly related to  #385 

In the video below, I clicked the disconnect button multiple times to reproduce the condition described above. Observe the command window underneath godcr and see the clicks register on the log. Also observe godcr disconnect button state.

https://user-images.githubusercontent.com/4796738/146572487-37a580b1-1ef8-4faf-80df-18fdfe69885f.mp4

crux25:
This PR fixes #724 and #385
Summary if issue fixed. If disconnect button isn't disabled after it is clicked it leads to a condition where it is possible to click it again and create another disconnect request. The issue is documented in #724 .

### DIFF
--- a/ui/page/overview/overview_layout.go
+++ b/ui/page/overview/overview_layout.go
@@ -247,7 +247,18 @@ func (pg *AppOverviewPage) Handle() {
 		if pg.rescanningBlocks {
 			pg.WL.MultiWallet.CancelRescan()
 		} else {
-			go pg.ToggleSync()
+			// If connected to the Decred network disable button. Prevents multiple clicks.
+			if pg.isConnnected {
+				pg.syncClickable.Enabled = false
+			}
+
+			// On exit of goroutine update button state.
+			go func() {
+				pg.ToggleSync()
+				if !pg.syncClickable.Enabled {
+					pg.syncClickable.Enabled = true
+				}
+			}()
 		}
 	}
 
--- a/ui/page/overview/overview_layout.go
+++ b/ui/page/overview/overview_layout.go
@@ -249,14 +249,14 @@ func (pg *AppOverviewPage) Handle() {
 		} else {
 			// If connected to the Decred network disable button. Prevents multiple clicks.
 			if pg.isConnnected {
-				pg.syncClickable.Enabled = false
+				pg.syncClickable.SetEnabled(false, nil)
 			}
 
 			// On exit update button state.
 			go func() {
 				pg.ToggleSync()
-				if !pg.syncClickable.Enabled {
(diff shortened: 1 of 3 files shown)


ui/page/overview/sync_details.go
```diff
@@ -64,6 +64,12 @@ func (pg *AppOverviewPage) syncStatusTextRow(gtx layout.Context, inset layout.In
 		return layout.Flex{Axis: layout.Horizontal, Spacing: layout.SpaceBetween}.Layout(gtx,
 			layout.Flexed(1, syncStatusLabel.Layout),
 			layout.Rigid(func(gtx C) D {
+
+				// Set gxt to Disabled (Sets Queue to nil) if syncClickable state is disabled, prevents double click.
+				if !pg.syncClickable.Enabled {
+					gtx = gtx.Disabled()
+				}
```
Sirmorrison: Why not handle the disabled functionality from the clickable widget? `pg.syncClicable.SetEnable(enabled bool)`

ui/page/overview/overview_layout.go
```diff
@@ -247,7 +247,18 @@ func (pg *AppOverviewPage) Handle() {
 		if pg.rescanningBlocks {
 			pg.WL.MultiWallet.CancelRescan()
 		} else {
-			go pg.ToggleSync()
+			// If connected to the Decred network disable button. Prevents multiple clicks.
+			if pg.isConnnected {
+				pg.syncClickable.Enabled = false
+			}
```
Sirmorrison:
```suggestion
			if pg.isConnnected {
				pg.syncClickable.SetEnabled(false)
			}
```

Sirmorrison (dismissed):
It is better to handle the state from the widget so we dont need to repeat this for all pages that may need the clickable widget disabled or enabled state.

ui/page/overview/sync_details.go
crux25: Nice idea, I had thought about that but realised it require heavy modification of the widget.

dreacot (approved):

PR #742, opened 2021-12-31
Add rebroadcast button for unpublished transactions 

### LINKED ISSUE #700: a rebroadcast button should be added for transactions that were created, but not published
Sometimes transactions are created, but not published to the network, a rebroadcast button should be added to the transaction details page of such transactions which calls the `wallet.PublishUnminedTransactions()` function

Screenshot from 2021-11-22 17-16-59

A screenshot from a similar implementation on dcrandroid

from discussions with @itswisdomagain,
I quote "I believe I've seen a similar issue. In my case, the tx was never actually broadcasted because of low fee, so it doesn't exist on the blockchain. The rescan will identify the outputs that were spent in that tx as unspent since there's no tx on the blockchain that spends those outputs. A rebroadcast button is a good attempt at fixing it but there's no guarantee the tx will eventually be accepted. I've actually seen the issue a couple times and the ideal fix I believe will be preventing low fees when creating txs on dcrlibwallet."

devchoplife:
Closes #700 
This PR adds rebroadcast button to transactions that have not been published to the network 

Image:
Screenshot from 2021-12-31 13-41-52


dreacot (commented):
I get this screen on the transaction details page

Screenshot 2021-12-31 at 1 48 55 PM

dreacot:
from the screenshot you posted, let's redesign the rebroadcast button to be like the sync button

ui/page/transaction/transaction_details_page.go
```diff
@@ -46,6 +46,9 @@ type TxDetailsPage struct {
 	inputsCollapsible               *decredmaterial.Collapsible
 	backButton                      decredmaterial.IconButton
 	infoButton                      decredmaterial.IconButton
+	rebroadcast                     decredmaterial.Label
+	rebroadcastClicable             *decredmaterial.Clickable
```
dreacot:
```suggestion
	rebroadcastClickable             *decredmaterial.Clickable
```

ui/page/transaction/transaction_details_page.go
```diff
@@ -82,8 +88,11 @@ func NewTransactionDetailsPage(l *load.Load, transaction *dcrlibwallet.Transacti
 		destAddressClickable:      new(widget.Clickable),
 		toDcrdata:                 l.Theme.NewClickable(true),
 
-		transaction: transaction,
-		wallet:      l.WL.MultiWallet.WalletWithID(transaction.WalletID),
+		transaction:         transaction,
+		wallet:              l.WL.MultiWallet.WalletWithID(transaction.WalletID),
+		rebroadcast:         rebroadcast,
+		rebroadcastClicable: l.Theme.NewClickable(true),
+		rebroadcastIcon:     decredmaterial.NewIcon(l.Icons.Cached),
```
dreacot: update the icon, to be same with mobile

ui/page/transaction/transaction_details_page.go
```diff
@@ -695,6 +730,11 @@ func (pg *TxDetailsPage) Handle() {
 			pg.ChangeFragment(NewTransactionDetailsPage(pg.Load, pg.ticketSpent))
 		}
 	}
+
+	if pg.rebroadcastClicable.Clicked() {
+		pg.rebroadcastClicable.SetEnabled(false, nil)
+		pg.wallet.PublishUnminedTransactions()
+	}
```
dreacot:
```suggestion
	if pg.rebroadcastClicable.Clicked() {
		pg.rebroadcastClicable.SetEnabled(false, nil)
		if !pg.Load.WL.MultiWallet.IsConnectedToDecredNetwork() {
			pg.Toast.NotifyError("not connected to the decred network") // there should be a localized string for this already, i'm guessing
			return
		}
		err := pg.wallet.PublishUnminedTransactions()
		if err != nil {
		    //display the error to the user, and if possible log it
		    pg.Toast.NotifyError(err.Error())
		} else {
		    pg.Toast.Notify("Republished unmined transactions to the decred network")
		}
	}
```

dreacot (changes requested):
The correct icon should be used

Screenshot 2021-12-31 at 7 06 38 PM

-i get no feedback when i click on the rebroadcast button

ui/page/transaction/transaction_details_page.go
dreacot: also the `wallet.PublishUnminedTransactions()` is a blocking function, so if you notice UI lags/freeze when spamming the rebroadcast button, it might be better to put the function in a goroutine(but only when UI lags are confirmed)

```
	if pg.rebroadcastClicable.Clicked() {
		go func() {
			pg.rebroadcastClicable.SetEnabled(false, nil)
			if !pg.Load.WL.MultiWallet.IsConnectedToDecredNetwork() {
				pg.Toast.NotifyError("not connected to the decred network") // there should be a localized string for this already, i'm guessing
				return
			}
			err := pg.wallet.PublishUnminedTransactions()
			if err != nil {
				//display the error to the user, and if possible log it
				pg.Toast.NotifyError(err.Error())
			} else {
				pg.Toast.Notify("Republished unmined transactions to the decred network")
			}
		}()
	}
```

ui/page/transaction/transaction_details_page.go
```diff
@@ -50,6 +50,7 @@ type TxDetailsPage struct {
 	rebroadcastClickable            *decredmaterial.Clickable
 	rebroadcastIcon                 *decredmaterial.Image
 	gtx                             *layout.Context
+	unmined                         bool
```
dreacot: this variable isn't needed, because republishing the transaction doesn't automatically make it mind, it just adds it to a queue of transactions to the network, and it might still get rejected

ui/page/transaction/transaction_details_page.go
```diff
@@ -731,21 +732,27 @@ func (pg *TxDetailsPage) Handle() {
 	}
 
 	if pg.rebroadcastClickable.Clicked() {
-		pg.rebroadcastClickable.SetEnabled(false, nil)
-		if !pg.Load.WL.MultiWallet.IsConnectedToDecredNetwork() {
-			// if user is not conected to the network, notify the user
-			pg.Toast.NotifyError("Not connected to the decred network")
-			return
-		}
+		if pg.unmined {
```
dreacot: this check isn't necessary

ui/page/transaction/transaction_details_page.go
```diff
@@ -695,6 +730,29 @@ func (pg *TxDetailsPage) Handle() {
...
+				pg.Toast.NotifyError("Not connected to the decred network")
+				return
+			}
+
+			err := pg.wallet.PublishUnminedTransactions()
+			if err != nil {
+				// If transactions are not published, notify the user
+				pg.Toast.NotifyError(err.Error())
+			} else {
+				pg.Toast.NotifyError("Republished unmined transactions to the decred network")
```
dreacot:
```suggestion
				pg.Toast.Notify("Republished unmined transactions to the decred network")
```

ui/page/transaction/transaction_details_page.go
```diff
@@ -695,6 +730,29 @@ func (pg *TxDetailsPage) Handle() {
...
+			}
+		}
+	}
+
+	if pg.transaction.BlockHeight > -1 {
+		pg.unmined = false
+		pg.rebroadcastClickable.SetEnabled(false, nil)
+	}
```
dreacot: this isn't needed as the button won't be visible if the height is greater than -1

dreacot (changes requested):

ui/page/transaction/transaction_details_page.go
```diff
@@ -247,6 +247,9 @@ func (pg *TxDetailsPage) txnBalanceAndStatus(gtx layout.Context) layout.Dimensio
 						layout.Rigid(func(gtx C) D {
 							if pg.transaction.BlockHeight == -1 {
 								pg.rebroadcastClickable.SetEnabled(true, nil)
```
dreacot: this line can be deleted, its enabled by default

dreacot (changes requested):

dreacot (approved):

PR #749, opened 2022-01-04
Ensure that amount is revalidated when source account is changed

### LINKED ISSUE #743: amount input field on the send page doesn't re-validate after changing the sending account
I tried sending funds from a wallet with insufficient balance, and i got the error below.
Also note that the next button is active even though the validation failed, that shouldn't be.
Screenshot from 2021-12-31 14-00-27

I then switched to a sending account with sufficient balance and was still getting the insufficient amount error, this is because the validation was not redone when a new sending account was selected
Screenshot from 2021-12-31 13-58-41

I had to clear the amount input field and re input the amount for the error to disappear

devchoplife:
Closes #743 
This PR:

- Ensures that send amount is revalidated when source account is changed 
- Keep the send button grey until amount is succesfully validated
- Fixes bug that prevents input on amount editors on send page

### DIFF
--- a/ui/page/components/account_selector.go
+++ b/ui/page/components/account_selector.go
@@ -25,10 +25,12 @@ type AccountSelector struct {
 
 	openSelectorDialog *decredmaterial.Clickable
 
-	wallets            []*dcrlibwallet.Wallet
-	selectedAccount    *dcrlibwallet.Account
-	selectedWalletName string
-	totalBalance       string
+	wallets               []*dcrlibwallet.Wallet
+	selectedAccount       *dcrlibwallet.Account
+	selectedWalletName    string
+	totalBalance          string
+	defaultAcccountNumber int32
+	isDefaultSet          bool
 }
 
 func NewAccountSelector(l *load.Load) *AccountSelector {
@@ -39,7 +41,8 @@ func NewAccountSelector(l *load.Load) *AccountSelector {
 		accountIsValid:     func(*dcrlibwallet.Account) bool { return true },
 		openSelectorDialog: l.Theme.NewClickable(true),
 
-		wallets: l.WL.SortedWalletList(),
+		wallets:      l.WL.SortedWalletList(),
+		isDefaultSet: false,
 	}
 }
 
@@ -58,7 +61,22 @@ func (as *AccountSelector) AccountSelected(callback func(*dcrlibwallet.Account))
 	return as
 }
 
+func (as *AccountSelector) Changed() bool {
+	if as.defaultAcccountNumber != as.selectedAccount.Number {
+		as.defaultAcccountNumber = as.selectedAccount.Number
+		return true
+	}
+
+	return false
+}
+
 func (as *AccountSelector) Handle() {
+	if !as.isDefaultSet {
+		as.defaultAcccountNumber = as.selectedAccount.Number
+		as.isDefaultSet = true
+	}
+	as.Changed()
+
 	for as.openSelectorDialog.Clicked() {
 		newAccountSelectorModal(as.Load, as.selectedAccount, as.wallets).
 			title(as.dialogTitle).
--- a/ui/page/components/account_selector.go
+++ b/ui/page/components/account_selector.go
@@ -25,12 +25,11 @@ type AccountSelector struct {
 
 	openSelectorDialog *decredmaterial.Clickable
 
-	wallets               []*dcrlibwallet.Wallet
-	selectedAccount       *dcrlibwallet.Account
-	selectedWalletName    string
-	totalBalance          string
-	defaultAcccountNumber int32
-	isDefaultSet          bool
+	wallets            []*dcrlibwallet.Wallet
+	selectedAccount    *dcrlibwallet.Account
+	selectedWalletName string
+	totalBalance       string
+	changed            bool
 }
 
 func NewAccountSelector(l *load.Load) *AccountSelector {
@@ -41,8 +40,7 @@ func NewAccountSelector(l *load.Load) *AccountSelector {
 		accountIsValid:     func(*dcrlibwallet.Account) bool { return true },
 		openSelectorDialog: l.Theme.NewClickable(true),
 
-		wallets:      l.WL.SortedWalletList(),
-		isDefaultSet: false,
+		wallets: l.WL.SortedWalletList(),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


ui/page/send/page.go
```diff
@@ -438,6 +440,18 @@ func (pg *Page) Handle() {
...
+	if pg.currentAccountNumber != pg.defaultAcccountNumber {
+		pg.defaultAcccountNumber = pg.currentAccountNumber
+		pg.amount.validateDCRAmount()
+	}
+
+	if pg.validate() {
+		fmt.Println("Validated")
+	} else {
+		fmt.Println("Not validated")
+	}
```
Sirmorrison: is this needed?

Sirmorrison (changes requested):

ui/page/send/page.go
devchoplife: It was a debugging code I ommited when cleaning up code

ui/page/send/page.go
```diff
@@ -52,9 +52,11 @@ type Page struct {
...
 	moreOptionIsOpen bool
 
-	exchangeRate   float64
-	usdExchangeSet bool
-	exchangeError  string
+	exchangeRate          float64
+	usdExchangeSet        bool
+	exchangeError         string
+	defaultAcccountNumber int32
+	currentAccountNumber  int32
```
Sirmorrison: Rather than have this here, you can simple add an `OnChanged()` to the account selector modal, and returns a bool when ever the account changes.

ui/page/send/page.go
```diff
@@ -153,6 +155,7 @@ func (pg *Page) OnResume() {
 		pg.usdExchangeSet = false
 	}
 	pg.Load.EnableKeyEvent = true
+	pg.defaultAcccountNumber = pg.sourceAccountSelector.SelectedAccount().Number
```
Sirmorrison: same applies here.

ui/page/send/page.go
```diff
@@ -429,14 +432,10 @@ func (pg *Page) Handle() {
...
-		pg.amount.setError("")
-		if currencyValue == values.USDExchangeValue {
-			pg.amount.dcrAmountEditor.Editor.SetText("")
-			pg.amount.usdAmountEditor.Editor.SetText("")
-		} else {
-			pg.amount.dcrAmountEditor.Editor.SetText("")
-		}
+	pg.currentAccountNumber = pg.sourceAccountSelector.SelectedAccount().Number
+	if pg.currentAccountNumber != pg.defaultAcccountNumber {
+		pg.defaultAcccountNumber = pg.currentAccountNumber
```
Sirmorrison: this then will be 
```suggestion
        // this is just a sample code
	if pg.sourceAccountSelector.Changed(){
```

Sirmorrison (changes requested):

ui/page/send/page.go
```diff
@@ -383,60 +383,57 @@ func (pg *Page) Handle() {
 		}
 	}
 
-	validAddress := pg.sendDestination.validate()
+	// if destination switch is equal to Address
 	if pg.sendDestination.sendToAddress {
-		addEditor := pg.sendDestination.destinationAddressEditor
-		if validAddress {
+		if pg.sendDestination.validate() {
+			fmt.Println("Validated")
```
Sirmorrison: Remove.

Sirmorrison (changes requested):

ui/page/components/account_selector.go
```diff
@@ -25,10 +25,12 @@ type AccountSelector struct {
...
 
-	wallets            []*dcrlibwallet.Wallet
-	selectedAccount    *dcrlibwallet.Account
-	selectedWalletName string
-	totalBalance       string
+	wallets               []*dcrlibwallet.Wallet
+	selectedAccount       *dcrlibwallet.Account
+	selectedWalletName    string
+	totalBalance          string
+	defaultAcccountNumber int32
```
Sirmorrison: not needed remove please.

ui/page/components/account_selector.go
```diff
@@ -25,10 +25,12 @@ type AccountSelector struct {
...
+	isDefaultSet          bool
```
Sirmorrison:
```suggestion
	changed          bool
```

ui/page/components/account_selector.go
```diff
@@ -39,7 +41,8 @@ func NewAccountSelector(l *load.Load) *AccountSelector {
 		accountIsValid:     func(*dcrlibwallet.Account) bool { return true },
 		openSelectorDialog: l.Theme.NewClickable(true),
 
-		wallets: l.WL.SortedWalletList(),
+		wallets:      l.WL.SortedWalletList(),
+		isDefaultSet: false,
```
Sirmorrison: remove

ui/page/components/account_selector.go
```diff
@@ -58,7 +61,22 @@ func (as *AccountSelector) AccountSelected(callback func(*dcrlibwallet.Account))
 	return as
 }
 
+func (as *AccountSelector) Changed() bool {
+	if as.defaultAcccountNumber != as.selectedAccount.Number {
+		as.defaultAcccountNumber = as.selectedAccount.Number
+		return true
+	}
+
+	return false
```
Sirmorrison:
```suggestion
	changed := as.changed
	as.changed = false
	return changed
```

ui/page/components/account_selector.go
```diff
@@ -58,7 +61,22 @@ func (as *AccountSelector) AccountSelected(callback func(*dcrlibwallet.Account))
...
+}
+
 func (as *AccountSelector) Handle() {
+	if !as.isDefaultSet {
+		as.defaultAcccountNumber = as.selectedAccount.Number
+		as.isDefaultSet = true
+	}
+	as.Changed()
+
```
Sirmorrison: Remove

ui/page/components/account_selector.go
```diff
@@ -58,7 +61,22 @@ func (as *AccountSelector) AccountSelected(callback func(*dcrlibwallet.Account))
...
 	for as.openSelectorDialog.Clicked() {
 		newAccountSelectorModal(as.Load, as.selectedAccount, as.wallets).
 			title(as.dialogTitle).
```
Sirmorrison, line 71: replace this with

```
for as.openSelectorDialog.Clicked() {
		newAccountSelectorModal(as.Load, as.selectedAccount, as.wallets).
			title(as.dialogTitle).
			accountValidator(as.accountIsValid).
			accountSelected(func(account *dcrlibwallet.Account) {
				if as.selectedAccount.Number != account.Number{
					as.changed = true
				}
				as.setupSelectedAccount(account)
				as.callback(account)
			}).Show()
	}
```

Sirmorrison (changes requested):

Sirmorrison:
So an end test case.

When on currency = none, and i am to send max without an address, i get prompted to set an address since it is needed to calculate the max balance.

When this address is removed, the max balance is still active.
Navigating to `My account`, the max amount is set since it is still active from the address tab.
navigating back to `address`, the max amount is returned even without an address set.

ezgif com-gif-maker(4)

Sirmorrison (approved):

PR #753, opened 2022-01-06
Send page password/Key event Bug

### LINKED ISSUE #739: can't input password on send page confirmation modal
while trying to send funds to my other account, i'm unable to input my password into the confirmation modal, the input editor keeps going out of focus

Screenshot 2021-12-30 at 1 53 33 PM

crux25:
This PR restructures how Key events are delivered to the modals and Pages, fixes the bug where pages and modals listen to key events on the same channel and try to handle the same events simultaneously. It also fixes #739  where the password confirmation editor goes out of focus when sending DCR to accounts.

### DIFF
--- a/ui/load/load.go
+++ b/ui/load/load.go
@@ -76,21 +76,21 @@ type Load struct {
 
 	Toast *notification.Toast
 
-	SelectedWallet            *int
-	SelectedAccount           *int
-	SelectedUTXO              map[int]map[int32]map[string]*wallet.UnspentOutput
+	SelectedWallet  *int
+	SelectedAccount *int
+	SelectedUTXO    map[int]map[int32]map[string]*wallet.UnspentOutput
 
-	ToggleSync       func()
-	RefreshWindow    func()
-	ShowModal        func(Modal)
-	DismissModal     func(Modal)
-	ChangeWindowPage func(page Page, keepBackStack bool)
-	PopWindowPage    func() bool
-	ChangeFragment   func(page Page)
-	PopFragment      func()
-	PopToFragment    func(pageID string)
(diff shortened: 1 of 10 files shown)


Sirmorrison (changes requested):
On the send page, when the send confirmation modal is opened, and the tab Key is pressed a couple of times, queues up the key events and closing the modal makes the tabs to jump.

ezgif com-gif-maker(2)

crux25:
> 

It is functioning according to design. The cursor goes between the input fields because the events was queued up when the App was pulling exchange rate and tx fees from network.

Sirmorrison:
> > 
> 
> It is functioning according to design. The cursor goes between the input fields because the events was queued up when the App was pulling exchange rate and tx fees from network.

This is a wrong design then because if i push the tab 20 times when the confirm modal is open, closing the modal, the tab event goes round all the input fields until the 20 queued tab events are completed. When the Confirmation modal is opened, there should be no queued event on the send page.

crux25:
> > > 
> > 
> > 
> > It is functioning according to design. The cursor goes between the input fields because the events was queued up when the App was pulling exchange rate and tx fees from network.
> 
> This is a wrong design then because if i push the tab 20 times when the confirm modal is open, closing the modal, the tab event goes round all the input fields until the 20 queued tab events are completed. When the Confirmation modal is opened, there should be no queued event on the send page.

I didn't realize you mean that the send page is receiving tab events for the confirm password input. 
I'll fix it.

Sirmorrison:
ezgif com-gif-maker(3)

Same behaviour is noticed when i switch to my address, and tap the tab button a couple of time when currency conversion is set to `none` and change back to address.

Sirmorrison (approved):

PR #756, opened 2022-01-08
Fix wallet option bug

### LINKED ISSUE #727: wallet option menu, doesn't show in full when there are multiple wallets
I have 5 wallets, and i needed to delete the last wallet on the list, on clicking the wallet options menu, i noticed, i wasn't able to see the full information on the menu, and there was no way to view it except going into full screen 

Screenshot from 2021-12-22 09-54-32

AdimekweEbuka:
Fixes #727 
godcr (mainnet) 2022-01-08 23-02-19

### DIFF
--- a/ui/decredmaterial/collapsible.go
+++ b/ui/decredmaterial/collapsible.go
@@ -160,7 +160,3 @@ func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C) 
 func (c *CollapsibleWithOption) MoreTriggered() bool {
 	return c.moreIconButton.Button.Clicked()
 }
-
-func (c *CollapsibleWithOption) IsExpanded() bool {
-	return c.isExpanded
-}
--- a/ui/decredmaterial/collapsible.go
+++ b/ui/decredmaterial/collapsible.go
@@ -160,3 +160,7 @@ func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C) 
 func (c *CollapsibleWithOption) MoreTriggered() bool {
 	return c.moreIconButton.Button.Clicked()
 }
+
+func (c *CollapsibleWithOption) IsExpanded() bool {
+	return c.isExpanded
+}
--- a/ui/page/wallets/wallet_page.go
+++ b/ui/page/wallets/wallet_page.go
@@ -2,10 +2,11 @@ package wallets
 
 import (
 	"context"
+	"fmt"
 	"image"
 	"image/color"
+	"math"
 	"sync"
-	"fmt"
 
 	"gioui.org/gesture"
 	"gioui.org/io/pointer"
@@ -75,8 +76,8 @@ type WalletPage struct {
 	addWalletMenu  []menuItem
 
 	container   *widget.List
-	backdrop                 *widget.Clickable
-		walletsList layout.List
+	backdrop    *widget.Clickable
+	walletsList layout.List
 
 	walletIcon               *decredmaterial.Image
 	walletAlertIcon          *decredmaterial.Image
@@ -94,12 +95,11 @@ type WalletPage struct {
 	separator            decredmaterial.Line
 
 	mt    unit.Value    // option menu top margin padding
-	click gesture.Click // page click event listener
+	click gesture.Click // page click event
 
 	openPopupIndex      int
 	isAddWalletMenuOpen bool
 	hasWatchOnly        bool
-	isExpanded bool
 }
 
 func NewWalletPage(l *load.Load) *WalletPage {
@@ -113,7 +113,7 @@ func NewWalletPage(l *load.Load) *WalletPage {
 		watchWalletsList:         l.Theme.NewClickableList(layout.Vertical),
 		card:                     l.Theme.Card(),
 		backdrop:                 new(widget.Clickable),
-				openAddWalletPopupButton: l.Theme.NewClickable(false),
+		openAddWalletPopupButton: l.Theme.NewClickable(false),
 		openPopupIndex:           -1,
 		shadowBox:                l.Theme.Shadow(),
 		separator:                l.Theme.Separator(),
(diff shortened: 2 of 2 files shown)


dreacot (changes requested):
The first wallet option no longer shows properly
Screenshot 2022-01-10 at 12 43 29 AM

The wallet popup menu, should open up at the point the mouse was clicked as it did before, the popup position should only change if there isn't enough room to display it, like in the case of the last wallet item

AdimekweEbuka:
> The first wallet option no longer shows properly Screenshot 2022-01-10 at 12 43 29 AM
> 
> The wallet popup menu, should open up at the point the mouse was clicked as it did before, the popup position should only change if there isn't enough room to display it, like in the case of the last wallet item

Resolved
godcr (mainnet) 2022-01-10 01-26-39

Sirmorrison (changes requested):
When i have only one wallet, or for wallets at the top, the menu covers the send and receive button.

It will be nice to make this menu dynamic that is the menu should only be inverted on the last item at the bottom of the screen or the last 2 iteams from the bottom of the screen.

image

Sirmorrison (dismissed):
ezgif com-gif-maker(6)

So this seems to kind of work when you only have the required amount of Wallet for the current page. 

If I have more than 4 wallet, the 4th wallet option is hidden. but when I scroll, to the bottom, everything seems ok.

As a side note, I will recommend you make use of the gioui `layout.List` struct position properties and also its scroll properties.

This will tell you when the page has been scrolled to the bottom or not.

Sirmorrison:
getting the following error.
```
Config file (/Users/sirmorrison/Library/Application Support/Godcr/godcr.conf) does not exist. Using defaults.
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x4ac6ac4]

goroutine 7 [running]:
github.com/planetdecred/dcrlibwallet.(*MultiWallet).ReadUserConfigValue(0x0, {0x5177536, 0x9}, {0x4eb4a80, 0xc000134070})
	/Users/sirmorrison/go/pkg/mod/github.com/planetdecred/dcrlibwallet@v1.6.2-0.20211208155954-3ba1589ff83f/multiwallet_config.go:64 +0x44
github.com/planetdecred/dcrlibwallet.(*MultiWallet).ReadBoolConfigValueForKey(...)
	/Users/sirmorrison/go/pkg/mod/github.com/planetdecred/dcrlibwallet@v1.6.2-0.20211208155954-3ba1589ff83f/multiwallet_config.go:110
github.com/planetdecred/godcr/ui/load.(*Load).RefreshTheme(0xc0002cc380)
	/Users/sirmorrison/Desktop/go projects/godcr-gio/ui/load/load.go:139 +0x5a
github.com/planetdecred/godcr/ui/page.(*startPage).OnResume(0xc0002ce2c0)
	/Users/sirmorrison/Desktop/go projects/godcr-gio/ui/page/start_page.go:52 +0x66
github.com/planetdecred/godcr/ui.(*Window).Start(0xc0001dac00)
	/Users/sirmorrison/Desktop/go projects/godcr-gio/ui/window.go:157 +0x42
github.com/planetdecred/godcr/ui.(*Window).Loop(0xc0001dac00, 0xc0006db200, 0x0)
	/Users/sirmorrison/Desktop/go projects/godcr-gio/ui/window.go:343 +0x1db
created by main.main
	/Users/sirmorrison/Desktop/go projects/godcr-gio/main.go:81 +0x505
```
Kindly have PR rebased.

Sirmorrison:
ezgif com-gif-maker(1)

There still exit a level of unexpected behavior.

Sirmorrison:
There seem to be a regression.
image

Kindly revert the last changes as it seems to be way better than what we currently have.

dreacot:
Screenshot from 2022-03-14 17-26-43

Screenshot from 2022-03-14 17-28-01

I don't see any progress with this PR, are there any blockers encountered?

AdimekweEbuka:
> Screenshot from 2022-03-14 17-26-43
> 
> Screenshot from 2022-03-14 17-28-01
> 
> I don't see any progress with this PR, are there any blockers encountered?

Currently, I am waiting for @Sirmorrison's go ahead on how we should proceed on this. The last commit before the changes were reverted, I believe, fixed the initial issue using the position and size of the wallet option to ensure that no part of it was hidden regardless of the amount of wallets but it did lead to some other concerns as inorder to achieve that the wallet option had to appear to the side when clicked and also the size was reduced. I also believe there are some other reservations on how we should proceed. I am currently waiting on @Sirmorrison for a way forward.

dreacot (commented):
This looks good, however there's a miscalculation when the account section is expanded, once that is fixed, then this is good to merge 👍

Screenshot from 2022-03-19 16-05-51
Screenshot from 2022-03-19 16-05-36

AdimekweEbuka:
From my observation the issue still persists on some wallets depending on the amount of wallets on the screen, please view video below.
godcr (testnet) 2022-03-20 20-12-53

Sirmorrison:
> From my observation the issue still persists on some wallets depending on the amount of wallets on the screen, please view video below. godcr (testnet) 2022-03-20 20-12-53

I get your point however, gioui does not currently implement css events.
The workaround currently implemented checks the list length against the click position in percentage. Should the click position be below 58% of the list length(which ideally is the distance before some section starts to hide, the sets the inset.

As long as there is room at the bottom to scroll the page and the click position with regards to the list length is not up to 58%, the uses the default inset.

devchoplife:
> > From my observation the issue still persists on some wallets depending on the amount of wallets on the screen, please view video below. godcr (testnet) 2022-03-20 20-12-53
> 
> I get your point however, gioui does not currently implement css events. The workaround currently implemented checks the list length against the click position in percentage. Should the click position be below 58% of the list length(which ideally is the distance before some section starts to hide, the sets the inset.
> 
> As long as there is room at the bottom to scroll the page and the click position with regards to the list length is not up to 58%, the uses the default inset.

In that case, I think it is good to go 
I tested again and there is enough room to scroll down to view the full option set so it is fine

dreacot (approved):

PR #759, opened 2022-01-13
make watch-only wallets visible on wallets page

itswisdomagain:
Watch-only wallets do not currently show up in the wallets page. This fixes that.

### DIFF
--- a/ui/decredmaterial/button.go
+++ b/ui/decredmaterial/button.go
@@ -106,6 +106,13 @@ func (t *Theme) IconButton(icon *widget.Icon) IconButton {
 	}
 }
 
+func (t *Theme) IconButtonWithStyle(ibs IconButtonStyle, colorStyle *values.ColorStyle) IconButton {
+	return IconButton{
+		ibs,
+		colorStyle,
+	}
+}
+
 func (b *Button) SetClickable(clickable *widget.Clickable) {
 	b.clickable = clickable
 }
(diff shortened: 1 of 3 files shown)


ui/decredmaterial/button.go
```diff
@@ -195,6 +202,8 @@ func (bl ButtonLayout) Layout(gtx layout.Context, w layout.Widget) layout.Dimens
 }
 
 // TODO: Test to ensure this works!
+// TODO: Doesn't workk, if ib.colorStyle was nil before this method is called,
+// it is temporarily changed but when ib.Layout is called, it returns to nil.
 func (ib IconButton) ChangeColorStyle(colorStyle *values.ColorStyle) {
```
itswisdomagain: We'll need an issue to investigate other uses of this method which likely won't be working as expected.

itswisdomagain:
Can test with this pub key `tpubVokqVVuZgvPWNuVQXUPMt3znUFJMqArBtuy4m8Xw2GGqBdXCWRqvPua21adM99n43CTn3dMnXnVVrncqkeCs4GukzgdxPRaFF2L322vdog6`.

ui/decredmaterial/button.go
```diff
@@ -195,6 +202,8 @@ func (bl ButtonLayout) Layout(gtx layout.Context, w layout.Widget) layout.Dimens
 }
 
 // TODO: Test to ensure this works!
+// TODO: Doesn't work, if ib.colorStyle was nil before this method is called,
+// it is temporarily changed but when ib.Layout is called, it returns to nil.
```
Sirmorrison, line 206: When will these `TODO`s be resolved? is this something that can go in to this PR?

Sirmorrison (changes requested):
The PR looks good... Just a few changes.
The option pop up width is too much to the right.
image

it should also have same alignment with the wallet pop up.
**wallet**
image

**watch only**
image

Sirmorrison:
Please add a description to this PR.

ui/decredmaterial/button.go
itswisdomagain: Will need to go into a separate issue to be examined separately.

Sirmorrison (approved):
LGTM

PR #764, opened 2022-01-20
update deps to rc3

itswisdomagain:
NOTE:
- Running this will update your wallet db versions; may want to back up before testing.
- Mixing may no longer work due to a change upstream. Requires a few changes to dcrlibwallet. Coming soon.


dreacot (approved):
LGTM

Issue #728, opened 2021-12-22
sync doesn't cancel when creating new wallet


dreacot:
It was noticed on godcr that while syncing a wallet, and attempting to create a new wallet, the log output gets stuck on `[INF] DLWL: Canceling sync. May take a while for sync to fully cancel.`, it's important to know the UI is still usable when this happens, quitting the app and reopening shows that the new wallet was never created.

The error was easily reproduced on a fresh install of godcr with no prior databases.

This dcrlibwallet PR https://github.com/planetdecred/dcrlibwallet/pull/156 added the capability for the library to automatically cancel sync while creating/deleting a wallet, as opposed to doing that manually in the client apps.

The change was implemented into dcrandroid by this PR https://github.com/planetdecred/dcrandroid/pull/534, and the functionality was tested to work as intended.

The change was implemented into dcrios by this PR https://github.com/planetdecred/dcrios/pull/736/files, and the functionality was tested to work as intended.

Issue #765, opened 2022-01-20
Window stuck if you remove a wallet during sync.


monsa00:
If a secondary wallet is removed during sync, the window will become frozen.

``` 2022-01-20 12:35:00.950 [INF] DLWL: Canceling sync. May take a while for sync to fully cancel. ```

May be related to https://github.com/planetdecred/godcr/pull/255

<img width="384" alt="Screen Shot 2022-01-20 at 12 48 33 PM" src="https://user-images.githubusercontent.com/66281245/150419860-e3797dab-de71-46ab-b719-dbec43769416.png">

itswisdomagain:
Maybe related https://github.com/planetdecred/godcr/issues/680.

dreacot:
Seems related to this also https://github.com/planetdecred/godcr/issues/728

crux25:
> Seems related to this also #728

Yes it is related to #728. I've seen this bug while fixing #728.

PR #775, opened 2022-01-27
Godcr stuck in cancel sync

crux25:
This PR close issue #728 and #765.

### DIFF
--- a/ui/page/wallets/wallet_page.go
+++ b/ui/page/wallets/wallet_page.go
@@ -88,6 +88,9 @@ type WalletPage struct {
 	backupAcctIcon           *decredmaterial.Icon
 
 	listLock sync.Mutex
+
+	ctx       context.Context
+	ctxCancel context.CancelFunc
 }
 
 func NewWalletPage(l *load.Load) *WalletPage {
@@ -151,6 +154,7 @@ func (pg *WalletPage) ID() string {
 func (pg *WalletPage) OnNavigatedTo() {
 	pg.ctx, pg.ctxCancel = context.WithCancel(context.TODO())
 	pg.listenForTxNotifications()
+	pg.listenForSyncNotifications()
 	pg.loadWalletAndAccounts()
 }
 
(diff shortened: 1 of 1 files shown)


ui/page/wallets/wallet_page.go
```diff
@@ -979,4 +987,32 @@ func (pg *WalletPage) HandleUserInteractions() {
 // Part of the load.Page interface.
 func (pg *WalletPage) OnNavigatedFrom() {
 	pg.closePopups()
+	pg.ctxCancel()
+}
+
+// listenForsyncNotifications spuns a go-routine that listens for
+// Wallet notifications. Currently we are interested in reading from
+// NotificationsUpdate channel, to prevent the channel from getting
+// filled thereby blocking Wallet create/delete actions.
```
itswisdomagain, line 1156: Seems to me that this is a bad design. Why do we have this channel in the first place and why do we need to drain it this way to prevent it from blocking?

crux25: The channel provides sync status updates mainly used by the overview page. If it navigates away from the overview page the channel is no longer consumed, the channel gets filled up and block.

crux25: I see use cases on the wallet page too. e.g Updating the status of operations on the UI, but that's beyond the scope of the current task.

itswisdomagain: It'd be better for pages that require such updates to create the channel and close it when done with it. A universal channel that is not always used and could become blocked isn't ideal. I'll propose some changes in a moment.

crux25: > It'd be better for pages that require such updates to create the channel and close it when done with it. A universal channel that is not always used and could become blocked isn't ideal. I'll propose some changes in a moment.

I had thought about that, but a single universal channel is used to deliver tx/mixer/politeia/sync/rescan notifications. Changes will require breaking this up into units, so pages can create the units they need and close them on exit. This changes the scope of the task from fixing the bugs mentioned in the PR to refactoring the channels used to deliver notifications. Pointing that out before I proceed.

PR #777, opened 2022-01-30
Refactor notification channel

### LINKED ISSUE #728: sync doesn't cancel when creating new wallet

### LINKED ISSUE #765: Window stuck if you remove a wallet during sync.

crux25:
This PR refactors how mixer/politeia/sync and rescan notifications are delivered. Fixes Godcr stuck in cancelling sync, 
Fix #728 and Fix #765 
Close #775


listeners/politeianotification.go
```diff
@@ -0,0 +1,51 @@
...
+		ProposalStatus: wallet.VoteFinished,
+		Proposal:       proposal,
+	}
+	pn.sendNotification(update)
+}
+
+func (pn *PoliteiaNotification) sendNotification(signal wallet.Proposal) {
+	if pn.PoliteiaNotifCh != nil {
+		pn.PoliteiaNotifCh <- signal
+	}
```
itswisdomagain: Doesn't make sense to have a listener with a nil channel. If the channel were nil for some reason, the code should crash. Others that I've seen in this PR don't have this if protection.

listeners/syncprogress.go
```diff
@@ -0,0 +1,82 @@
+package listeners
+
+import (
+	"github.com/planetdecred/dcrlibwallet"
+	"github.com/planetdecred/godcr/wallet"
+)
+
+type SyncProgress struct {
```
itswisdomagain: Should be `SyncProgressListener`. Same for the others; these are all listeners and are used that way with the different multiwallet methods for registering listeners.

listeners/syncprogress.go
```diff
@@ -0,0 +1,82 @@
...
+		ProgressReport: cfiltersFetchProgress,
+	})
+}
+
+func (sp *SyncProgress) OnHeadersFetchProgress(headersFetchProgress *dcrlibwallet.HeadersFetchProgressReport) {
+	sp.sendNotification(wallet.SyncStatusUpdate{
+		Stage: wallet.HeadersFetchProgress,
+		ProgressReport: wallet.SyncHeadersFetchProgress{
+			Progress: headersFetchProgress,
+		},
```
itswisdomagain: Please let's improve this while we're here; the following just doesn't make sense:
```
type SyncHeadersFetchProgress struct {
	Progress *dcrlibwallet.HeadersFetchProgressReport
}
```

Either do `type SyncHeadersFetchProgress *dcrlibwallet.HeadersFetchProgressReport` or just use `*dcrlibwallet.HeadersFetchProgressReport` directly.

itswisdomagain: Applies to a few other such types in the wallet/listener.go file.

listeners/syncprogress.go
```diff
@@ -0,0 +1,82 @@
...
+		Stage: wallet.SyncCanceled,
+	})
+}
+func (sp *SyncProgress) OnSyncEndedWithError(err error)          {}
+func (sp *SyncProgress) Debug(debugInfo *dcrlibwallet.DebugInfo) {}
+
+func (sp *SyncProgress) sendNotification(signal wallet.SyncStatusUpdate) {
+	if sp.SyncStatus != nil {
+		sp.SyncStatus <- signal
+	}
```
itswisdomagain: Don't think this check is necessary. Channel should never be nil.

listeners/syncprogress.go
```diff
@@ -0,0 +1,82 @@
+package listeners
+
+import (
+	"github.com/planetdecred/dcrlibwallet"
+	"github.com/planetdecred/godcr/wallet"
+)
+
+type SyncProgress struct {
+	SyncStatus chan wallet.SyncStatusUpdate
```
itswisdomagain: `SyncStatusChan`. Append `Chan` always (for other listeners in this pkg as well) so when this variable is used in code in other sites, the variable type is quite immediately obvious.

listeners/txandblocknotification.go
```diff
@@ -0,0 +1,50 @@
...
+func NewTxAndBlockNotification(txAndBlockNotif chan TxNotification) *TxAndBlockNotification {
+	return &TxAndBlockNotification{
+		TxAndBlockNotifCh: txAndBlockNotif,
+	}
+}
+
+func (txAndBlk *TxAndBlockNotification) OnTransaction(transaction string) {
+	var tx dcrlibwallet.Transaction
+	err := json.Unmarshal([]byte(transaction), &tx)
+	if err == nil {
```
itswisdomagain: Log non-nil errors please.

ui/load/load.go
```diff
@@ -86,6 +85,8 @@ type Load struct {
 	SubscribeKeyEvent   func(eventChan chan *key.Event, pageID string) // Widgets call this function to recieve key events.
 	UnsubscribeKeyEvent func(pageID string) error
 	ReloadApp           func()
+	DesktopNotifier     func(notifier interface{})
+	UpdateBalance       func()
```
itswisdomagain: Comments please. I recognise this hasn't been the custom, but it'd help a code reader understand why these properties exist without having to dig to find where and how they're used.

ui/page/governance/proposal_details_page.go
```diff
@@ -163,6 +169,11 @@ func (pg *ProposalDetails) listenForSyncNotifications() {
 // Part of the load.Page interface.
 func (pg *ProposalDetails) OnNavigatedFrom() {
 	pg.ctxCancel()
+
+	if pg.PoliteiaNotifCh != nil {
+		close(pg.PoliteiaNotifCh)
+	}
+	pg.WL.MultiWallet.Politeia.RemoveNotificationListener(ProposalDetailsPageID)
```
itswisdomagain: Remove before closing the channel. Also makes sense to set `pg.PoliteiaNotifCh` to nil, wouldn't be receiving any notifications anymore until you recall `MultiWallet.Politeia.AddNotificationListener`.

itswisdomagain: Similar considerations for other uses of notification listeners.

ui/page/main_page.go
```diff
@@ -664,3 +660,65 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
 		}),
 	)
 }
+
+func (mp *MainPage) desktopNotifier(notifier interface{}) {
```
itswisdomagain: Isn't it better to have this page create its own notification listeners and do as it pleases rather than rely on sub pages calling this method when necessary? Also makes the code clearer that way.

ui/load/load.go
itswisdomagain: Update: I'm sure we can do without them even.

itswisdomagain (changes requested):
Initial review. Idea looks solid. Will review again after these are addressed.

ui/page/main_page.go
crux25: Those are not listeners, It's a func used to post desktop notification from the main page. I have added comment to the code to clarify that section of the code.

listeners/accountmixerNotif.go
```diff
@@ -0,0 +1,37 @@
+package listeners
+
+import (
+	"github.com/planetdecred/godcr/wallet"
+)
```
Sirmorrison:
```suggestion
import  "github.com/planetdecred/godcr/wallet"
```

ui/page/components/account_selector.go
```diff
@@ -33,14 +36,16 @@ type AccountSelector struct {
 	selectedWalletName string
 	totalBalance       string
 	changed            bool
+	subbedTxNotif      bool
```
Sirmorrison: by `subbedTxNotif` you mean? maybe an inline comment can help out here.
also i think `subbedTxNotification` is better.

ui/page/components/account_selector.go
```diff
@@ -200,39 +205,59 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
 }
 
 func (as *AccountSelector) ListenForTxNotifications(ctx context.Context) {
+	// Complain if it try to listen for TX Notifications
+	// When not subscribed.
+	if !as.subbedTxNotif {
+		log.Warn("Account Selector not subscribed for Tx notifications but listening")
+	}
```
Sirmorrison: Oh i get it `subbedTxNotif` is subscribed? 
Thinking again, i think `IsTxNotificationEnabled` is better.

I don't think this is needed as the users do not get to see the message anyways to act on it. Plus it is not an error.

listeners/txandblocknotification.go
```diff
@@ -0,0 +1,54 @@
...
+	update := TxNotification{
+		NotificationType: NewTx,
+		Transaction:      &tx,
+	}
+	txAndBlk.UpdateNotification(update)
+}
+
+func (txAndBlk *TxAndBlockNotificationListener) OnBlockAttached(walletID int, blockHeight int32) {
+	txAndBlk.UpdateNotification(TxNotification{
+		NotificationType: BlkAttached,
```
Sirmorrison:
```suggestion
		NotificationType: BlockAttached,
```

listeners/types.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"github.com/planetdecred/dcrlibwallet"
+)
+
+type TxNotifType int
+
+const (
+	// Transaction notification types
+	NewTx       TxNotifType = iota // 0 = New transaction.
+	BlkAttached                    // 1 = block attached.
```
Sirmorrison:
```suggestion
	BlockAttached                    // 1 = block attached.
```

ui/page/components/account_selector.go
```diff
@@ -200,39 +205,59 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
...
-					as.selectorModal.setupWalletAccounts()
-				}
-				as.RefreshWindow()
-			case wallet.SyncStatusUpdate:
-				// refresh wallet account and balance on every new block
-				// only if sync is completed.
-				if n.Stage == wallet.BlockAttached && as.WL.MultiWallet.IsSynced() {
+			case n := <-as.TxAndBlockNotifChan:
+				switch n.NotificationType {
+				case listeners.BlkAttached:
```
Sirmorrison:
```suggestion
				case listeners.BlockAttached:
```

ui/page/components/account_selector.go
```diff
@@ -200,39 +205,59 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
...
+					// refresh wallet account and balance on every new block
+					// only if sync is completed.
+					if as.WL.MultiWallet.IsSynced() {
+						as.UpdateSelectedAccountBalance()
+						if as.selectorModal != nil {
+							as.selectorModal.setupWalletAccounts()
+						}
+						as.RefreshWindow()
+					}
+				case listeners.NewTx:
```
Sirmorrison:
```suggestion
				case listeners.NewTransaction:
```

listeners/types.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/planetdecred/dcrlibwallet"
+)
+
+type TxNotifType int
+
+const (
+	// Transaction notification types
+	NewTx       TxNotifType = iota // 0 = New transaction.
```
Sirmorrison:
```suggestion
	NewTransaction       TxNotifType = iota // 0 = New transaction.
```

ui/page/components/account_selector.go
```diff
@@ -200,39 +205,59 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
...
 		}
 	}()
 }
 
+// SubscribeTxNotifications registers AccountSelector for tx and
+// Notifications from dcrlibwallet.
+func (as *AccountSelector) SubscribeTxNotifications() {
+	log.Info("Account selector subbed")
+	as.WL.MultiWallet.AddTxAndBlockNotificationListener(as.TxAndBlockNotificationListener, true, AccoutSelectorID)
+	as.subbedTxNotif = true
```
Sirmorrison:
```suggestion
	as.IsTxNotificationEnabled = true
```

ui/page/components/account_selector.go
```diff
@@ -200,39 +205,59 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
...
+}
+
+// UnsubscribeTxNotifications removes AccountSelector from tx and
+// Notifications from dcrlibwallet.
+func (as *AccountSelector) UnsubscribeTxNotifications() {
+	log.Info("Account selector unsubscribed!!!")
+	as.WL.MultiWallet.RemoveTxAndBlockNotificationListener(AccoutSelectorID)
+	// Close notification channel.
+	close(as.TxAndBlockNotifChan)
+	as.subbedTxNotif = false
```
Sirmorrison: `IsTxNotificationEnabled`

ui/window.go
```diff
@@ -118,15 +118,18 @@ func (win *Window) NewLoad() (*load.Load, error) {
...
 		},
 
 		Toast: notification.NewToast(th),
 
 		Printer: message.NewPrinter(language.English),
 	}
 
+	l.Receiver = &load.Receiver{
+		KeyEvents: win.keyEvents,
+	}
```
Sirmorrison: This looks to be repeated as we have same on line 120.

Sirmorrison (changes requested):

ui/page/components/account_selector.go
crux25: My idea is not to have it there for the user. It is there for debugging, just like the rest of the debugging messages.

Sirmorrison: Question is what are you debugging? in `SubscribeTxNotifications()`, you log sub and the set `subbedTxNotif` to true. 

I believe the `subbedTxNotif` is actually redundant and can be removed.

Sirmorrison (commented):
Functionality looks good to me.

listeners/accountmixerNotif.go
```diff
@@ -0,0 +1,35 @@
+package listeners
+
+import "github.com/planetdecred/godcr/wallet"
+
+// AccountMixerNotificationListener satisfies dcrlibwallet AccountMixerNotificationListener
+// interface contract. Consumers interested in mixer notification instantiates this type.
```
itswisdomagain: Over the max 80 char per line limit. Also, just `interface` not `interface contract`.
A couple other cases like this as well.

listeners/log.go
```diff
@@ -0,0 +1,25 @@
...
+// See LICENSE for details.
+
+package listeners
+
+import "github.com/decred/slog"
+
+// log is a logger that is initialized with no output filters.  This
+// means the package will not perform any logging by default until the caller
+// requests it.
+var apiLog = slog.Disabled
```
itswisdomagain: Is `apiLog` used?

log.go
```diff
@@ -48,10 +50,12 @@ var (
...
-	walletLog = backendLog.Logger("WALL")
-	winLog    = backendLog.Logger("UI")
-	dlwlLog   = backendLog.Logger("DLWL")
-	pageLog   = backendLog.Logger("PAGE")
+	walletLog  = backendLog.Logger("WALL")
+	winLog     = backendLog.Logger("UI")
+	dlwlLog    = backendLog.Logger("DLWL")
+	pageLog    = backendLog.Logger("PAGE")
+	lstnersLog = backendLog.Logger("LSTNER")
+	cmpntLog   = backendLog.Logger("CMPNT")
```
itswisdomagain: These codes are usually 4 characters long from what I've seen in practice.

ui/load/load.go
```diff
@@ -83,6 +82,8 @@ type Load struct {
 	SubscribeKeyEvent   func(eventChan chan *key.Event, pageID string) // Widgets call this function to recieve key events.
 	UnsubscribeKeyEvent func(pageID string) error
 	ReloadApp           func()
+	DesktopNotifier     func(notifier interface{}) // Widgets call this function to post desktop notifications.
+	UpdateBalance       func()                     // Widgets call this function to update total balance shown on the main page.s
```
itswisdomagain: These comments are surely helpful. A previous comment recommended NOT using these however. Reason is, from what I've seen, these are only set in the mainpage and are only called when a transaction/block notification is received. Instead of having subpages call these methods when a new tx or block is observed, let the mainpage register a listener for itself and call these methods directly when needed.

ui/page/components/account_selector.go
```diff
@@ -200,39 +205,57 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
 }
 
 func (as *AccountSelector) ListenForTxNotifications(ctx context.Context) {
+	// Complain if it try to listen for TX Notifications
+	// When not subscribed.
+	if !as.IsTxNotificationEnabled {
+		log.Warn("Account Selector not subscribed for Tx notifications but listening")
+	}
```
itswisdomagain: Does this happen currently? I'd rather this method register the listener and proceed to listen for notifications in a goroutine. That ensures that the goroutine doesn't just lie idle, never receiving a notification.

ui/page/components/account_selector.go
```diff
@@ -29,18 +32,20 @@ type AccountSelector struct {
...
 
-	dialogTitle        string
-	selectedWalletName string
-	totalBalance       string
-	changed            bool
+	dialogTitle             string
+	selectedWalletName      string
+	totalBalance            string
+	changed                 bool
+	IsTxNotificationEnabled bool // Indicates if the account selector is enabled to receive Tx and block notifications.
```
itswisdomagain: I'm quite worried about this. The account selector widget should always be listening for tx and block notifications.

ui/page/components/account_selector.go
```diff
@@ -200,39 +205,57 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
...
+// SubscribeTxNotifications registers AccountSelector for tx and
+// Notifications from dcrlibwallet.
+func (as *AccountSelector) SubscribeTxNotifications() {
+	as.WL.MultiWallet.AddTxAndBlockNotificationListener(as.TxAndBlockNotificationListener, true, AccoutSelectorID)
+	as.IsTxNotificationEnabled = true
+}
+
+// UnsubscribeTxNotifications removes AccountSelector from tx and
+// Notifications from dcrlibwallet.
+func (as *AccountSelector) UnsubscribeTxNotifications() {
```
itswisdomagain: This should automatically happen when the `ctx` passed to `ListenForTxNotifications` is closed. Shouldn't need these new methods `SubscribeTxNotifications` and `UnsubscribeTxNotifications`. The `ListenForTxNotifications` method which already existed and was primarily meant for tx/block notif listening, can do all these. When called, it registers the listener and then starts the goroutine to listen for notifications. When `ctx` is canceled, it unregisters the listener and exits the goroutine.

ui/page/components/log.go
```diff
@@ -0,0 +1,25 @@
...
+// See LICENSE for details.
+
+package components
+
+import "github.com/decred/slog"
+
+// log is a logger that is initialized with no output filters.  This
+// means the package will not perform any logging by default until the caller
+// requests it.
+var apiLog = slog.Disabled
```
itswisdomagain: Not used.

ui/page/dexclient/create_wallet_modal.go
```diff
@@ -90,10 +90,12 @@ func (md *createWalletModal) Dismiss() {
 
 func (md *createWalletModal) OnDismiss() {
 	md.ctxCancel()
+	md.sourceAccountSelector.UnsubscribeTxNotifications()
```
itswisdomagain: Explains a previous comment about automatically unregistering the listener when the ctx is canceled. Don't need to be a separate method.

ui/page/dexclient/create_wallet_modal.go
```diff
@@ -90,10 +90,12 @@ func (md *createWalletModal) Dismiss() {
...
 }
 
 func (md *createWalletModal) OnResume() {
 	md.ctx, md.ctxCancel = context.WithCancel(context.TODO())
+	md.sourceAccountSelector.SubscribeTxNotifications()
 	md.sourceAccountSelector.ListenForTxNotifications(md.ctx)
```
itswisdomagain: Explains a previous comment about automatically subscribing the listener when `ListenForTxNotifications` is called. Don't need to be a separate method.

ui/page/governance/proposal_details_page.go
```diff
@@ -98,6 +100,13 @@ func (pg *ProposalDetails) ID() string {
 func (pg *ProposalDetails) OnNavigatedTo() {
 	pg.ctx, pg.ctxCancel = context.WithCancel(context.TODO())
 	pg.listenForSyncNotifications()
+
+	if pg.ProposalNotificationListener == nil {
+		pg.ProposalNotificationListener = listeners.NewProposalNotificationListener(make(chan wallet.Proposal, 4))
```
itswisdomagain: I wouldn't mind doing this always, regardless of whether `pg.ProposalNotificationListener` is nil or not. It's practically the same thing and saves 3 lines of code.

listeners/txandblocknotification.go
```diff
@@ -0,0 +1,54 @@
...
+// TxAndBlockNotificationListener satisfies dcrlibwallet TxAndBlockNotificationListener interface contract.
+type TxAndBlockNotificationListener struct {
+	TxAndBlockNotifChan chan TxNotification
+}
+
+func NewTxAndBlockNotificationListener(txAndBlockNotif chan TxNotification) *TxAndBlockNotificationListener {
+	return &TxAndBlockNotificationListener{
+		TxAndBlockNotifChan: txAndBlockNotif,
+	}
+}
```
itswisdomagain: This comment applies to other similar instances. I don't think the caller needs to provide a channel, it can be created here and the caller can access it using `listener.NotifChan`. In almost all cases where this kind of function has been called in this PR, the value passed in isn't special and can easily be used directly here.

itswisdomagain (changes requested):

listeners/accountmixerNotif.go
crux25: 80 char limit per line for comments or both comments and code? Please link me to the style guide that specify this limit.

listeners/types.go
```diff
@@ -0,0 +1,23 @@
...
+const (
+	// Transaction notification types
+	NewTransaction TxNotifType = iota // 0 = New transaction.
+	BlockAttached                     // 1 = block attached.
+	TxConfirmed                       // 2 = Transaction confirmed.
+)
+
+// TxNotification models transaction notifications.
+type TxNotification struct {
+	NotificationType TxNotifType
```
itswisdomagain:
```suggestion
type TxNotification struct {
	Type TxNotifType
```

To read as `TxNotification.Type` instead of `TxNotification.NotificationType`.

ui/page/components/account_selector.go
```diff
@@ -200,34 +204,35 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
 }
 
 func (as *AccountSelector) ListenForTxNotifications(ctx context.Context) {
+	as.WL.MultiWallet.AddTxAndBlockNotificationListener(as.TxAndBlockNotificationListener, true, AccoutSelectorID)
```
itswisdomagain: This might return an error. If a non-nil error is returned, log the error AND do not start the goroutine.

itswisdomagain: I'd say do `as.TxAndBlockNotificationListener = listeners.NewTxAndBlockNotificationListener()` here instead of in `NewAccountSelector`. The notif channel may have been closed if this method was previously called with a ctx that was canceled. Also, should ensure that either `as.TxAndBlockNotificationListener == nil` or the channel is closed before recreating/assigning `as.TxAndBlockNotificationListener` to avoid starting a fresh listener when the previous one hasn't been closed.

ui/page/main_page.go
```diff
@@ -192,14 +202,12 @@ func (mp *MainPage) initNavItems() {
 // Part of the load.Page interface.
 func (mp *MainPage) OnNavigatedTo() {
 	// register for notifications, unregister when the page disappears
```
itswisdomagain: This comment seems misplaced, the next line doesn't register for notifications.

ui/page/main_page.go
```diff
@@ -192,14 +202,12 @@ func (mp *MainPage) initNavItems() {
...
-	mp.WL.MultiWallet.AddAccountMixerNotificationListener(mp, MainPageID)
-	mp.WL.MultiWallet.Politeia.AddNotificationListener(mp, MainPageID)
-	mp.WL.MultiWallet.AddTxAndBlockNotificationListener(mp, true, MainPageID) // notification methods will be invoked asynchronously to prevent potential deadlocks
-	mp.WL.MultiWallet.AddSyncProgressListener(mp, MainPageID)
-	mp.WL.MultiWallet.SetBlocksRescanProgressListener(mp)
-
+	mp.WL.Wallet.SaveConfigValueForKey(load.SeedBackupNotificationConfigKey, false)
```
itswisdomagain: I think this line of code has been removed from master.

ui/page/main_page.go
```diff
@@ -411,10 +422,9 @@ func (mp *MainPage) OnNavigatedFrom() {
 		mp.receivePage.OnNavigatedFrom()
 	}
 
-	mp.WL.MultiWallet.RemoveAccountMixerNotificationListener(MainPageID)
-	mp.WL.MultiWallet.Politeia.RemoveNotificationListener(MainPageID)
-	mp.WL.MultiWallet.RemoveTxAndBlockNotificationListener(MainPageID)
-	mp.WL.MultiWallet.RemoveSyncProgressListener(MainPageID)
+	mp.WL.Wallet.SaveConfigValueForKey(load.SeedBackupNotificationConfigKey, false)
```
itswisdomagain: I think this line of code has been removed from master.

ui/page/main_page.go
```diff
@@ -666,3 +676,131 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
 		}),
 	)
 }
+
+// desktopNotifier posts desktop notifications.
+func (mp *MainPage) desktopNotifier(notifier interface{}) {
```
itswisdomagain: `postDesktopNotification` / `postNotificationOnDesktop` / `postNotification`.

ui/page/main_page.go
```diff
@@ -666,3 +676,131 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
...
+
+	err = beeep.Notify("Decred Godcr Wallet", n, filepath.Join(absoluteWdPath, "ui/assets/decredicons/qrcodeSymbol.png"))
+	if err != nil {
+		log.Info("could not initiate desktop notification, reason:", err.Error())
+	}
+}
+
+// listenForSyncNotifications starts a goroutine to watch for sync updates
+// and update the UI accordingly.
+func (mp *MainPage) listenForSyncNotifications() {
```
itswisdomagain: listenForNotifications, this doesn't only listen for sync notifications.

ui/page/staking/auto_ticket_modal.go
```diff
@@ -68,7 +68,6 @@ func (tb *ticketBuyerModal) OnCancel(cancel func()) *ticketBuyerModal {
 func (tb *ticketBuyerModal) OnResume() {
 	tb.initializeAccountSelector()
 	tb.ctx, tb.ctxCancel = context.WithCancel(context.TODO())
-	tb.accountSelector.ListenForTxNotifications(tb.ctx)
```
itswisdomagain: Why are we removing this? Mistake maybe?

ui/page/staking/purchase_modal.go
```diff
@@ -88,7 +88,6 @@ func (tp *stakingModal) OnResume() {
 	tp.initializeAccountSelector()
 
 	tp.ctx, tp.ctxCancel = context.WithCancel(context.TODO())
-	tp.accountSelector.ListenForTxNotifications(tp.ctx)
```
itswisdomagain, line 90: This change intended or a mistake?

ui/page/wallets/privacy_page.go
```diff
@@ -446,5 +447,6 @@ func (pg *PrivacyPage) listenForMixerNotifications() {
 // components unless they'll be recreated in the OnNavigatedTo() method.
 // Part of the load.Page interface.
 func (pg *PrivacyPage) OnNavigatedFrom() {
-	pg.ctxCancel()
+	pg.WL.MultiWallet.RemoveAccountMixerNotificationListener(PrivacyPageID)
+	close(pg.MixerChan)
```
itswisdomagain: I might have missed this while reviewing previous pages with similar setup, but please apply this wherever a page has a ctx that is created `OnNavigatedTo` and canceled `OnNavigatedFrom`. The listener should be removed after the context is canceled, by the same goroutine that checks if the ctx is canceled via `case <-ctx.Done():`. Only then should the listener be removed and the channel closed. The ctx should always be canceled in `OnNavigatedFrom`, to ensure the goroutine that was started to listen for the notifications is not left running indefinitely.

itswisdomagain (changes requested):
Looking much better, stopping review here however as I suspect I'd probably be repeating review comments if I go further. Will re-review after these.

ui/page/main_page.go
crux25: I did not add this comment. I'll remove it since it isn't relevant to that section of the code.

ui/page/main_page.go
crux25: It was removed.

log.go
```diff
@@ -61,6 +75,13 @@ func init() {
 	dcrlibwallet.UseLogger(dlwlLog)
 	page.UseLogger(pageLog)
 	load.UseLogger(log)
+	listeners.UseLogger(lstnersLog)
+	components.UseLogger(cmpntLog)
+	transaction.UseLogger(txntLog)
+	governance.UseLogger(govnLog)
+	walletPage.UseLogger(walpLog)
+	overview.UseLogger(ovewLog)
+	staking.UseLogger(stakLog)
```
itswisdomagain: I don't think individual pages should have separate loggers, they should use the UI logger.

ui/page/components/account_selector.go
```diff
@@ -200,34 +203,44 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
 }
 
 func (as *AccountSelector) ListenForTxNotifications(ctx context.Context) {
+	if as.TxAndBlockNotificationListener == nil {
+		as.TxAndBlockNotificationListener = listeners.NewTxAndBlockNotificationListener()
+	}
```
itswisdomagain: To prevent duplicate listeners being sent, this should abort/return `if as.TxAndBlockNotificationListener != nil`. This means that when this listening is done, also set `as.TxAndBlockNotificationListener` to nil after closing the channel.

ui/page/components/account_selector.go
```diff
@@ -200,34 +203,44 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
...
+
+	err := as.WL.MultiWallet.AddTxAndBlockNotificationListener(as.TxAndBlockNotificationListener, true, AccoutSelectorID)
+	if err != nil {
+		log.Errorf("Error Adding Tx notifications listener: %+v", err)
```
itswisdomagain: Not sure what's going on with the casing here. Also, any specific reason you're using `%+v` rather than just `%v'?
```suggestion
		log.Errorf("Error adding tx notifications listener: %v", err)
```

ui/page/components/account_selector.go
```diff
@@ -200,34 +203,44 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
...
+					// refresh accounts list when new transaction is received
 					as.UpdateSelectedAccountBalance()
 					if as.selectorModal != nil {
 						as.selectorModal.setupWalletAccounts()
 					}
 					as.RefreshWindow()
 				}
+			case <-ctx.Done():
+				as.WL.MultiWallet.RemoveTxAndBlockNotificationListener(AccoutSelectorID)
+				// Close notification channel.
```
itswisdomagain: Not sure this comment is necessary. It's quite obvious what's happening in the next line.

ui/page/governance/proposal_details_page.go
```diff
@@ -129,26 +131,32 @@ func (pg *ProposalDetails) HandleUserInteractions() {
...
 					proposal, err := pg.WL.MultiWallet.Politeia.GetProposalRaw(pg.proposal.Token)
 					if err == nil {
 						pg.proposal = proposal
 						pg.RefreshWindow()
 					}
 				}
+			case <-pg.ctx.Done():
+				pg.WL.MultiWallet.Politeia.RemoveNotificationListener(ProposalDetailsPageID)
+				close(pg.ProposalNotifChan)
+
```
itswisdomagain: I think you should also set `pg.ProposalNotificationListener` to nil so that if this method is recalled, it'll be re-initialized instead of trying to use a closed channel.

ui/page/governance/proposals_page.go
```diff
@@ -383,26 +385,29 @@ func (pg *ProposalsPage) layoutStartSyncSection(gtx C) D {
 }
 
 func (pg *ProposalsPage) listenForSyncNotifications() {
+	pg.ProposalNotificationListener = listeners.NewProposalNotificationListener()
```
itswisdomagain, line 389: I think it's a good idea to abort if this listener isn't currently nil, to prevent duplicate listeners. Also, set the listener to nil after listening is done.

ui/page/main_page.go
```diff
@@ -666,3 +672,150 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
...
+
+	err = beeep.Notify("Decred Godcr Wallet", n, filepath.Join(absoluteWdPath, "ui/assets/decredicons/qrcodeSymbol.png"))
+	if err != nil {
+		log.Info("could not initiate desktop notification, reason:", err.Error())
+	}
+}
+
+// listenForSyncNotifications starts a goroutine to watch for sync updates
+// and update the UI accordingly.
+func (mp *MainPage) listenForNotifications() {
```
itswisdomagain: Update comment.

ui/page/main_page.go
```diff
@@ -666,3 +672,150 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
...
+	// Setup listeners
+	if mp.SyncProgressListener == nil {
+		mp.SyncProgressListener = listeners.NewSyncProgress()
+	}
```
itswisdomagain: If ANY of the listeners is non-nil, abort. Always recreate the listeners; which also mean you should always set them to nil after closing their channels.

ui/page/main_page.go
```diff
@@ -666,3 +672,150 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
 		}),
 	)
 }
+
+// postDdesktopNotification posts notifications to the desktop.
+func (mp *MainPage) postDesktopNotification(notifier interface{}) {
+	proposalNotification := mp.WL.Wallet.ReadBoolConfigValueForKey(load.ProposalNotificationConfigKey)
```
itswisdomagain: Move this to inside `case wallet.Proposal:`.

ui/page/main_page.go
```diff
@@ -666,3 +672,150 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
...
+		for {
+			select {
+			case n := <-mp.TxAndBlockNotifChan:
+				switch n.Type {
+				case listeners.NewTransaction:
+					mp.UpdateBalance(false)
+					transactionNotification := mp.WL.Wallet.ReadBoolConfigValueForKey(load.TransactionNotificationConfigKey)
+					update := wallet.NewTransaction{
+						Transaction: n.Transaction,
+					}
```
itswisdomagain: Move this to inside the `if` block below.

itswisdomagain (changes requested):
Stopped midway too to avoid repeating comments, some of these comments will apply to other files.

log.go
crux25: Don't you think It'll be difficult to track down the particular source of an error if all the pages use UI logger?

itswisdomagain: The individual page can/should add context, such as fmt.Errorf("wallets page: error doing so and so: %v", err). So, we don't have to create separate logs for each page. The UI logger should cover all UI-related errors, I believe.

crux25: I can understand. But isn't it better not to rely on hardcoded page name in the error log. I'll refactor it but ultimately I believe it is more succinct if individual page have different logger.

ui/page/main_page.go
```diff
@@ -666,3 +672,158 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
...
+	}
+
+	err = beeep.Notify("Decred Godcr Wallet", n, filepath.Join(absoluteWdPath, "ui/assets/decredicons/qrcodeSymbol.png"))
+	if err != nil {
+		log.Info("could not initiate desktop notification, reason:", err.Error())
+	}
+}
+
+// listenForNotifications starts a goroutine to watch for sync updates
+// and update the UI accordingly.
```
itswisdomagain: This method does not only watch for sync updates.

ui/page/main_page.go
```diff
@@ -666,3 +672,158 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
...
+func (mp *MainPage) listenForNotifications() {
+	// Return if any of the listener is not nill.
```
itswisdomagain:
```suggestion
	// Return if any of the listener is not nil.
```

ui/page/overview/overview_layout.go
```diff
@@ -372,13 +389,46 @@ func (pg *AppOverviewPage) listenForSyncNotifications() {
 				case wallet.SyncStarted:
 					fallthrough
 				case wallet.SyncCanceled:
-					fallthrough
+					pg.loadTransactions()
+					pg.RefreshWindow()
```
itswisdomagain: Why not `fallthrough`? It's one line shorter and maintains the `fallthrough` sequence from the previous case to the last.

ui/page/overview/overview_layout.go
```diff
@@ -372,13 +389,46 @@ func (pg *AppOverviewPage) listenForSyncNotifications() {
...
 					pg.RefreshWindow()
-				case wallet.BlockAttached:
+				}
+
+			case n := <-pg.TxAndBlockNotifChan:
+				switch n.Type {
+				case listeners.NewTransaction:
+					pg.loadTransactions()
+					pg.RefreshWindow()
+				}
```
itswisdomagain: Missing a `BlockAttached` case to refresh the window.

ui/page/main_page.go
```diff
@@ -666,3 +672,158 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
...
+		return
+	}
+
+	go func() {
+		for {
+			select {
+			case n := <-mp.TxAndBlockNotifChan:
+				switch n.Type {
+				case listeners.NewTransaction:
+					mp.UpdateBalance(false)
```
itswisdomagain: It seems master does `mp.UpdateBalance(true)`, not sure why though.

ui/page/main_page.go
```diff
@@ -666,3 +672,158 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
...
+						update := wallet.NewTransaction{
+							Transaction: n.Transaction,
+						}
+						mp.postDesktopNotification(update)
+					}
+					mp.RefreshWindow()
+				case listeners.BlockAttached:
+					beep := mp.WL.Wallet.ReadBoolConfigValueForKey(dcrlibwallet.BeepNewBlocksConfigKey)
+					if beep {
+						_ = beeep.Beep(5, 1)
```
itswisdomagain: Master logs the error:
```
err := beeep.Beep(5, 1)
if err != nil {
	log.Error(err.Error())
}
```

ui/page/main_page.go
```diff
@@ -666,3 +672,158 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
...
+					}
+
+					mp.UpdateBalance(false)
```
itswisdomagain: Master does `mp.UpdateBalance(true)`.

ui/page/main_page.go
```diff
@@ -666,3 +672,158 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
...
+					mp.RefreshWindow()
+				case listeners.TxConfirmed:
+					mp.UpdateBalance(false)
```
itswisdomagain: `mp.UpdateBalance(true)`

ui/page/main_page.go
```diff
@@ -666,3 +672,158 @@ func (mp *MainPage) LayoutTopBar(gtx layout.Context) layout.Dimensions {
...
+					mp.RefreshWindow()
+
+				}
+			case notification := <-mp.ProposalNotifChan:
+				// Don't notify on wallet synced event.
```
itswisdomagain: Might read better as
`Post desktop notification for all events except the Synced event.`

ui/page/staking/list_page.go
```diff
@@ -93,29 +93,40 @@ func (pg *ListPage) OnNavigatedTo() {
...
 			select {
-			case notification = <-pg.Receiver.NotificationsUpdate:
+			case n := <-pg.TxAndBlockNotifChan:
+				if n.Type == listeners.BlockAttached {
+					selectedWallet := pg.wallets[pg.walletDropDown.SelectedIndex()]
+					if selectedWallet.ID == n.WalletID {
+						pg.fetchTickets()
+						pg.RefreshWindow()
+					}
+				} else if n.Type == listeners.NewTransaction {
```
itswisdomagain: `if  n.Type == listeners.BlockAttached || n.Type == listeners.NewTransaction`

ui/page/wallets/privacy_page.go
```diff
@@ -412,18 +415,19 @@ func (pg *PrivacyPage) showModalPasswordStartAccountMixer() {
 }
 
 func (pg *PrivacyPage) listenForMixerNotifications() {
+	if pg.AccountMixerNotificationListener == nil {
+		pg.AccountMixerNotificationListener = listeners.NewAccountMixerNotificationListener()
+	}
```
itswisdomagain: `pg.AccountMixerNotificationListener` might not be nil but the channel might be closed, causing issues later when a channel send/read is attempted. Use the previously recommended partner of setting listeners to nil after closing their channels and returning if the listener is non-nil.

itswisdomagain (changes requested):
Will test after these and if testing checks out, should be good to merge.

ui/page/overview/overview_layout.go
```diff
@@ -376,9 +393,43 @@ func (pg *AppOverviewPage) listenForSyncNotifications() {
...
+
+			case n := <-pg.BlockRescanChan:
+				if n.Stage == wallet.RescanEnded {
+					pg.RefreshWindow()
+				}
+			case <-pg.ctx.Done():
+				pg.WL.MultiWallet.RemoveSyncProgressListener(OverviewPageID)
+				pg.WL.MultiWallet.RemoveTxAndBlockNotificationListener(OverviewPageID)
+				pg.WL.MultiWallet.Politeia.RemoveNotificationListener(OverviewPageID)
```
itswisdomagain, line 419: Also, do `pg.WL.MultiWallet.SetBlocksRescanProgressListener(nil)`.

ui/page/wallets/privacy_page.go
```diff
@@ -412,18 +415,21 @@ func (pg *PrivacyPage) showModalPasswordStartAccountMixer() {
...
 
 func (pg *PrivacyPage) listenForMixerNotifications() {
+	if pg.AccountMixerNotificationListener != nil {
+		return
+	}
+
+	pg.AccountMixerNotificationListener = listeners.NewAccountMixerNotificationListener()
+	err := pg.WL.MultiWallet.AddAccountMixerNotificationListener(pg, PrivacyPageID)
+	if err != nil {
+		log.Errorf("Error adding account Mixer notification listener: %+v", err)
```
itswisdomagain:
```suggestion
		log.Errorf("Error adding account mixer notification listener: %v", err)
```

ui/page/governance/proposal_details_page.go
```diff
@@ -129,26 +131,34 @@ func (pg *ProposalDetails) HandleUserInteractions() {
...
-			case <-pg.ctx.Done():
-				return
-			}
-
-			switch n := notification.(type) {
-			case wallet.Proposal:
-				if n.ProposalStatus == wallet.Synced {
+			case notification := <-pg.ProposalNotifChan:
+				switch notification.ProposalStatus {
+				case wallet.Synced:
```
itswisdomagain: Isn't an if statement here simpler and more straightforward?
```suggestion
				if notification.ProposalStatus == wallet.Synced {
```

ui/page/overview/overview_layout.go
```diff
@@ -376,9 +393,43 @@ func (pg *AppOverviewPage) listenForSyncNotifications() {
...
+				switch n.Type {
+				case listeners.NewTransaction:
+					pg.loadTransactions()
+					pg.RefreshWindow()
+				case listeners.BlockAttached:
 					pg.RefreshWindow()
 				}
+			case n := <-pg.ProposalNotifChan:
+				switch n.ProposalStatus {
+				case wallet.Synced:
```
itswisdomagain:
```suggestion
				if n.ProposalStatus == wallet.Synced {
```

ui/page/overview/overview_layout.go
```diff
@@ -376,9 +393,43 @@ func (pg *AppOverviewPage) listenForSyncNotifications() {
...
+					pg.loadRecentProposals()
+					pg.RefreshWindow()
+				}
+
+			case n := <-pg.BlockRescanChan:
+				if n.Stage == wallet.RescanEnded {
```
itswisdomagain: Missing a `pg.rescanUpdate = &n`:
```suggestion
			case n := <-pg.BlockRescanChan:
				pg.rescanUpdate = &n
				if n.Stage == wallet.RescanEnded {
```

itswisdomagain (changes requested):
Mostly nits. Will test after this. Preferably make this changes in a new commit, so I only review that commit.

itswisdomagain (dismissed):
I've done some testing but will recommend more, extensive testing before merging.

Sirmorrison (approved):

PR #779, opened 2022-01-31
Fix wallet acount balance bug

Sirmorrison:
- [x] Fix bug with wallet page balance update until page is exited and reopened.
- [x] Fix account selector selected wallet account balance difference with modal difference.
image

### DIFF
--- a/ui/page/components/account_selector.go
+++ b/ui/page/components/account_selector.go
@@ -20,9 +20,6 @@ import (
 type AccountSelector struct {
 	*load.Load
 
-	ctx       context.Context // modal context
-	ctxCancel context.CancelFunc
-
 	multiWallet     *dcrlibwallet.MultiWallet
 	selectedAccount *dcrlibwallet.Account
 
@@ -82,16 +79,12 @@ func (as *AccountSelector) Handle() {
 			}).
 			onModalExit(func() {
 				as.selectorModal = nil
-				as.ctxCancel()
 			})
 		as.ShowModal(as.selectorModal)
 	}
 }
 
 func (as *AccountSelector) SelectFirstWalletValidAccount() error {
-	as.ctx, as.ctxCancel = context.WithCancel(context.TODO())
-	as.listenForTxNotifications()
-
 	if as.selectedAccount != nil && as.accountIsValid(as.selectedAccount) {
 		as.UpdateSelectedAccountBalance()
 		// no need to select account
@@ -206,14 +199,14 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
 	)
 }
 
-func (as *AccountSelector) listenForTxNotifications() {
+func (as *AccountSelector) ListenForTxNotifications(ctx context.Context) {
 	go func() {
 		for {
 			var notification interface{}
 
 			select {
 			case notification = <-as.Receiver.NotificationsUpdate:
-			case <-as.ctx.Done():
+			case <-ctx.Done():
 				return
 			}
 
@@ -279,11 +272,11 @@ func newAccountSelectorModal(l *load.Load, currentSelectedAccount *dcrlibwallet.
 		isCancelable:           true,
 	}
 
-	asm.modal.ShowScrollbar(true)
 	asm.walletInfoButton = l.Theme.IconButton(asm.Icons.ActionInfo)
 	asm.walletInfoButton.Size = values.MarginPadding15
 	asm.walletInfoButton.Inset = layout.UniformInset(values.MarginPadding0)
 
+	asm.modal.ShowScrollbar(true)
 	return asm
 }
 
@@ -337,6 +330,7 @@ func (asm *AccountSelectorModal) Handle() {
 			for _, account := range accounts {
 				for account.clickable.Clicked() {
 					asm.callback(account.Account)
+					asm.onExit()
 					asm.Dismiss()
 				}
 			}
--- a/ui/page/components/account_selector.go
+++ b/ui/page/components/account_selector.go
@@ -1,8 +1,8 @@
 package components
 
 import (
+	"context"
 	"errors"
-	"fmt"
 
 	"gioui.org/io/event"
 	"gioui.org/layout"
@@ -13,20 +13,20 @@ import (
 	"github.com/planetdecred/godcr/ui/decredmaterial"
 	"github.com/planetdecred/godcr/ui/load"
 	"github.com/planetdecred/godcr/ui/values"
+	"github.com/planetdecred/godcr/wallet"
 )
 
 type AccountSelector struct {
 	*load.Load
-	multiWallet *dcrlibwallet.MultiWallet
-	dialogTitle string
+	multiWallet     *dcrlibwallet.MultiWallet
+	selectedAccount *dcrlibwallet.Account
 
 	accountIsValid func(*dcrlibwallet.Account) bool
 	callback       func(*dcrlibwallet.Account)
 
 	openSelectorDialog *decredmaterial.Clickable
 
-	wallets            []*dcrlibwallet.Wallet
-	selectedAccount    *dcrlibwallet.Account
+	dialogTitle        string
 	selectedWalletName string
 	totalBalance       string
 	changed            bool
@@ -39,8 +39,6 @@ func NewAccountSelector(l *load.Load) *AccountSelector {
 		multiWallet:        l.WL.MultiWallet,
 		accountIsValid:     func(*dcrlibwallet.Account) bool { return true },
 		openSelectorDialog: l.Theme.NewClickable(true),
-
-		wallets: l.WL.SortedWalletList(),
 	}
 }
 
@@ -67,7 +65,7 @@ func (as *AccountSelector) Changed() bool {
 
 func (as *AccountSelector) Handle() {
 	for as.openSelectorDialog.Clicked() {
-		newAccountSelectorModal(as.Load, as.selectedAccount, as.wallets).
+		newAccountSelectorModal(as.Load, as.selectedAccount).
 			title(as.dialogTitle).
 			accountValidator(as.accountIsValid).
 			accountSelected(func(account *dcrlibwallet.Account) {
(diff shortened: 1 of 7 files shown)


ui/page/components/account_selector.go
```diff
@@ -234,16 +241,17 @@ func newAccountSelectorModal(l *load.Load, currentSelectedAccount *dcrlibwallet.
...
-	walletAccounts := make(map[int][]*selectorAccount)
+	asm.ctx, asm.ctxCancel = context.WithCancel(context.TODO())
+	asm.setupAccountBalance()
+	asm.listenForTxNotifications()
+}
 
-	// TODO use a sorted wallet list
-	for _, wal := range asm.wallets {
-		// filter all accounts
+func (asm *AccountSelectorModal) setupAccountBalance() {
```
itswisdomagain: `setupWalletAccounts` or just `setupAccounts`.

ui/page/components/account_selector.go
```diff
@@ -234,16 +241,17 @@ func newAccountSelectorModal(l *load.Load, currentSelectedAccount *dcrlibwallet.
...
+	walletAccounts := make(map[int][]*selectorAccount)
+	for _, wal := range asm.WL.SortedWalletList() {
 		accountsResult, err := wal.GetAccountsRaw()
 		if err != nil {
-			fmt.Println("Error getting accounts:", err)
-			return
+			continue
```
itswisdomagain: Why completely ignoring this error? At least a log is necessary.

ui/page/components/account_selector.go
```diff
@@ -256,15 +264,8 @@ func (asm *AccountSelectorModal) OnResume() {
 				})
 			}
 		}
-
-		// add wallet only if it has valid accounts
-		if len(walletAccounts[wal.ID]) > 0 {
-			wallets = append(wallets, wal)
-		}
+		asm.accounts = walletAccounts
```
itswisdomagain: Assign outside the for loop.

ui/page/components/account_selector.go
```diff
@@ -491,3 +489,35 @@ func (asm *AccountSelectorModal) walletInfoPopup(gtx layout.Context) layout.Dime
...
+
+			select {
+			case notification = <-asm.Receiver.NotificationsUpdate:
+			case <-asm.ctx.Done():
+				return
+			}
+
+			switch n := notification.(type) {
+			case wallet.NewTransaction:
+				// refresh wallets when new transaction is received
```
itswisdomagain:
```suggestion
				// refresh accounts list when new transaction is received
				// to update their individual balances
```

ui/page/components/account_selector.go
```diff
@@ -491,3 +489,35 @@ func (asm *AccountSelectorModal) walletInfoPopup(gtx layout.Context) layout.Dime
 		})
 	})
 }
+
+func (asm *AccountSelectorModal) listenForTxNotifications() {
```
itswisdomagain: I wonder if this should be on `accountSelector` which may then update the modal if it is open. Does the selected balance update when a new tx or block comes in?

ui/page/receive_page.go
```diff
@@ -397,11 +404,41 @@ func (pg *ReceivePage) handleCopyEvent(gtx layout.Context) {
 	}
 }
 
+func (pg *ReceivePage) listenForTxNotifications() {
```
itswisdomagain: Ah, I see. Related to my previous comment. The `pg.selector` should listen for updates and update the selected balance rather than each page that uses it having to do that and call `pg.selector.UpdateSelectedAccountBalance()`.

ui/page/components/account_selector.go
```diff
@@ -67,7 +65,7 @@ func (as *AccountSelector) Changed() bool {
 
 func (as *AccountSelector) Handle() {
 	for as.openSelectorDialog.Clicked() {
-		newAccountSelectorModal(as.Load, as.selectedAccount, as.wallets).
+		newAccountSelectorModal(as.Load, as.selectedAccount).
```
itswisdomagain: Suggestion:
- Assign this modal to `as.selectorModal` when `openSelectorDialog` is clicked.
- When the modal is closed, set `as.selectorModal = nil`.
- When the proposed `as.listenForTxNotifications` sees a new tx or block, it calls `as. UpdateSelectedAccountBalance()` and `if as.selectorModal != nil`, then it also calls `as.selectorModal.setupAccountBalance()/setupAccounts` and refreshes the window.

One listener, does everything.

itswisdomagain (changes requested):
Initial review.

ui/page/components/account_selector.go
```diff
@@ -297,6 +343,7 @@ func (asm *AccountSelectorModal) Handle() {
 	}
 
 	if asm.modal.BackdropClicked(asm.isCancelable) {
+		asm.onExit()
```
itswisdomagain, line 341: `onExit` should also be called when the modal closes because an account was selected, yeah?

ui/page/components/account_selector.go
```diff
@@ -76,17 +79,26 @@ func (as *AccountSelector) Handle() {
 				}
 				as.SetSelectedAccount(account)
 				as.callback(account)
-			}).Show()
+			}).
+			onModalExit(func() {
+				as.selectorModal = nil
+				as.ctxCancel()
```
itswisdomagain: I don't know about stopping the tx+block notification listener when this modal is closed. It would still be needed to update the selected account when necessary. Also, if the modal were reopened, the listener isn't restarted. The listener should ideally only be canceled when the page holding the AccountSelector widget is closed.

Here's a thought: the widget should keep listening for potential balance changes as long as the page that is displaying it is open. To ensure that the listener is stopped when the page is no longer being displayed, add a `AccountSelector.WatchForBalanceChanges(ctx)` method takes a `ctx` parameter that will be supplied by the page that wants to display the selector widget. The method should start the listener. The page should cancel that ctx when the page is navigated from, which ensures that the listener is stopped.

ui/page/wallets/wallet_page.go
```diff
@@ -1079,6 +1085,61 @@ func (pg *WalletPage) deleteBadWallet(badWalletID int) {
...
+				if n.Stage == wallet.BlockAttached && pg.WL.MultiWallet.IsSynced() {
+					pg.UpdateAccountBalance()
+					pg.RefreshWindow()
+				}
+			}
+		}
+	}()
+}
+
+func (pg *WalletPage) UpdateAccountBalance() {
```
itswisdomagain: Private method if not called outside this package. Doesn't look like it is called outside.

ui/page/wallets/wallet_page.go
```diff
@@ -1079,6 +1085,61 @@ func (pg *WalletPage) deleteBadWallet(badWalletID int) {
...
+		for _, acc := range accountsResult.Acc {
+			totalBalance += acc.TotalBalance
+		}
+
+		for _, item := range listItems {
+			if item.wal.ID == wal.ID {
+				item.totalBalance = dcrutil.Amount(totalBalance).String()
+				item.accounts = accountsResult.Acc
+			}
+		}
```
itswisdomagain: If this updates the item in `pg.listItems`, then you want to keep `pg.listLock` locked till this method completes. If it doesn't, then this looks pointless because `item` isn't used after here.

ui/page/wallets/wallet_page.go
```diff
@@ -1079,6 +1085,61 @@ func (pg *WalletPage) deleteBadWallet(badWalletID int) {
...
+			continue
+		}
+
+		var totalBalance int64
+		for _, acc := range accountsResult.Acc {
+			totalBalance += acc.TotalBalance
+		}
+
+		for _, item := range listItems {
+			if item.wal.ID == wal.ID {
```
itswisdomagain: Instead of ranging through all wallets and doing this check, while not only range through `listItems` and get the wallet using `mw.WalletWithID`?

ui/page/wallets/wallet_page.go
```diff
@@ -1004,7 +1010,7 @@ func (pg *WalletPage) HandleUserInteractions() {
 											pg.Toast.Notify("Account created")
 											tim.Dismiss()
 										}
-										pg.loadWalletAndAccounts()
+										pg.UpdateAccountBalance()
```
itswisdomagain: Does `UpdateAccountBalance` also update the accounts list? Because a new account has just been added.

itswisdomagain (changes requested):

ui/page/wallets/wallet_page.go
Sirmorrison: yes it does. This also prevents the entire page from refreshing which in turn close all opened wallet modals when a new account is created.

ui/page/components/account_selector.go
```diff
@@ -186,47 +199,79 @@ func (as *AccountSelector) Layout(gtx layout.Context) layout.Dimensions {
...
 		modal:        *l.Theme.ModalFloatTitle(),
 		walletsList:  layout.List{Axis: layout.Vertical},
 		accountsList: layout.List{Axis: layout.Vertical},
 
 		currentSelectedAccount: currentSelectedAccount,
-		wallets:                wallets,
 		isCancelable:           true,
 	}
 
-	asm.modal.ShowScrollbar(true)
```
itswisdomagain, line 229: No longer necessary?

ui/page/wallets/wallet_page.go
```diff
@@ -1079,6 +1085,60 @@ func (pg *WalletPage) deleteBadWallet(badWalletID int) {
...
+
+			var totalBalance int64
+			for _, acc := range accountsResult.Acc {
+				totalBalance += acc.TotalBalance
+			}
+
+			pg.listLock.Lock()
+			item.totalBalance = dcrutil.Amount(totalBalance).String()
+			item.accounts = accountsResult.Acc
+			pg.listLock.Unlock()
```
itswisdomagain: Does this method take long to complete? Why not `pg.listLock.Lock()` and `defer pg.listLock.Unlock()` at the top of the method, then range through `pg.listItems` and make modifications as necessary.

itswisdomagain (changes requested):
Yet to test but looking much better. Just 2 Qs.

itswisdomagain (approved):
Tests well.

dreacot (approved):

PR #782, opened 2022-02-03
Disable actions on modals when a process is running.

### LINKED ISSUE #774: Remove text from dex intro screen
Remove the following line of text from the DEX intro screen. Please leave the whitespace (blank line):

Screen Shot 2022-01-25 at 7 18 41 AM

Sirmorrison:
* Currently, when the send/confirm button is selected, the editors and other actionable widgets on the modals are still active. 
This PR aims to disable the modals while the action is in progress.

also Fix #774

### DIFF
--- a/ui/decredmaterial/editor.go
+++ b/ui/decredmaterial/editor.go
@@ -48,9 +48,6 @@ type Editor struct {
 	// isEditorButtonClickable passes a clickable icon button if true and regular icon if false
 	isEditorButtonClickable bool
 
-	// disabled sets the state of the editor to either enabled state or disbled state
-	IsDisabled bool
-
 	requiredErrorText string
 
 	editorIcon       *Icon
@@ -234,14 +231,7 @@ func (e Editor) editor(gtx layout.Context) layout.Dimensions {
 						Top:    e.m5,
 						Bottom: e.m5,
 					}
-					return inset.Layout(gtx, func(gtx C) D {
-						editorGtx := gtx
-						if e.IsDisabled {
-							editorGtx = gtx.Disabled()
-						}
-
-						return e.EditorStyle.Layout(editorGtx)
-					})
+					return inset.Layout(gtx, e.EditorStyle.Layout)
 				}),
 			)
 		}),
--- a/ui/decredmaterial/editor.go
+++ b/ui/decredmaterial/editor.go
@@ -48,6 +48,9 @@ type Editor struct {
 	// isEditorButtonClickable passes a clickable icon button if true and regular icon if false
 	isEditorButtonClickable bool
 
+	// disabled sets the state of the editor to either enabled state or disbled state
+	Disabled bool
+
 	requiredErrorText string
 
 	editorIcon       *Icon
@@ -232,7 +235,12 @@ func (e Editor) editor(gtx layout.Context) layout.Dimensions {
 						Bottom: e.m5,
 					}
 					return inset.Layout(gtx, func(gtx C) D {
-						return e.EditorStyle.Layout(gtx)
+						editorGtx := gtx
+						if e.Disabled {
+							editorGtx = gtx.Disabled()
+						}
+
+						return e.EditorStyle.Layout(editorGtx)
 					})
 				}),
 			)
--- a/ui/decredmaterial/editor.go
+++ b/ui/decredmaterial/editor.go
@@ -49,7 +49,7 @@ type Editor struct {
 	isEditorButtonClickable bool
 
 	// disabled sets the state of the editor to either enabled state or disbled state
-	Disabled bool
+	IsDisabled bool
 
 	requiredErrorText string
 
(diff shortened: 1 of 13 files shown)


ui/decredmaterial/editor.go
```diff
@@ -48,6 +48,9 @@ type Editor struct {
 	// isEditorButtonClickable passes a clickable icon button if true and regular icon if false
 	isEditorButtonClickable bool
 
+	// disabled sets the state of the editor to either enabled state or disbled state
+	IsDisabled bool
```
itswisdomagain: This is currently never set to true, don't think we need it.

ui/modal/create_password_modal.go
```diff
@@ -230,6 +229,8 @@ func (cm *CreatePasswordModal) Handle() {
 		}
 	}
 
+	cm.modal.SetDisabled(cm.isLoading)
```
itswisdomagain: Move this to inside the `cm.SetLoading` method so whenever `cm.isLoading` is modified, this is also done. Also want to point out that `cm.isLoading` is never set to false once it's set to true - is this right?

itswisdomagain: I also see a confusing `cm.isEnabled` property. Looking at the code, looks like that should be a `cm.canSubmit` property? It's also never used outside of the `Handle` method, so I'm sure it can be a function-scope variable that is given the appropriate value at the beginning of the `Handle` method.

ui/modal/create_watch_only_modal.go
```diff
@@ -107,6 +107,8 @@ func (cm *CreateWatchOnlyModal) WatchOnlyCreated(callback func(walletName, extPu
 }
 
 func (cm *CreateWatchOnlyModal) Handle() {
+	cm.modal.SetDisabled(cm.isLoading)
```
itswisdomagain: `cm.isLoading` might be false at this point and the modal will be enabled but could change to true later in this method in line 139 and the modal will not be disabled. Also worth noting that the `cm.SetLoading(false)` method may be called outside this file and the modal will need to be enabled if previously disabled (as it should be disabled when cm.isLoading became true).

In general, the modal's enabled state should be changed when some action makes it necessary to change the modal state such as when a button is clicked and the form is being submitted or when the form submission completes and the modal can be renabled. In the previous 2 files I've reviewed, there's a `SetLoading` method that seems to be designed to handle moments when the modal is about to be busy or stops being busy. That would be the ideal place to do `modal.SetDisabled`.

itswisdomagain (changes requested):

ui/modal/create_password_modal.go
Sirmorrison: > Move this to inside the `cm.SetLoading` method so whenever `cm.isLoading` is modified, this is also done. Also want to point out that `cm.isLoading` is never set to false once it's set to true - is this right?

This is set from the page calling create_password_modal using the `SetLoading()`.

ui/page/dexclient/add_dex_modal.go
```diff
@@ -74,6 +74,8 @@ func (md *addDexModal) OnResume() {
 }
 
 func (md *addDexModal) Handle() {
+	md.modal.SetDisabled(md.isSending)
```
itswisdomagain: Where is the value of `md.isSending` modified? That's where `md.modal.SetDisabled` should be called too.

ui/page/dexclient/create_wallet_modal.go
```diff
@@ -94,6 +94,8 @@ func (md *createWalletModal) OnResume() {
 }
 
 func (md *createWalletModal) Handle() {
+	md.modal.SetDisabled(md.isSending)
```
itswisdomagain: Where is the value of `md.isSending` modified? That's where `md.modal.SetDisabled` should be called too.

ui/page/send/send_confirm_modal.go
```diff
@@ -115,6 +115,8 @@ func (scm *sendConfirmModal) broadcastTransaction() {
 }
 
 func (scm *sendConfirmModal) Handle() {
+	scm.modal.SetDisabled(scm.isSending)
```
itswisdomagain: Where is the value of `scm.isSending` modified? That's where `scm.modal.SetDisabled` should be called too.

ui/page/staking/purchase_modal.go
```diff
@@ -339,6 +339,8 @@ func (tp *stakingModal) calculateTotals() {
 }
 
 func (tp *stakingModal) Handle() {
+	tp.modal.SetDisabled(tp.isLoading)
```
itswisdomagain: Where is the value of `tp.isLoading` modified? That's where `tp.modal.SetDisabled` should be called too.

itswisdomagain (changes requested):
Looking better, just a few more changes.

dreacot (approved):

PR #784, opened 2022-02-08
Fix USD balance display bug

### LINKED ISSUE #783: USD balance not populating on some occasions
The USD equivalent of the wallet balance does not show up on the screen , see image below:

Screenshot from 2022-02-03 22-15-00

devchoplife:
This PR closes #783.
This PR ensures that the USD balance container is not displayed when the USD balance has not been calculated. 
The USD container along with the balance is only shown if the USD balance has been calculated.

Also, it ensures that the USD exchange rate is fetched on every block and transaction  added, before updating the USD balance.

### DIFF
--- a/ui/page/listeners.go
+++ b/ui/page/listeners.go
@@ -16,6 +16,7 @@ import (
 
 func (mp *MainPage) OnTransaction(transaction string) {
 	transactionNotification := mp.WL.Wallet.ReadBoolConfigValueForKey(load.TransactionNotificationConfigKey)
+	load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
 	mp.UpdateBalance()
 
 	var tx dcrlibwallet.Transaction
@@ -41,6 +42,7 @@ func (mp *MainPage) OnBlockAttached(walletID int, blockHeight int32) {
 		}
 	}
 
+	load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
 	mp.UpdateBalance()
 	mp.UpdateNotification(wallet.SyncStatusUpdate{
 		Stage: wallet.BlockAttached,
@@ -48,6 +50,7 @@ func (mp *MainPage) OnBlockAttached(walletID int, blockHeight int32) {
 }
 
 func (mp *MainPage) OnTransactionConfirmed(walletID int, hash string, blockHeight int32) {
+	load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
 	mp.UpdateBalance()
 }
 
@@ -148,6 +151,7 @@ func (mp *MainPage) OnHeadersRescanProgress(headersRescanProgress *dcrlibwallet.
 	})
 }
 func (mp *MainPage) OnSyncCompleted() {
+	load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
 	mp.UpdateBalance()
 	mp.UpdateNotification(wallet.SyncStatusUpdate{
 		Stage: wallet.SyncCompleted,
--- a/ui/page/listeners.go
+++ b/ui/page/listeners.go
@@ -16,8 +16,7 @@ import (
 
 func (mp *MainPage) OnTransaction(transaction string) {
 	transactionNotification := mp.WL.Wallet.ReadBoolConfigValueForKey(load.TransactionNotificationConfigKey)
-	load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
-	mp.UpdateBalance()
+	mp.UpdateBalance(true)
 
 	var tx dcrlibwallet.Transaction
 	err := json.Unmarshal([]byte(transaction), &tx)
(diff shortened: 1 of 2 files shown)


ui/page/main_page.go
```diff
@@ -553,9 +552,10 @@ func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
 }
 
 func (mp *MainPage) LayoutUSDBalance(gtx layout.Context) layout.Dimensions {
+	mp.UpdateBalance()
 	return layout.Flex{Axis: layout.Vertical}.Layout(gtx,
 		layout.Rigid(func(gtx C) D {
-			if mp.usdExchangeSet && mp.dcrUsdtBittrex.LastTradeRate != "" {
+			if mp.usdExchangeSet && mp.dcrUsdtBittrex.LastTradeRate != "" && len(mp.totalBalanceUSD) > 0 {
```
Sirmorrison, line 559: I really dont think this is the solution to this issue. What you seem to have implemented is a temporary fix to hide USD layout if no total USD Bal.

However, total USD bal is supposed to have data since there exist fetched USD value as can be seen from the send page currency conversion.
image

Sirmorrison (changes requested):

Sirmorrison:
You need to find a way to refresh the view to display the USD balance the moment it is available.
image

currently, there is USD value but not USD total bal. It is only displayed after the app is restarted.

devchoplife:
> You need to find a way to refresh the view to display the USD balance the moment it is available. image
> 
> currently, there is USD value but not USD total bal. It is only displayed after the app is restarted.

Address is needed to calculate the transaction fee and address is not set. Testing from my end, It works fine when address is set. Please set address and test again to check if we have the same behavior.

ui/page/main_page.go
```diff
@@ -229,6 +226,7 @@ func (mp *MainPage) setLanguageSetting() {
 }
 
 func (mp *MainPage) UpdateBalance() {
+	go load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
```
Sirmorrison: the `UpdateBalance` looks to be called multiple times, which will start up multiple go routines... not good. 
You can simply add a condition to check if the usd values is available. if it is not, the fetch it and refresh window. if it is, skip fetching it.

Issue i see here is that if the USD exchange val is updated, the total bal will still be same as `UpdateBalance` is called when there is a new block and tx from the listener.go file.

You would need a way to track and ensure that the USD exchange value is updated with each new block and tx.

Sirmorrison (changes requested):

ui/page/main_page.go
```diff
@@ -553,9 +553,10 @@ func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
 }
 
 func (mp *MainPage) LayoutUSDBalance(gtx layout.Context) layout.Dimensions {
+	mp.UpdateBalance()
```
Sirmorrison: Looking carefully, i dont think this should be in the `LayoutUSDBalance` reason is this is meant for just the USD balance display.

What happens if it is moved to the `Layout`?

ui/page/main_page.go
```diff
@@ -229,6 +226,9 @@ func (mp *MainPage) setLanguageSetting() {
 }
 
 func (mp *MainPage) UpdateBalance() {
+	if len(mp.totalBalanceUSD) == 0 {
+		load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
+	}
```
```suggestion
func (mp *MainPage) UpdateBalance(fetchExch bool) {
	if len(mp.totalBalanceUSD) == 0 || fetchExch {
		load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
	}
```

ui/page/listeners.go
```diff
@@ -16,6 +16,7 @@ import (
 
 func (mp *MainPage) OnTransaction(transaction string) {
 	transactionNotification := mp.WL.Wallet.ReadBoolConfigValueForKey(load.TransactionNotificationConfigKey)
+	load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
 	mp.UpdateBalance()
```
Sirmorrison:
```suggestion
	mp.UpdateBalance(true)
```

ui/page/listeners.go
```diff
@@ -41,13 +42,15 @@ func (mp *MainPage) OnBlockAttached(walletID int, blockHeight int32) {
 		}
 	}
 
+	load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
 	mp.UpdateBalance()
```
Sirmorrison:
```suggestion
	mp.UpdateBalance(true)
```

ui/page/listeners.go
```diff
@@ -41,13 +42,15 @@ func (mp *MainPage) OnBlockAttached(walletID int, blockHeight int32) {
...
 	mp.UpdateNotification(wallet.SyncStatusUpdate{
 		Stage: wallet.BlockAttached,
 	})
 }
 
 func (mp *MainPage) OnTransactionConfirmed(walletID int, hash string, blockHeight int32) {
+	load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
 	mp.UpdateBalance()
```
Sirmorrison:
```suggestion
	mp.UpdateBalance(true)
```

ui/page/listeners.go
```diff
@@ -148,6 +151,7 @@ func (mp *MainPage) OnHeadersRescanProgress(headersRescanProgress *dcrlibwallet.
 	})
 }
 func (mp *MainPage) OnSyncCompleted() {
+	load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
 	mp.UpdateBalance()
```
Sirmorrison:
```suggestion
	mp.UpdateBalance(true)
```

ui/page/main_page.go
```diff
@@ -497,6 +497,7 @@ func (mp *MainPage) popToFragment(pageID string) {
 // to be eventually drawn on screen.
 // Part of the load.Page interface.
 func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
+	mp.UpdateBalance()
```
```suggestion
	mp.UpdateBalance(false)
```

Sirmorrison (changes requested):

Sirmorrison (approved):

PR #788, opened 2022-02-11
 Init multi wallet in main.go from window.go

### LINKED ISSUE #786: GoDCR crashes after 10-15 minutes of inactivity
GoDcr crashes after a few minutes of inactivity, giving the error in the attached screenshot

Issue

Issue also occurs 5-10 minutes after syncing has commenced.

issue 2

AdimekweEbuka:
Fix #786

### DIFF
--- a/main.go
+++ b/main.go
@@ -70,6 +70,12 @@ func main() {
 		return
 	}
 
+	newError := wal.InitMultiWallet()
+	if newError != nil {
+		log.Errorf("init multiwallet error: %v", newError)
+		return
+	}
+
 	go func() {
 		win.HandleEvents() // blocks until the app window is closed
 		wal.Shutdown()
--- a/main.go
+++ b/main.go
@@ -70,9 +70,9 @@ func main() {
 		return
 	}
 
-	newError := wal.InitMultiWallet()
-	if newError != nil {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -70,6 +70,12 @@ func main() {
 		return
 	}
 
+	newError := wal.InitMultiWallet()
```
dreacot:
```suggestion
	err = wal.InitMultiWallet()
```

devchoplife (commented):
@AdimekweEbuka Please add details about what you did to resolve the issue

AdimekweEbuka:
> @AdimekweEbuka Please add details about what you did to resolve the issue

`InitMultiWallet()` function was being called multiple times because of its location in the StageRunning event handler when it should be only called once. It was then moved to the main.go to ensure that it's only being called once.

AdimekweEbuka:
> @AdimekweEbuka Please add details about what you did to resolve the issue
Capture2

This was from Wisdom in the Godcr room, when the issue was brought up. And moving it to the main function stopped the crash. 

Please assist in testing to see if it has also been rectified at your end.

Sirmorrison (approved):

PR #800, opened 2022-03-01
Show ticket associated with Voted/Revocation transaction

### LINKED ISSUE #792: View associated tickets button does not open any ticket
On the transaction details page:  if the transaction is a `Vote` or `Revocation` transaction, the `View associated ticket` button is shown. However, when the button is clicked, it does not open the ticket associated with the transaction.

See the video below for more details:
https://streamable.com/6zx9ue

devchoplife:
Closes #792 
This PR fixes the bug that hinders the `View associated ticket` button from showing the details of the ticket associated with a Vote/Revocation transaction.

### DIFF
--- a/ui/page/transaction/transaction_details_page.go
+++ b/ui/page/transaction/transaction_details_page.go
@@ -54,6 +54,7 @@ type TxDetailsPage struct {
 	transaction   *dcrlibwallet.Transaction
 	ticketSpender *dcrlibwallet.Transaction // vote or revoke ticket
 	ticketSpent   *dcrlibwallet.Transaction // ticket spent in a vote or revoke
+	txBackStack   *dcrlibwallet.Transaction // track original transaction
 	wallet        *dcrlibwallet.Wallet
 
 	txSourceAccount      string
@@ -98,10 +99,14 @@ func NewTransactionDetailsPage(l *load.Load, transaction *dcrlibwallet.Transacti
 	pg.dot = decredmaterial.NewIcon(l.Icons.ImageBrightness1)
 	pg.dot.Color = l.Theme.Color.Gray1
 
+	return pg
+}
+
+func (pg *TxDetailsPage) getTXSourceAccountAndDirection() {
 	// find source account
-	if transaction.Direction == dcrlibwallet.TxDirectionSent ||
-		transaction.Direction == dcrlibwallet.TxDirectionTransferred {
-		for _, input := range transaction.Inputs {
+	if pg.transaction.Direction == dcrlibwallet.TxDirectionSent ||
+		pg.transaction.Direction == dcrlibwallet.TxDirectionTransferred {
+		for _, input := range pg.transaction.Inputs {
 			if input.AccountNumber != -1 {
 				accountName, err := pg.wallet.AccountName(input.AccountNumber)
 				if err != nil {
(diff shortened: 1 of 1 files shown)


devchoplife:
From my investigation, the `View associated tickets` button does not navigate to the ticket details page because it is trying to create another instance of the transaction details page (because the current page is also an instance of the transaction details page). By the design of the page functionality, the two instances share the same page ID. The condition below (within the `ChangeFragment` function which handles page navigation) checks if the current page and the new page have the same IDs, then exits the function if they do:
https://github.com/planetdecred/godcr/blob/733b3411857f604c49d23d501ac9e680906c4dd9/ui/page/main_page.go#L431

The two commits in this PR present two approaches to fixing this issue:

- The first ensures that the condition evaluates to false if the current page ID points to the transaction details page 
- The second adds a boolean argument to the `ChangeFragment` function which when set to true, evaluates the condition to false. I believe this is a better approach because it narrows the scope to a specific navigation instance.

ui/page/transaction/transaction_details_page.go
```diff
@@ -721,7 +721,7 @@ func (pg *TxDetailsPage) HandleUserInteractions() {
 
 	for pg.associatedTicketClickable.Clicked() {
 		if pg.ticketSpent != nil {
-			pg.ChangeFragment(NewTransactionDetailsPage(pg.Load, pg.ticketSpent))
+			pg.ChangeFragment(NewTransactionDetailsPage(pg.Load, pg.ticketSpent), true)
```
Sirmorrison:
```suggestion
			if pg.ticketSpent != nil {
			pg.txBackStack = pg.transaction
			pg.transaction = pg.ticketSpent

			pg.getTXSourceAccountAndDirection()
			pg.txnWidgets = initTxnWidgets(pg.Load, pg.transaction)
			pg.RefreshWindow()
		}
```

Sirmorrison (changes requested):
First of all i have a challenge with the approach used for viewing the related ticket. You call the transaction details page from inside the transactions details page this is bad development and not scalable.

From my review, the `ticketSpent` is a transaction on its own. why not just update the page to view the `ticketSpent` details than recalling the transaction details page.

First approach will be to clean up the entire code to accommodate this.
* Add this `txBackStack *dcrlibwallet.Transaction // track original tx` to the `TxDetailsPage` struct.

* Move these codes from line 102 to 122 to a new function

```
func (pg *TxDetailsPage) getTXSourceAccountAndDirection() {
	// find source account
	if pg.transaction.Direction == dcrlibwallet.TxDirectionSent ||
		pg.transaction.Direction == dcrlibwallet.TxDirectionTransferred {
		for _, input := range pg.transaction.Inputs {
			if input.AccountNumber != -1 {
				accountName, err := pg.wallet.AccountName(input.AccountNumber)
				if err != nil {
					// log.Error(err)
				} else {
					pg.txSourceAccount = accountName
				}
			}
		}
	}
	//	find destination address
	if pg.transaction.Direction == dcrlibwallet.TxDirectionSent {
		for _, output := range pg.transaction.Outputs {
			if output.AccountNumber == -1 {
				pg.txDestinationAddress = output.Address
			}
		}
	}
}
```
*  move `pg.txnWidgets = initTxnWidgets(pg.Load, pg.transaction)` to the `OnNavigatedTo`.
`OnNavigatedTo` should look like this.

```
// OnNavigatedTo is called when the page is about to be displayed and
// may be used to initialize page features that are only relevant when
// the page is displayed.
func (pg *TxDetailsPage) OnNavigatedTo() {
	if pg.transaction.TicketSpentHash != "" {
		pg.ticketSpent, _ = pg.wallet.GetTransactionRaw(pg.transaction.TicketSpentHash)
	}

	if ok, _ := pg.wallet.TicketHasVotedOrRevoked(pg.transaction.Hash); ok {
		pg.ticketSpender, _ = pg.wallet.TicketSpender(pg.transaction.Hash)
	}

	pg.getTXSourceAccountAndDirection()
	pg.txnWidgets = initTxnWidgets(pg.Load, pg.transaction)
}
```

Finally, in the `Layout()` add this to `Back:`
```
InfoButton: pg.infoButton,
Back: func() {
if pg.txBackStack == nil {
	pg.PopFragment()
	return
}
pg.transaction = pg.txBackStack
pg.getTXSourceAccountAndDirection()
pg.txnWidgets = initTxnWidgets(pg.Load, pg.transaction)
pg.txBackStack = nil
pg.RefreshWindow()
},
```

this issue should be fixed...

Sirmorrison (approved):

PR #802, opened 2022-03-03
Fix app crash bug when wallet is deleted

### LINKED ISSUE #793: App crashes on wallet deletion
When I attempt to remove a wallet from my device, GoDCR crashes with the error log below returned in the terminal:
This issue has been reproduced on Ubuntu, Fedora and Windows.

`goroutine 10 [running]:
decred.org/dcrwallet/v2/wallet.(*Wallet).AccountBalance(0xc000453408, {0x1bd83b8, 0xc0083e6140}, 0x18e2dd80, 0xc0)
	/home/devchoplife/go/pkg/mod/decred.org/dcrwallet/v2@v2.0.0-rc3/wallet/wallet.go:1907 +0xc9
github.com/planetdecred/dcrlibwallet.(*Wallet).GetAccountBalance(0xc00025a3c0, 0x1bd83b8)
	/home/devchoplife/go/pkg/mod/github.com/planetdecred/dcrlibwallet@v1.6.2-0.20220203081357-951ce632e9d2/accounts.go:108 +0xe6
github.com/planetdecred/dcrlibwallet.(*Wallet).GetAccountsRaw(0xc00025a3c0)
	/home/devchoplife/go/pkg/mod/github.com/planetdecred/dcrlibwallet@v1.6.2-0.20220203081357-951ce632e9d2/accounts.go:41 +0x1b4
github.com/planetdecred/godcr/ui/page.(*MainPage).CalculateTotalWalletsBalance(0xc0000c8540)
	/home/devchoplife/GitRepos/godcr/ui/page/main_page.go:257 +0x7b
github.com/planetdecred/godcr/ui/page.(*MainPage).UpdateBalance(0xc000188a80, 0x20)
	/home/devchoplife/GitRepos/godcr/ui/page/main_page.go:235 +0xd5
github.com/planetdecred/godcr/ui/page.(*MainPage).Layout(0xc000188a80, {{{0x0, 0x0}, {0x3ff, 0x344}}, {0x3f800000, 0x3f800000}, {0x1bae680, 0xc00003e518}, {0xc07d98dca681f1fa, ...}, ...})
	/home/devchoplife/GitRepos/godcr/ui/page/main_page.go:500 +0x38
gioui.org/layout.Stack.Layout({0x30}, {{{0x3ff, 0x344}, {0x3ff, 0x344}}, {0x3f800000, 0x3f800000}, {0x1bae680, 0xc00003e518}, {0xc07d98dca681f1fa, ...}, ...}, ...)
	/home/devchoplife/go/pkg/mod/gioui.org@v0.0.0-20211011183043-05f0f5c20f45/layout/stack.go:60 +0x895
github.com/planetdecred/godcr/ui.(*Window).drawWindowUI(0xc0003c5cc0, {{{0x3ff, 0x344}, {0x3ff, 0x344}}, {0x3f800000, 0x3f800000}, {0x1bae680, 0xc00003e518}, {0xc07d98dca681f1fa, ...}, ...})
	/home/devchoplife/GitRepos/godcr/ui/window.go:254 +0x3d9
github.com/planetdecred/godcr/ui.(*Window).displayWindow(0xc0003c5cc0, {{0xc07d98dca681f1fa, 0xb9fcf4773a, 0x2825bc0}, {0x3f800000, 0x3f800000}, {0x3ff, 0x344}, {{0x0, 0x0}, ...}, ...})
	/home/devchoplife/GitRepos/godcr/ui/window.go:217 +0x16c
github.com/planetdecred/godcr/ui.(*Window).HandleEvents(0xc0003c5cc0)
	/home/devchoplife/GitRepos/godcr/ui/window.go:180 +0x10c
main.main.func2()
	/home/devchoplife/GitRepos/godcr/main.go:80 +0x26
created by main.main
	/home/devchoplife/GitRepos/godcr/main.go:79 +0x4d8`

Below is the link to a video that shows how to reproduce the bug:
https://streamable.com/srzm0u

Sirmorrison:
Fix #793
- add usd balance refresh/retry capability
- clean up code

### DIFF
--- a/ui/load/load.go
+++ b/ui/load/load.go
@@ -83,6 +83,8 @@ type Load struct {
 	SubscribeKeyEvent   func(eventChan chan *key.Event, pageID string) // Widgets call this function to recieve key events.
 	UnsubscribeKeyEvent func(pageID string) error
 	ReloadApp           func()
+
+	IsCurrencySettingUpdated bool
 }
 
 func (l *Load) RefreshTheme() {
--- a/ui/load/load.go
+++ b/ui/load/load.go
@@ -83,8 +83,6 @@ type Load struct {
 	SubscribeKeyEvent   func(eventChan chan *key.Event, pageID string) // Widgets call this function to recieve key events.
 	UnsubscribeKeyEvent func(pageID string) error
 	ReloadApp           func()
-
-	IsCurrencySettingUpdated bool
 }
 
 func (l *Load) RefreshTheme() {
--- a/ui/page/components/utils.go
+++ b/ui/page/components/utils.go
@@ -18,7 +18,7 @@ func ContextDone(ctx context.Context) bool {
 }
 
 // RetryFunc implements retry policy for processes that needs to be executed
-// the after initial failure.
+// after initial failure.
 func RetryFunc(retryAttempts int, sleepDur time.Duration, funcDesc string, errFunc func() error) (int, error) {
 	var err error
 	for i := 0; i < retryAttempts; i++ {
--- a/ui/page/components/utils.go
+++ b/ui/page/components/utils.go
@@ -3,7 +3,6 @@ package components
 import (
 	"context"
 	"fmt"
-	"log"
 	"time"
 )
 
@@ -18,14 +17,16 @@ func ContextDone(ctx context.Context) bool {
 	}
 }
 
-// RetryFunc implement retry policy for processes that needs to be executed
-// in the background after initial failure.
-func RetryFunc(retryAttempts int, sleepDur time.Duration, errFunc func() error) (int, error) {
+// RetryFunc implements retry policy for processes that needs to be executed
+// the after initial failure.
+func RetryFunc(retryAttempts int, sleepDur time.Duration, funcDesc string, errFunc func() error) (int, error) {
 	var err error
 	for i := 0; i < retryAttempts; i++ {
 		if i > 0 {
-			sleepDur *= 2
-			log.Printf("waiting %v to retry after error: %v", sleepDur, err)
+			if i > 1 {
+				sleepDur *= 2
+			}
+			fmt.Printf("waiting %s to retry function %s after error: %v\n", sleepDur, funcDesc, err)
 			time.Sleep(sleepDur)
 		}
 		err = errFunc()
@@ -34,5 +35,5 @@ func RetryFunc(retryAttempts int, sleepDur time.Duration, errFunc func() error) 
 		}
 	}
 
-	return retryAttempts, fmt.Errorf("after %d attempts, last error: %s", retryAttempts, err)
+	return retryAttempts, fmt.Errorf("last error: %s", err)
 }
--- a/ui/page/components/utils.go
+++ b/ui/page/components/utils.go
@@ -7,8 +7,6 @@ import (
 	"time"
 )
 
-var RetryAttempts = 7
-
 // done returns whether the context's Done channel was closed due to
 // cancellation or exceeded deadline.
 func ContextDone(ctx context.Context) bool {
@@ -20,17 +18,21 @@ func ContextDone(ctx context.Context) bool {
 	}
 }
 
-func RetryFunc(attempts int, sleep time.Duration, errFunc func() error) (att int, err error) {
-	for i := 0; i < attempts; i++ {
+// RetryFunc implement retry policy for processes that needs to be executed
+// in the background after initial failure.
+func RetryFunc(retryAttempts int, sleepDur time.Duration, errFunc func() error) (int, error) {
+	var err error
+	for i := 0; i < retryAttempts; i++ {
 		if i > 0 {
-			log.Println("retrying after error:", err)
-			time.Sleep(sleep)
-			sleep *= 2
+			sleepDur *= 2
+			log.Printf("waiting %v to retry after error: %v", sleepDur, err)
+			time.Sleep(sleepDur)
 		}
 		err = errFunc()
 		if err == nil {
-			return -1, nil
+			return i, nil
 		}
 	}
-	return attempts, fmt.Errorf("after %d attempts, last error: %s", attempts, err)
+
+	return retryAttempts, fmt.Errorf("after %d attempts, last error: %s", retryAttempts, err)
 }
--- a/ui/page/components/utils.go
+++ b/ui/page/components/utils.go
@@ -1,6 +1,13 @@
 package components
 
-import "context"
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+)
+
+var RetryAttempts = 7
 
 // done returns whether the context's Done channel was closed due to
 // cancellation or exceeded deadline.
@@ -11,4 +18,19 @@ func ContextDone(ctx context.Context) bool {
 	default:
 		return false
 	}
+}
+
+func RetryFunc(attempts int, sleep time.Duration, errFunc func() error) (att int, err error) {
+	for i := 0; i < attempts; i++ {
+		if i > 0 {
+			log.Println("retrying after error:", err)
+			time.Sleep(sleep)
+			sleep *= 2
+		}
+		err = errFunc()
+		if err == nil {
+			return -1, nil
+		}
+	}
+	return attempts, fmt.Errorf("after %d attempts, last error: %s", attempts, err)
 }
--- a/ui/page/main_page.go
+++ b/ui/page/main_page.go
@@ -1,12 +1,16 @@
 package page
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
+	"time"
 
+	"gioui.org/font/gofont"
 	"gioui.org/layout"
 	"gioui.org/unit"
 	"gioui.org/widget"
+	"gioui.org/widget/material"
 
 	"github.com/decred/dcrd/dcrutil/v4"
 	"github.com/planetdecred/dcrlibwallet"
@@ -31,6 +35,8 @@ const (
 var (
 	NavDrawerWidth          = unit.Value{U: unit.UnitDp, V: 160}
 	NavDrawerMinimizedWidth = unit.Value{U: unit.UnitDp, V: 72}
+
+	retryAttempts = 10
 )
 
 type HideBalanceItem struct {
@@ -59,12 +65,16 @@ type MainPage struct {
 	sendPage      *send.Page   // reuse value to keep data persistent onresume.
 	receivePage   *ReceivePage // pointer to receive page. to avoid duplication.
 
+	loader       material.LoaderStyle
+	refreshValue *decredmaterial.Clickable
+
 	// page state variables
-	usdExchangeSet  bool
-	dcrUsdtBittrex  load.DCRUSDTBittrex
-	isBalanceHidden bool
-	totalBalance    dcrutil.Amount
-	totalBalanceUSD string
+	usdExchangeSet    bool
+	isLoadingUSDValue bool
+	dcrUsdtBittrex    load.DCRUSDTBittrex
+	isBalanceHidden   bool
+	totalBalance      dcrutil.Amount
+	totalBalanceUSD   string
 }
 
 func NewMainPage(l *load.Load) *MainPage {
@@ -91,6 +101,11 @@ func NewMainPage(l *load.Load) *MainPage {
 	l.PopToFragment = mp.popToFragment
 
 	mp.initNavItems()
+
+	th := material.NewTheme(gofont.Collection())
+	mp.loader = material.Loader(th)
+
+	mp.refreshValue = mp.Theme.NewClickable(true)
 
 	// Show a seed backup prompt if necessary.
 	mp.WL.Wallet.SaveConfigValueForKey(load.SeedBackupNotificationConfigKey, false)
@@ -200,6 +215,8 @@ func (mp *MainPage) OnNavigatedTo() {
 
 	mp.setLanguageSetting()
 
+	go mp.fetchExchangeValue()
+
 	if mp.currentPage == nil {
 		mp.currentPage = overview.NewOverviewPage(mp.Load)
 	}
@@ -218,6 +235,8 @@ func (mp *MainPage) OnNavigatedTo() {
 			go mp.WL.MultiWallet.Politeia.Sync()
 		}
 	}
+
+	mp.UpdateBalance()
 }
 
 func (mp *MainPage) setLanguageSetting() {
@@ -225,10 +244,48 @@ func (mp *MainPage) setLanguageSetting() {
 	values.SetUserLanguage(langPre)
 }
 
-func (mp *MainPage) UpdateBalance(fetchExch bool) {
-	if len(mp.totalBalanceUSD) == 0 || fetchExch {
+func (mp *MainPage) fetchExchangeValue() {
+	err, attempts := retry(retryAttempts, 2*time.Second, func() error {
 		load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
+		if mp.dcrUsdtBittrex.LastTradeRate != "" {
+			log.Info("exchange value fetched")
+			mp.UpdateBalance()
+			mp.RefreshWindow()
+			mp.isLoadingUSDValue = false
+			return nil
+		}
+
+		mp.isLoadingUSDValue = true
+		return errors.New("failed to fetch dcrUsdtBittrex value")
+	})
+
+	if err != nil || mp.dcrUsdtBittrex.LastTradeRate == "" {
+		log.Error(err)
+
+		if attempts == retryAttempts {
+			mp.isLoadingUSDValue = false
+		}
+		return
 	}
+
+}
+
+func retry(attempts int, sleep time.Duration, errFunc func() error) (err error, att int) {
+	for i := 0; i < attempts; i++ {
+		if i > 0 {
+			log.Error("retrying after error:", err)
+			time.Sleep(sleep)
+			sleep *= 2
+		}
+		err = errFunc()
+		if err == nil {
+			return nil, -1
+		}
+	}
+	return fmt.Errorf("after %d attempts, last error: %s", attempts, err), attempts
+}
+
+func (mp *MainPage) UpdateBalance() {
 	currencyExchangeValue := mp.WL.Wallet.ReadStringConfigValueForKey(dcrlibwallet.CurrencyConversionConfigKey)
 	mp.usdExchangeSet = currencyExchangeValue == values.USDExchangeValue
 
@@ -315,6 +372,10 @@ func (mp *MainPage) UnlockWalletForSyncing(wal *dcrlibwallet.Wallet) {
 func (mp *MainPage) HandleUserInteractions() {
 	if mp.currentPage != nil {
 		mp.currentPage.HandleUserInteractions()
+	}
+
+	if mp.refreshValue.Clicked() && mp.usdExchangeSet {
+		go mp.fetchExchangeValue()
 	}
 
 	mp.drawerNav.CurrentPage = mp.currentPageID()
@@ -497,7 +558,6 @@ func (mp *MainPage) popToFragment(pageID string) {
 // to be eventually drawn on screen.
 // Part of the load.Page interface.
 func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
-	mp.UpdateBalance(false)
 	return layout.Stack{}.Layout(gtx,
 		layout.Expanded(func(gtx C) D {
 			return decredmaterial.LinearLayout{
@@ -554,31 +614,47 @@ func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
 }
 
 func (mp *MainPage) LayoutUSDBalance(gtx layout.Context) layout.Dimensions {
-	return layout.Flex{Axis: layout.Vertical}.Layout(gtx,
-		layout.Rigid(func(gtx C) D {
-			if mp.usdExchangeSet && mp.dcrUsdtBittrex.LastTradeRate != "" && len(mp.totalBalanceUSD) > 0 {
-				inset := layout.Inset{
-					Top:  values.MarginPadding3,
-					Left: values.MarginPadding8,
-				}
-				border := widget.Border{Color: mp.Theme.Color.Gray2, CornerRadius: unit.Dp(8), Width: unit.Dp(0.5)}
-				return inset.Layout(gtx, func(gtx C) D {
-					padding := layout.Inset{
-						Top:    values.MarginPadding3,
-						Bottom: values.MarginPadding3,
-						Left:   values.MarginPadding6,
-						Right:  values.MarginPadding6,
-					}
-					return border.Layout(gtx, func(gtx C) D {
-						return padding.Layout(gtx, func(gtx C) D {
-							return mp.Theme.Body2(mp.totalBalanceUSD).Layout(gtx)
-						})
-					})
-				})
+	switch {
+	case mp.usdExchangeSet && mp.isLoadingUSDValue && mp.dcrUsdtBittrex.LastTradeRate == "":
+		gtx.Constraints.Max.Y = gtx.Px(values.MarginPadding18)
+		gtx.Constraints.Max.X = gtx.Constraints.Max.Y
+		return layout.Inset{
+			Top:  values.MarginPadding8,
+			Left: values.MarginPadding5,
+		}.Layout(gtx, func(gtx C) D {
+			return mp.loader.Layout(gtx)
+		})
+	case mp.usdExchangeSet && !mp.isLoadingUSDValue && mp.dcrUsdtBittrex.LastTradeRate == "":
+		return layout.Inset{
+			Top:  values.MarginPadding7,
+			Left: values.MarginPadding5,
+		}.Layout(gtx, func(gtx C) D {
+			return mp.refreshValue.Layout(gtx, func(gtx C) D {
+				return mp.Icons.Restore.Layout16dp(gtx)
+			})
+		})
+	case mp.usdExchangeSet && mp.dcrUsdtBittrex.LastTradeRate != "" && len(mp.totalBalanceUSD) > 0:
+		inset := layout.Inset{
+			Top:  values.MarginPadding3,
+			Left: values.MarginPadding8,
+		}
+		border := widget.Border{Color: mp.Theme.Color.Gray2, CornerRadius: unit.Dp(8), Width: unit.Dp(0.5)}
+		return inset.Layout(gtx, func(gtx C) D {
+			padding := layout.Inset{
+				Top:    values.MarginPadding3,
+				Bottom: values.MarginPadding3,
+				Left:   values.MarginPadding6,
+				Right:  values.MarginPadding6,
 			}
-			return D{}
-		}),
-	)
+			return border.Layout(gtx, func(gtx C) D {
+				return padding.Layout(gtx, func(gtx C) D {
+					return mp.Theme.Body2(mp.totalBalanceUSD).Layout(gtx)
+				})
+			})
+		})
+	default:
+		return D{}
+	}
 }
 
 func (mp *MainPage) totalDCRBalance(gtx layout.Context) layout.Dimensions {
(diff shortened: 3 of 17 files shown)


ui/page/send/page.go
```diff
@@ -163,34 +166,51 @@ func (pg *Page) OnNavigatedTo() {
...
-		var dcrUsdtBittrex load.DCRUSDTBittrex
+	var dcrUsdtBittrex load.DCRUSDTBittrex
+	errText := "failed to fetch dcrUsdtBittrex value"
+	attempts, err := components.RetryFunc(components.RetryAttempts, 2*time.Second, func() error {
 		err := load.GetUSDExchangeValue(&dcrUsdtBittrex)
 		if err != nil {
-			pg.exchangeError = err.Error()
-			return
+			if strings.Contains(err.Error(), "no such host") {
+				pg.exchangeError = "Exchange rate not fetched. Kinly check internet connection. Retrying..."
```
devchoplife:
```suggestion
				pg.exchangeError = "Exchange rate not fetched. Kindly check internet connection. Retrying..."
```

devchoplife (changes requested):

ui/load/load.go
```diff
@@ -83,6 +83,8 @@ type Load struct {
 	SubscribeKeyEvent   func(eventChan chan *key.Event, pageID string) // Widgets call this function to recieve key events.
 	UnsubscribeKeyEvent func(pageID string) error
 	ReloadApp           func()
+
+	IsCurrencySettingUpdated bool
```
itswisdomagain: Concerned about this. Not sure why this should be this global. Revisit.

ui/page/components/utils.go
```diff
@@ -1,6 +1,13 @@
...
 
-import "context"
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+)
+
+var RetryAttempts = 7
```
itswisdomagain: This isn't used here, perhaps this isn't the best/right place to define this.

ui/page/components/utils.go
```diff
@@ -12,3 +19,18 @@ func ContextDone(ctx context.Context) bool {
 		return false
 	}
 }
+
+func RetryFunc(attempts int, sleep time.Duration, errFunc func() error) (att int, err error) {
```
itswisdomagain: - Needs a comment.
- Also `errFunc` -> `runFunc`, it mustn't return an error, it may return a nil error which is valid.
- `att` is never used/assigned. Remove the named return variables and instead declare a `var err error` before the for loop below.

ui/page/components/utils.go
```diff
@@ -12,3 +19,18 @@ func ContextDone(ctx context.Context) bool {
...
+	for i := 0; i < attempts; i++ {
+		if i > 0 {
+			log.Println("retrying after error:", err)
+			time.Sleep(sleep)
+			sleep *= 2
```
itswisdomagain: I'd reorder this, this way (PS: for logging purposes, it'd be helpful to pass a `funcDesc string` parameter that can be used in the suggested log below, I also renamed `sleep` to `sleepDur`):
```suggestion
			sleepDur *= i
			log.Println("waiting %s to retry function %s after error: %v", sleepDur, funcDesc, err)
			time.Sleep(sleepDur)
```

ui/page/components/utils.go
```diff
@@ -12,3 +19,18 @@ func ContextDone(ctx context.Context) bool {
...
+		}
+		err = errFunc()
+		if err == nil {
+			return -1, nil
```
itswisdomagain: `return i, nil`?

ui/page/components/utils.go
```diff
@@ -12,3 +19,18 @@ func ContextDone(ctx context.Context) bool {
...
+		}
+	}
+	return attempts, fmt.Errorf("after %d attempts, last error: %s", attempts, err)
```
itswisdomagain: Depending on how this returned error is used, this message might not be clear enough.

ui/page/main_page.go
```diff
@@ -59,12 +63,16 @@ type MainPage struct {
 	sendPage      *send.Page   // reuse value to keep data persistent onresume.
 	receivePage   *ReceivePage // pointer to receive page. to avoid duplication.
 
+	loader       material.LoaderStyle
+	refreshValue *decredmaterial.Clickable
```
itswisdomagain: Is `loader` a page loader or specific to some page element? The name can use some love.
`refreshValue` is a clickable that refreshes a value? Perhaps `refreshExchangeRateBtn`?

ui/page/main_page.go
```diff
@@ -218,20 +231,51 @@ func (mp *MainPage) OnNavigatedTo() {
 			go mp.WL.MultiWallet.Politeia.Sync()
 		}
 	}
+
+	mp.updateExchangeSetting()
+
+	mp.UpdateBalance()
```
itswisdomagain, line 229: Does fetching the balance take time? If it doesn't, this should be in `OnNavigatedTo` since it doesn't make sense to fetch the balance when the page isn't being displayed yet, while it makes sense to refresh the balance whenever the page is about to be displayed.

ui/page/main_page.go
```diff
@@ -218,20 +231,51 @@ func (mp *MainPage) OnNavigatedTo() {
...
+func (mp *MainPage) updateExchangeSetting() {
 	currencyExchangeValue := mp.WL.Wallet.ReadStringConfigValueForKey(dcrlibwallet.CurrencyConversionConfigKey)
 	mp.usdExchangeSet = currencyExchangeValue == values.USDExchangeValue
+	if mp.usdExchangeSet {
+		go mp.fetchExchangeValue()
+	}
+}
+
+func (mp *MainPage) fetchExchangeValue() {
+	attempts, err := components.RetryFunc(components.RetryAttempts, 2*time.Second, func() error {
```
itswisdomagain: It doesn't make much sense to me for `components` to have a `RetryAttempts` property - how many times something should be retried is best defined by the caller of the function. I'd rather declare and use a `const fetchExchangeRateRetries` field in this file.

ui/page/main_page.go
```diff
@@ -218,20 +231,51 @@ func (mp *MainPage) OnNavigatedTo() {
...
-	}
+func (mp *MainPage) updateExchangeSetting() {
 	currencyExchangeValue := mp.WL.Wallet.ReadStringConfigValueForKey(dcrlibwallet.CurrencyConversionConfigKey)
 	mp.usdExchangeSet = currencyExchangeValue == values.USDExchangeValue
+	if mp.usdExchangeSet {
+		go mp.fetchExchangeValue()
+	}
+}
+
+func (mp *MainPage) fetchExchangeValue() {
```
itswisdomagain:
```suggestion
func (mp *MainPage) fetchExchangeRate() {
```

ui/page/main_page.go
```diff
@@ -218,20 +231,51 @@ func (mp *MainPage) OnNavigatedTo() {
...
+
+	if err != nil || mp.dcrUsdtBittrex.LastTradeRate == "" {
+		log.Error(err)
+
+		if attempts == components.RetryAttempts {
+			mp.isLoadingUSDValue = false
+		}
+		return
+	}
+}
```
itswisdomagain: I initially had a few thoughts about this method, found it easier to just recommend an alternative snippet:
```
func (mp *MainPage) fetchExchangeValue() {
	if mp.isLoadingUSDValue {
		return
	}
	maxAttempts := 7
	delayBtwAttempts := 2 * time.Second
	mp.isLoadingUSDValue = true
	attempts, err := components.RetryFunc(maxAttempts, delayBtwAttempts, func() error {
		return load.GetUSDExchangeValue(&mp.dcrUsdtBittrex)
	})
	if err != nil {
		log.Errorf("error fetching usd exchange rate value after %d attempts: %v", attempts, err)
	} else if mp.dcrUsdtBittrex.LastTradeRate == "" {
		log.Errorf("no error while fetching usd exchange rate in %d tries, but no rate was fetched", attempts)
	} else {
		log.Infof("exchange rate value fetched: %s", mp.dcrUsdtBittrex.LastTradeRate)
		mp.UpdateBalance()
		mp.RefreshWindow()
	}
	mp.isLoadingUSDValue = false
}
```

ui/page/main_page.go
```diff
@@ -313,10 +357,19 @@ func (mp *MainPage) UnlockWalletForSyncing(wal *dcrlibwallet.Wallet) {
 // displayed.
 // Part of the load.Page interface.
 func (mp *MainPage) HandleUserInteractions() {
+	if mp.IsCurrencySettingUpdated {
```
itswisdomagain: Not sure what is meant by `currency setting`, this name needs some love at least. Also, a global variable of this form to track when a particular setting is changed doesn't sound good. Will propose something later.

ui/page/main_page.go
```diff
@@ -313,10 +357,19 @@ func (mp *MainPage) UnlockWalletForSyncing(wal *dcrlibwallet.Wallet) {
...
+		mp.updateExchangeSetting()
+		mp.IsCurrencySettingUpdated = false
+	}
+
 	if mp.currentPage != nil {
 		mp.currentPage.HandleUserInteractions()
 	}
 
+	if mp.refreshValue.Clicked() && mp.usdExchangeSet {
```
itswisdomagain: Would there even be a refresh button to click if the exchange rate setting isn't configured?
```suggestion
	if mp.refreshValue.Clicked() {
```

ui/page/main_page.go
```diff
@@ -554,31 +606,47 @@ func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
...
-				})
+	switch {
+	case mp.usdExchangeSet && mp.isLoadingUSDValue && mp.dcrUsdtBittrex.LastTradeRate == "":
+		gtx.Constraints.Max.Y = gtx.Px(values.MarginPadding18)
+		gtx.Constraints.Max.X = gtx.Constraints.Max.Y
+		return layout.Inset{
+			Top:  values.MarginPadding8,
+			Left: values.MarginPadding5,
+		}.Layout(gtx, func(gtx C) D {
+			return mp.loader.Layout(gtx)
```
itswisdomagain: Instead of having a `mp.loader` property always assigned and ready, which may never be used or will not be needed after some point, create the loader object here and lay it out - strictly when needed.
```suggestion
			loader := material.NewTheme(gofont.Collection()).Loader(mp.Theme)
			return loader.Layout(gtx)
```

ui/page/main_page.go
```diff
@@ -554,31 +606,47 @@ func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
...
-						Right:  values.MarginPadding6,
-					}
-					return border.Layout(gtx, func(gtx C) D {
-						return padding.Layout(gtx, func(gtx C) D {
-							return mp.Theme.Body2(mp.totalBalanceUSD).Layout(gtx)
-						})
-					})
-				})
+	switch {
+	case mp.usdExchangeSet && mp.isLoadingUSDValue && mp.dcrUsdtBittrex.LastTradeRate == "":
```
itswisdomagain: Seems all the cases require `mp.usdExchangeSet`. Can remove that condition from all the cases and do this before the switch:
```
if !mp.usdExchangeSet { return D{} }
```

ui/page/main_page.go
```diff
@@ -554,31 +606,47 @@ func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
...
+	case mp.usdExchangeSet && !mp.isLoadingUSDValue && mp.dcrUsdtBittrex.LastTradeRate == "":
+		return layout.Inset{
+			Top:  values.MarginPadding7,
+			Left: values.MarginPadding5,
+		}.Layout(gtx, func(gtx C) D {
+			return mp.refreshValue.Layout(gtx, func(gtx C) D {
+				return mp.Icons.Restore.Layout16dp(gtx)
+			})
+		})
+	case mp.usdExchangeSet && mp.dcrUsdtBittrex.LastTradeRate != "" && len(mp.totalBalanceUSD) > 0:
```
itswisdomagain: Looks like this should only be concerned with `len(mp.totalBalanceUSD) > 0` - display the USD amount as long as it is already calculated.
```suggestion
	case len(mp.totalBalanceUSD) > 0:
```

itswisdomagain (changes requested):
First pass. Will re-review after these.

ui/load/load.go
Sirmorrison: need the info here to update the main page from list_preference.go

ui/page/components/utils.go
Sirmorrison: Why so is i didnt want to re-declare same for the individual pages. (send page and main page)

ui/load/load.go
itswisdomagain: I saw this, works for now but isn't ideal. I made a related comment below.

ui/page/components/utils.go
itswisdomagain: I recommended an alternative in the main page that can also apply to the send page. This is one scenario where duplication is okay.

ui/page/main_page.go
Sirmorrison: it is currently in the `OnNavigatedTo()`

ui/page/components/utils.go
Sirmorrison: `sleepDur *= i` `i` does not work... Mismatch type.

ui/page/components/utils.go
```diff
@@ -12,3 +17,22 @@ func ContextDone(ctx context.Context) bool {
 		return false
 	}
 }
+
+// RetryFunc implement retry policy for processes that needs to be executed
+// in the background after initial failure.
```
itswisdomagain: This function doesn't execute any process in the background. This comment could be (is) misleading.

ui/page/components/utils.go
```diff
@@ -12,3 +17,22 @@ func ContextDone(ctx context.Context) bool {
...
+func RetryFunc(retryAttempts int, sleepDur time.Duration, errFunc func() error) (int, error) {
+	var err error
+	for i := 0; i < retryAttempts; i++ {
+		if i > 0 {
+			sleepDur *= 2
```
itswisdomagain:
```suggestion
			sleepDur *= (2-i)
```
So when i == 1, sleepDur is sleepDur *= 1 and only begins to increase when i > 1.

ui/page/components/utils.go
```diff
@@ -12,3 +17,22 @@ func ContextDone(ctx context.Context) bool {
...
+			log.Printf("waiting %v to retry after error: %v", sleepDur, err)
```
itswisdomagain: To retry what? This wouldn't be a very helpful log message. A previous review comment recommended adding a `funcDesc` parameter to this function which can be included in this log for better clarity.

ui/page/main_page.go
itswisdomagain: Then we certainly don't need it here. Any balance calculated here will not be displayed until `OnNavigatedTo` is called and that will re-calculate the balance.

ui/page/send/layout.go
```diff
@@ -276,11 +278,17 @@ func (pg *Page) toSection(gtx layout.Context) layout.Dimensions {
 					layout.Rigid(func(gtx C) D {
 						return layout.Flex{Axis: layout.Horizontal}.Layout(gtx,
 							layout.Rigid(func(gtx C) D {
-								label := pg.Theme.Body2("Exchange rate not fetched")
+								label := pg.Theme.Body2(pg.exchangeError)
```
itswisdomagain: This should just be `"Exchange rate not fetched. Check internet connection."` I explain why below.

ui/page/send/layout.go
```diff
@@ -276,11 +278,17 @@ func (pg *Page) toSection(gtx layout.Context) layout.Dimensions {
...
 								label.Color = pg.Theme.Color.Danger
+								if strings.Contains(pg.exchangeError, "Retrying") {
+									label.Color = pg.Theme.Color.Primary
+								}
 								return label.Layout(gtx)
 							}),
 							layout.Rigid(func(gtx C) D {
+								if strings.Contains(pg.exchangeError, "Retrying") {
```
itswisdomagain: Sounds like you need a `pg.isFetchingExchangeRate` bool property, as you have in the mainpage. I see that `pg.exchangeError` is set for each failed attempt but the set error message may not show up on the UI without a `RefreshWindow` call. I think you should display a loader or some brief _`fetching exchange rate`_ text while the fetching is ongoing instead of showing an error message for each failed try. At the end of all tries, `if !pg.isFetchingExchangeRate && (the rate isn't ready)`, you can then show the message I recommended above.

ui/page/send/page.go
```diff
@@ -163,34 +166,54 @@ func (pg *Page) OnNavigatedTo() {
...
+	errText := "failed to fetch dcrUsdtBittrex value"
+
+	maxAttempts := 7
+	delayBtwAttempts := 2 * time.Second
+	attempts, err := components.RetryFunc(maxAttempts, delayBtwAttempts, func() error {
 		err := load.GetUSDExchangeValue(&dcrUsdtBittrex)
 		if err != nil {
-			pg.exchangeError = err.Error()
-			return
+			if strings.Contains(err.Error(), "no such host") {
```
itswisdomagain: This string comparison is not reliable, best to do without it and always assume internet connection issues. Secondly, no need to display an error message each time `load.GetUSDExchangeValue` returns an error. Just return the error and let the retrials continue to  happen in the background. Only show an error message when the rate cannot be fetched after the maxAttempts.

itswisdomagain (changes requested):
The send page's `fetchExchangeRate()` should be a lot more like the recently modified main page method of the same name.

ui/page/components/utils.go
Sirmorrison: it does not work getting the following error. `invalid operation: sleepDur *= 2 - i (mismatched types time.Duration and int)`

using `time.Duration(2 - i)` results to 0s.

ui/page/main_page.go
Sirmorrison: From my understanding, want to be able to update the balance when the page loads, and when there is an update from a listener. 

Moving this from `OnNavigatedTo`, i am not sure where else to place it. You can review the update i did if it works.

itswisdomagain: Oy my, missed the fact that is in `OnNavigatedTo`. Sorry, it's fine here.

ui/page/components/utils.go
```diff
@@ -12,3 +17,22 @@ func ContextDone(ctx context.Context) bool {
...
 	}
 }
+
+// RetryFunc implement retry policy for processes that needs to be executed
+// the after initial failure.
+func RetryFunc(retryAttempts int, sleepDur time.Duration, funcDesc string, errFunc func() error) (int, error) {
+	var err error
+	for i := 0; i < retryAttempts; i++ {
+		if i > 0 {
+			sleepDur *= 2
```
itswisdomagain: Then surround this with an if statement:
```suggestion
			if i > 1 { sleepDur *= 2 }
```

ui/page/send/page.go
```diff
@@ -163,34 +166,49 @@ func (pg *Page) OnNavigatedTo() {
...
+		pg.ExchangeRateMassage = ""
 		exchangeRate, err := strconv.ParseFloat(dcrUsdtBittrex.LastTradeRate, 64)
 		if err != nil {
-			pg.exchangeError = err.Error()
+			pg.ExchangeRateMassage = err.Error()
 			return
 		}
 
-		pg.exchangeError = ""
+		pg.ExchangeRateMassage = ""
```
itswisdomagain: Already cleared above.

ui/page/send/page.go
```diff
@@ -54,12 +56,13 @@ type Page struct {
...
+	moreOptionIsOpen       bool
+	isFetchingExchangeRate bool
 
-	exchangeRate   float64
-	usdExchangeSet bool
-	exchangeError  string
-	confirmTxModal *sendConfirmModal
+	exchangeRate        float64
+	usdExchangeSet      bool
+	ExchangeRateMassage string
```
itswisdomagain:
```suggestion
	exchangeRateMessage string
```

ui/preference/list_preference.go
```diff
@@ -108,6 +108,9 @@ func (lp *ListPreferenceModal) Handle() {
 		lp.currentValue = lp.optionsRadioGroup.Value
 		lp.wallet.SaveConfigValueForKey(lp.preferenceKey, lp.optionsRadioGroup.Value)
 		lp.updateButtonClicked()
+		if lp.title == values.StrCurrencyConversion {
+			lp.IsCurrencySettingUpdated = true
+		}
 		lp.DismissModal(lp)
```
itswisdomagain: I think we can do away with the `IsCurrencySettingUpdated` field now. `lp.DismissModal` will trigger a window refresh, which will call the main page's `Layout` method. The main page's layout method can call the `updateExchangeSetting()` method to check if the setting has changed. That method will need to be re-written like this:

```
func (mp *MainPage) updateExchangeSetting() {
	currencyExchangeValue := mp.WL.Wallet.ReadStringConfigValueForKey(dcrlibwallet.CurrencyConversionConfigKey)
	usdExchangeSet := currencyExchangeValue == values.USDExchangeValue
	if mp.usdExchangeSet == usdExchangeSet {
		return // nothing has changed
	}
	mp.usdExchangeSet = usdExchangeSet
	if mp.usdExchangeSet {
		go mp.fetchExchangeRate()
	}
}
```

ui/page/main_page.go
```diff
@@ -218,20 +226,48 @@ func (mp *MainPage) OnNavigatedTo() {
...
+	} else {
+		log.Infof("exchange rate value fetched: %s", mp.dcrUsdtBittrex.LastTradeRate)
+		mp.UpdateBalance()
+		mp.RefreshWindow()
+	}
+	mp.isFetchingExchangeRate = false
+}
 
+func (mp *MainPage) UpdateBalance() {
+	mp.updateExchangeSetting()
```
itswisdomagain: This isn't necessary here. This only needs to be done when the setting has likely changed. My previous comment recommends making this check in `Layout`.

itswisdomagain (changes requested):

ui/page/components/utils.go
```diff
@@ -12,3 +17,24 @@ func ContextDone(ctx context.Context) bool {
 		return false
 	}
 }
+
+// RetryFunc implement retry policy for processes that needs to be executed
+// the after initial failure.
```
itswisdomagain:
```suggestion
// RetryFunc implements retry policy for processes that needs to be executed
// after initial failure.
```

ui/page/main_page.go
```diff
@@ -497,7 +540,7 @@ func (mp *MainPage) popToFragment(pageID string) {
 // to be eventually drawn on screen.
 // Part of the load.Page interface.
 func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
-	mp.UpdateBalance(false)
+	mp.updateExchangeSetting()
```
itswisdomagain: Let's add a comment here to give future code maintainers an idea why this is being done each time `Layout` is called.
```suggestion
	mp.updateExchangeSetting() // the setting may have changed, leading to this window refresh, let's check
```

ui/page/send/page.go
```diff
@@ -163,34 +166,48 @@ func (pg *Page) OnNavigatedTo() {
...
 		}
 
-		pg.exchangeError = ""
 		pg.exchangeRate = exchangeRate
 		pg.amount.setExchangeRate(exchangeRate)
 		pg.validateAndConstructTx() // convert estimates to usd
-	}()
+		pg.RefreshWindow()
+	}
+	pg.isFetchingExchangeRate = false
```
itswisdomagain: Call `pg.RefreshWindow()` after setting `pg.isFetchingExchangeRate = false`. This ensures that the updated `pg.exchangeRateMessage` is displayed on the UI even if there was an error fetching the exchange rate.

ui/page/send/page.go
```diff
@@ -163,34 +166,48 @@ func (pg *Page) OnNavigatedTo() {
...
 		if err != nil {
-			pg.exchangeError = err.Error()
+			pg.exchangeRateMessage = err.Error()
 			return
 		}
 
-		pg.exchangeError = ""
 		pg.exchangeRate = exchangeRate
 		pg.amount.setExchangeRate(exchangeRate)
 		pg.validateAndConstructTx() // convert estimates to usd
```
itswisdomagain: If you return when err != nil, the `isFetchingExchangeRate` value won't be set to false and the window will not refresh to show the new message.
This instead:
```
		if err != nil {
			pg.exchangeRateMessage = err.Error()
		} else {
			pg.exchangeRate = exchangeRate
			pg.amount.setExchangeRate(exchangeRate)
			pg.validateAndConstructTx() // convert estimates to usd
		}
```

ui/page/components/utils.go
```diff
@@ -12,3 +17,24 @@ func ContextDone(ctx context.Context) bool {
...
+func RetryFunc(retryAttempts int, sleepDur time.Duration, funcDesc string, errFunc func() error) (int, error) {
+	var err error
+	for i := 0; i < retryAttempts; i++ {
+		if i > 0 {
+			if i > 1 {
+				sleepDur *= 2
+			}
+			log.Printf("waiting %s to retry function %s after error: %v", sleepDur, funcDesc, err)
```
itswisdomagain: Just seeing this, `log` is an imported package. Not sure how it works but it certainly does not use the same logging pattern I've seen in the rest of this godcr codebase. I don't think the components pkg has a logger yet, I think #777 introduces one. This can be `fmt.Printf` for now and should end with a `\n`.

ui/page/components/utils.go
```diff
@@ -12,3 +17,24 @@ func ContextDone(ctx context.Context) bool {
...
+			time.Sleep(sleepDur)
+		}
+		err = errFunc()
+		if err == nil {
+			return i, nil
+		}
+	}
+
+	return retryAttempts, fmt.Errorf("after %d attempts, last error: %s", retryAttempts, err)
```
itswisdomagain: This doesn't read well when paired with another message, it currently says `error fetching usd exchange rate value after 3 attempts: after 3 attempts, last error: Get "https://api.bittrex.com/v3/markets/DCR-USDT/ticker": dial tcp: lookup api.bittrex.com: no such host`.

Can omit the number of tries, the caller already knows.
```suggestion
	return retryAttempts, fmt.Errorf("last error: %s", err)
```

ui/page/main_page.go
```diff
@@ -218,20 +226,51 @@ func (mp *MainPage) OnNavigatedTo() {
...
+	if mp.usdExchangeSet {
+		go mp.fetchExchangeRate()
+	}
+}
 
+func (mp *MainPage) fetchExchangeRate() {
+	if mp.isFetchingExchangeRate {
+		return
+	}
+	maxAttempts := 7
```
itswisdomagain: In testing, this is quite a long time. I've experimented with 3 tries and that seems okay. 3 tries will wait for a total of 6 seconds, aside the time each network call attempt takes. Can experiment with a different figure and see what is better for UX. While we want to try repeatedly, user should know soon enough when it isn't working so they can check their connection. Applies to send page as well.

itswisdomagain (changes requested):

ui/page/components/utils.go
```diff
@@ -12,3 +16,24 @@ func ContextDone(ctx context.Context) bool {
 		return false
 	}
 }
+
+// RetryFunc implements retry policy for processes that needs to be executed
+// the after initial failure.
```
itswisdomagain:
```suggestion
// after initial failure.
```

ui/page/main_page.go
```diff
@@ -554,31 +597,51 @@ func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
...
+	}
+	switch {
+	case mp.isFetchingExchangeRate && mp.dcrUsdtBittrex.LastTradeRate == "":
+		gtx.Constraints.Max.Y = gtx.Px(values.MarginPadding18)
+		gtx.Constraints.Max.X = gtx.Constraints.Max.Y
+		return layout.Inset{
+			Top:  values.MarginPadding8,
+			Left: values.MarginPadding5,
+		}.Layout(gtx, func(gtx C) D {
+			loader := material.Loader(material.NewTheme(gofont.Collection()))
```
itswisdomagain: I jsut remembered we already have a theme!
```suggestion
			loader := material.Loader(mp.Theme.Base)
```

itswisdomagain (dismissed):
2 nits, good to go after these

dreacot (approved):
LGTM

PR #826, opened 2022-03-24
Add wallet balance to staking UI

### LINKED ISSUE #781: Show available/locked balance in Staking UI
Currently there is no way to see how much DCR you have available for staking or how much is already in a locked staking state. 

Add something like this:
Untitled 2

Sirmorrison:
Fix #781

This PR add total wallet balance to the staking page with progress bar showing locked and spendable balance.

image

The use of a vertical alignment layout instead of the horizontal alignment was due to space constraint as the balance tend to wrap to the next line.

image

* Possible options are 
image

### DIFF
--- a/ui/page/main_page.go
+++ b/ui/page/main_page.go
@@ -298,27 +298,6 @@ func (mp *MainPage) updateBalance() {
 	}
 }
 
-// func (mp *MainPage) CalculateTotalWalletsBalance() (dcrutil.Amount, error) {
-// 	totalBalance := int64(0)
-// wallets := mp.WL.SortedWalletList()
-// 	if len(wallets) == 0 {
-// 		return 0, nil
-// 	}
-
-// 	for _, wallet := range wallets {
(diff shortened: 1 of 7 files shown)


ui/page/main_page.go
```diff
@@ -298,26 +298,26 @@ func (mp *MainPage) updateBalance() {
...
+// 			return 0, err
+// 		}
+
+// 		for _, account := range accountsResult.Acc {
+// 			totalBalance += account.TotalBalance
+// 		}
+// 	}
+
+// 	return dcrutil.Amount(totalBalance), nil
+// }
```
dreacot: this commented code doesn't seem relevant again, sine we have another method with the same name that does the work

dreacot (changes requested):
The texts above the progress bar are hardly readable, could we increase it a bit?

dreacot (approved):
LGTM

PR #848, opened 2022-03-29
set default exchange and currency value on app first load

### LINKED ISSUE #838: Currency conversion and app language bug
Launching the godcr app for the first time and going to the settings page, the  selected/defualt currency conversion value and language isnt displayed until it is selected the first time. it should show the default.

image

Sirmorrison:
fix #838.

easiest way to test is to rename your current testnet3 to test net and start setup afresh.


dreacot (approved):
LGTM

PR #850, opened 2022-03-29
update send page darkmode colors

### LINKED ISSUE #831: Send page formatting in dark mode and text update
Update Send page to say "Destination Address" instead of just "Address".  

Additionally, the page rendering is not correct in dark mode for the send page:

Screen Shot 2022-03-25 at 2 35 49 PM

Sirmorrison:
Fix #831

This Pr updates the dark mode colors on the send page.

### DIFF
--- a/ui/decredmaterial/theme.go
+++ b/ui/decredmaterial/theme.go
@@ -102,10 +102,23 @@ func (t *Theme) updateStyles(isDarkModeOn bool) {
 	t.Styles.SwitchStyle.ActiveColor = t.Color.Primary
 	t.Styles.SwitchStyle.InactiveColor = t.Color.Gray3
 	t.Styles.SwitchStyle.ThumbColor = t.Color.White
+	t.Styles.SwitchStyle.ActiveTextColor = color.NRGBA{}
+	t.Styles.SwitchStyle.InactiveTextColor = color.NRGBA{}
+
+	// update SwitchButtonText style colors
+	t.Styles.SwitchButtonTextStyle.ActiveColor = t.Color.Surface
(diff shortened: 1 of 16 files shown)


dreacot (commented):
Screenshot from 2022-03-31 14-18-45

Switching from light to dark mode and then light mode back;
1. the max button is still in dark mode
2. the next button is still in dark mode
3. typing a text in the amount input field doesn't show the amount being inputted(it's white)

dreacot:
1. when the input fields are out of focus, the hint text are not dark mode compatible
2. the exchange icon between the dcr amount and usd amount are not dark mode compatible
3. the arrow to expand the fee section isn't dark mode compatible
Screenshot from 2022-04-13 21-21-50

the more section is dark mode compatible
Screenshot from 2022-04-13 21-32-01

ui/decredmaterial/theme.go
```diff
@@ -83,6 +81,7 @@ func (t *Theme) SwitchDarkMode(isDarkModeOn bool, decredIcons map[string]image.I
 	expandIcon := "expand_icon"
 	collapseIcon := "collapse_icon"
 	if isDarkModeOn {
+		fmt.Println("in theme 1")
```
dreacot: i'm guessing this was used for debugging purposes, we can remove if so

ui/window.go
```diff
@@ -144,6 +144,19 @@ func (win *Window) NewLoad() (*load.Load, error) {
 		}
 	}
 
+	// DarkModeSettingChanged checks if any page or any
+	// modal
```
dreacot: seems like an incomplete statement

dreacot (approved):
LGTM

PR #857, opened 2022-03-30
Copy updates for cspp manual setup and file name typo fix

monsa00:
fix filename typo and minor wording changes on cspp manual setup page.

### DIFF
--- a/ui/page/privacy/manual_mixer_setup_page.go
+++ b/ui/page/privacy/manual_mixer_setup_page.go
@@ -124,7 +124,7 @@ func (pg *ManualMixerSetupPage) Layout(gtx layout.Context) layout.Dimensions {
 											}),
 											layout.Rigid(func(gtx C) D {
 												txt := `<span style="text-color: grayText2">
-											<b>Make sure to select the corresponding accounts to the previous setup. </b> Failing to do so could damage wallet privacy.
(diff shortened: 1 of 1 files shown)


ui/page/privacy/manual_mixer_setup_page.go
```diff
@@ -124,7 +124,7 @@ func (pg *ManualMixerSetupPage) Layout(gtx layout.Context) layout.Dimensions {
 											}),
 											layout.Rigid(func(gtx C) D {
 												txt := `<span style="text-color: grayText2">
-											<b>Make sure to select the corresponding accounts to the previous setup. </b> Failing to do so could damage wallet privacy.
+											<b>Make sure to select the same accounts from the previous privacy setup. </b><br>Failing to do so could compromise wallet privacy.<br> You may not select the same account for mixed and unmixed.
```
Sirmorrison, line 127: the use of `May` from my point of view, make it feel like it a choice not to select same wallet for mixed and unmixed. I would rather use `Should`. 

```suggestion
											<b>Make sure to select the same accounts from the previous privacy setup. </b><br>Failing to do so could compromise wallet privacy.<br> You should not select the same account for mixed and unmixed.
```

monsa00: `should` makes it optional whereas `may` is mandatory. Could use `must` but I think `may` is better. Another option is `can not`.

Sirmorrison (approved):

PR #879, opened 2022-04-08
allow amount validation without address

### LINKED ISSUE #834: Send page should show an error for 'not enough funds' when an invalid amount is used.
Send page should show an error for not enough funds when an invalid amount is used, like mobile does. (this does work if destination is set, but it should always work)

### LINKED ISSUE #839: Max button update.
Currently, the max button depends on the destination address to be set else it returns an error to have the address set.

The max button should be updated to not depend on destination being set. 

The max amount should be calculated with any of the selected wallet valid address and when an address is provided, we recalculate the amount and update?

Sirmorrison:
Fix #834 
Fix #839 

* This PR allows amount validation with out destination address set.
* Set send address to destination address, bump dcrlibwallet version


ui/page/send/send_destination.go
```diff
@@ -27,7 +27,7 @@ func newSendDestination(l *load.Load) *destination {
 		Load: l,
 	}
 
-	dst.destinationAddressEditor = l.Theme.Editor(new(widget.Editor), "Address")
+	dst.destinationAddressEditor = l.Theme.Editor(new(widget.Editor), "Destination address")
```
dreacot:
```suggestion
	dst.destinationAddressEditor = l.Theme.Editor(new(widget.Editor), "Destination Address")
```

dreacot (changes requested):

dreacot (approved):
LGTM

Issue #816, opened 2022-03-22
Prevent duplication of wallets when importing a wallet from seeds by validating wallet seed against existing wallets.


Sirmorrison:
Currently, if i have a wallet that the seed is backed up, i can re-import that wallet into the application.

This is bad because it give a false total balance and the wallet address are same as the default and as such transactions will be duplicated.

Original wallet.
<img width="800" alt="image" src="https://user-images.githubusercontent.com/27733432/159470380-7f08a73a-9c16-4484-9c7d-d0f88b0d3a6e.png">

duplicated
<img width="795" alt="image" src="https://user-images.githubusercontent.com/27733432/159470557-6b8db7c2-15a7-46aa-89f2-30e9795fe5a4.png">

HOW TO REPRODUCE
=================
Back up and already existing wallet seed, then try importing wallet from existing seed of the just backed up wallet.

itswisdomagain:
I think this should also apply to adding watch-only wallets using xpub of an existing wallet. It'll result in similar issues as presented above, the only difference being that the watch-only wallet cannot be used to send funds.

In a similar vein, if a watch-only wallet already exists and a user tries to import a new wallet using the seed for the existing watch-only wallet, the app should probably show a warning and migrate the watch-only wallet to a full wallet rather than keeping the watch-only wallet and importing a new wallet that has the same tx history and balance as the existing watch-only wallet.

dreacot:
resolved by https://github.com/planetdecred/godcr/pull/891

PR #881, opened 2022-04-11
Fix balance disparity when wallet is imported multiple times

crux25:
This PR closes issue #816.  
There is technically no means of looking at the seeds of the wallets already registered on godcr when a new wallet is imported i.e The seeds of the existing wallets are either encrypted or cleared if verified. So no efficient mechanism to stop the import of the same wallet twice.
This PR calculate the total balance of all the wallets by summing the value of distinct unspent outputs. So even if a wallet is imported more than once its balance is only summed once.

[Dcrlibwallet PR 242](https://github.com/planetdecred/dcrlibwallet/pull/242) prevents same tx from showing up multiple times on the overview page.

### DIFF
--- a/ui/page/main_page.go
+++ b/ui/page/main_page.go
@@ -298,6 +298,42 @@ func (mp *MainPage) updateBalance() {
 	}
 }
 
+func (mp *MainPage) CalculateTotalWalletsBalance() (dcrutil.Amount, error) {
+	totalBalance := int64(0)
+	wallets := mp.WL.SortedWalletList()
+	if len(wallets) == 0 {
+		return 0, nil
+	}
+
+	var txHashMap = make(map[string]int64)
+
+	for _, wallet := range wallets {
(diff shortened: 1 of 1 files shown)


Sirmorrison:
> This PR closes issue #816. There is technically no means of looking at the seeds of the wallets already registered on godcr when a new wallet is imported i.e The seeds of the existing wallets are either encrypted or cleared if verified. So no efficient mechanism to stop the import of the same wallet twice. This PR calculate the total balance of all the wallets by summing the value of distinct unspent outputs. So even if a wallet is imported more than once its balance is only summed once.
> 
> [Dcrlibwallet PR 242](https://github.com/planetdecred/dcrlibwallet/pull/242) prevents same tx from showing up multiple times on the overview page.

The issue with the calculated balance is that they only show up after sync and at that moment, wallet has already been imported.
@itswisdomagain what do you think?

Sirmorrison:
Resolved by #891

PR #882, opened 2022-04-12
copy update transaction overview

monsa00:
minor copy update for transactions overview.


ui/values/localizable/en.go
```diff
@@ -22,7 +22,7 @@ const EN = `
 "currentTotalBalance" = "Current Total Balance";
 "walletStatus" = "Wallet Status";
 "blockHeaderFetched" = "Block header fetched";
-"noTransactionsYet" = "No transactions yet.";
+"noTransactionsYet" = "No transactions";
```
Sirmorrison:
```suggestion
"noTransactions" = "No transactions";
```
This will also need to be updated for every instance where it is used.

Sirmorrison (changes requested):

ui/values/localizable/en.go
monsa00: Ok, updated

Sirmorrison (approved):
LGTM

PR #885, opened 2022-04-13
makes wallet page remember expanded state

### LINKED ISSUE #846: Wallets page doesn't remember expanded state
From wallets page, expand a wallet, click on an account, it goes to a view of the account, use the back arrow, the wallets that were expanded are now all closed.

Vanvians:
Closes #846 

This PR makes it such that when you expand a wallet, click an account to go to the account view, and the return back to the wallets page with the back arrow; the wallets that were expanded, remain expanded.

### DIFF
--- a/ui/decredmaterial/collapsible.go
+++ b/ui/decredmaterial/collapsible.go
@@ -113,18 +113,17 @@ func (c *Collapsible) IsExpanded() bool {
 
 var rememberExpand map[int]bool
 
-func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C) D, more func(C), walletID int) layout.Dimensions {
+func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C) D, more func(C), rowID int) layout.Dimensions {
 	if rememberExpand == nil {
 		rememberExpand = make(map[int]bool)
 	}
 
-	for c.button.Clicked() {
-		c.isExpanded = !c.isExpanded
-		rememberExpand[walletID] = c.isExpanded
+	if c.button.Clicked() {
+		rememberExpand[rowID] = !rememberExpand[rowID]
 	}
 
 	icon := c.collapsedIcon
-	if rememberExpand[walletID] {
+	if rememberExpand[rowID] {
 		icon = c.expandedIcon
 	}
 
@@ -155,7 +154,7 @@ func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C) 
 				})
 			}),
 			layout.Rigid(func(gtx C) D {
-				if rememberExpand[walletID] {
+				if rememberExpand[rowID] {
 					return body(gtx)
 				}
 				return D{}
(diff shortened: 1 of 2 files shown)


ui/decredmaterial/collapsible.go
```diff
@@ -111,13 +111,20 @@ func (c *Collapsible) IsExpanded() bool {
 	return c.isExpanded
 }
 
-func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C) D, more func(C)) layout.Dimensions {
+var rememberExpand map[int]bool
+
+func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C) D, more func(C), walletID int) layout.Dimensions {
+	if rememberExpand == nil {
+		rememberExpand = make(map[int]bool)
+	}
```
dreacot, line 119: seeing as this isn't used outside this method, we can declare and initialize here
```suggestion
	rememberExpand := make(map[int]bool)
```

dreacot, line 119: seeing as this isn't used outside this method, we can declare and initialize here
```suggestion
	rememberExpand := make(map[int]bool)
```

ui/decredmaterial/collapsible.go
```diff
@@ -111,13 +111,20 @@ func (c *Collapsible) IsExpanded() bool {
 	return c.isExpanded
 }
 
-func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C) D, more func(C)) layout.Dimensions {
+var rememberExpand map[int]bool
+
+func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C) D, more func(C), walletID int) layout.Dimensions {
```
dreacot: we can replace `walletID` with `rowID` or `listID`, since the collapsible class can be used with any list other than one associated with a wallet
```suggestion
func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C) D, more func(C), rowID int) layout.Dimensions {
```

dreacot: we can replace `walletID` with `rowID` or `listID`, since the collapsible class can be used with any list other than one associated with a wallet
```suggestion
func (c *CollapsibleWithOption) Layout(gtx layout.Context, header, body func(C) D, more func(C), rowID int) layout.Dimensions {
```

ui/decredmaterial/collapsible.go
dreacot: or does it override the initial saved map? each time the layout is drawn?

dreacot (changes requested):
when a wallet row is expanded, leave the page, and then return to the wallets page, clicking on a row to collapse it does nothing on the first click, you have to click a row twice before the row get's collapsed.

clicking every other time works as intended

ui/decredmaterial/collapsible.go
Vanvians: it needs to be in an if statement, if not the map resets when you go back to the wallets page. We need it to remember the values when you leave the page.

Sirmorrison (approved):

Sirmorrison:
LGTM @dreacot kindly review.

dreacot (approved):

PR #891, opened 2022-04-21
Prevent duplicate wallets.

crux25:
This PR prevents users from importing wallets using the seed of an already existing wallet. Also checks the xpub of a watch-only wallet a user is about to import against the xpub of existing wallets, it warns and prevents the import if a match is detected. Summary: It prevents duplication of wallets. Closes issue #816  and PR #881. Depends on https://github.com/planetdecred/dcrlibwallet/pull/244

### DIFF
--- a/ui/page/wallets/restore_page.go
+++ b/ui/page/wallets/restore_page.go
@@ -441,15 +441,15 @@ func (pg *Restore) verifySeeds() bool {
 		}
 	}
 
-	// Compare seed with existing wallets seed. On positive match abort,
-	// import to prevent duplicate wallet. seedMatch == walled.ID if there is a match.
-	seedMatch, err := pg.WL.MultiWallet.WalletWithSeed(pg.seedPhrase)
+	// Compare seed with existing wallets seed. On positive match abort import
+	// to prevent duplicate wallet. walletWithSameSeed >= 0 if there is a match.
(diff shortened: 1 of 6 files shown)


ui/page/wallets/restore_page.go
```diff
@@ -417,6 +417,19 @@ func (pg *Restore) validateSeeds() bool {
 		return false
 	}
 
+	// Compare seed with existing wallets seed. On positive match abort,
+	// import to prevent duplicate wallet. seedMatch == walled.ID if there is a match.
```
itswisdomagain:
```suggestion
	// Compare seed with existing wallets seed. On positive match abort import
	// to prevent duplicate wallet. seedMatch >= 0 if there is a match.
```

itswisdomagain: I'm not so sure about the `seedMatch` variable name though.

crux25: Resolved.

Sirmorrison (changes requested):
This looks good for wallet import but not same for watch only wallets.

The watch only wallet fails to validate for everytime i try to import the wallet.

https://user-images.githubusercontent.com/27733432/165051212-cc13ebcf-6f3c-47c1-81b8-b99edd37284f.mov

Sirmorrison (approved):

PR #892, opened 2022-04-21
Skip Set up needed accounts page when there is only the default account.

### LINKED ISSUE #853: Skip "Set up needed accounts" option screen for privacy if no additional accounts detected in wallet
Privacy setup currently allows the user to manually setup accounts even if the wallet has no additional accounts to map. This should be auto detected (the lack of accounts) and can infer the user has to setup new accounts if none exist.

godcr-issue

crux25:
This PR Skips the setup needed accounts page when setting up Stake-Shuffle, for cases where additional accounts are not detected. Closes #853 .

### DIFF
--- a/ui/page/privacy/setup_privacy_page.go
+++ b/ui/page/privacy/setup_privacy_page.go
@@ -5,7 +5,6 @@ import (
 
 	"gioui.org/layout"
 	"gioui.org/text"
-	"gioui.org/widget"
 
 	"github.com/planetdecred/dcrlibwallet"
 	"github.com/planetdecred/godcr/ui/decredmaterial"
@@ -13,7 +12,6 @@ import (
 	"github.com/planetdecred/godcr/ui/modal"
 	"github.com/planetdecred/godcr/ui/page/components"
 	"github.com/planetdecred/godcr/ui/values"
-	"golang.org/x/exp/shiny/materialdesign/icons"
 )
 
 const SetupPrivacyPageID = "SetupPrivacy"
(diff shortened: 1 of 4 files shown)


ui/page/privacy/setup_privacy_page.go
```diff
@@ -151,7 +151,19 @@ func (pg *SetupPrivacyPage) privacyIntroLayout(gtx layout.Context) layout.Dimens
 // Part of the load.Page interface.
 func (pg *SetupPrivacyPage) HandleUserInteractions() {
 	if pg.toPrivacySetup.Clicked() {
-		pg.ChangeFragment(NewSetupMixerAccountsPage(pg.Load, pg.wallet))
+		accounts, err := pg.wallet.GetAccountsRaw()
+		if err != nil {
+			log.Error(err)
+		}
+		// We are using 2 here because of imported wallet.
+		if accounts.Count <= 2 {
```
Sirmorrison: I believe it will be more health to check what the 2 wallets are. 

```suggestion
		if accounts.Count <= 2 && accounts.Acc[n] == "imported" {
```
this is just a random code suggestion.

Sirmorrison (commented):
PR looks good.
Just this comment i think we can discuss.

ui/page/privacy/setup_privacy_page.go
crux25: I didn't want to waste the extra computational resources looping through accounts slice, since the Count element already hold the len of the slice. So unless there is a way to delete the "imported" wallet, accounts.Count should always be => 2.

Sirmorrison (approved):

PR #893, opened 2022-04-21
update transaction row information

### LINKED ISSUE #868: update staking transaction to match the mobile wallets
There seems to be some discrepancies in the way tickets are shown on godcr as compared to the Decred wallet mobile apps and Decrediton.

1.
This is a vote transaction on godcr
Screenshot from 2022-03-31 22-44-58

Clicking on "View associated ticket", produces the page below
Screenshot from 2022-03-31 22-45-58

2. 
While this is a vote transaction on mobile

And clicking on "View associated ticket" produces the screen below

3.
The Transactions showing as "Voted" on the list, are actually just purchased tickets
Screenshot from 2022-03-31 22-50-58

They should be labelled "Purchased" as seen here

4.
This says "Yourself"
Screenshot from 2022-03-31 23-18-15

It should say "Transferred"

5.
There's a negative reward for revocation, this doesn't indicate that.
Screenshot from 2022-03-31 23-20-40
Screenshot from 2022-03-31 23-20-33

Also it shouldn't say "Revocation" but should say "Revoked"

6.
Clicking on "View associated ticket" for any transaction should always lead to a page titled "Ticket Purchase" that looks like this (including the icon)

7. 
Group similar outputs from mixed transactions

8.
Add category count to dropdown and change "Yourself" to "Mixed"

dreacot:
Resolves #868

**Screenshots**
Screenshot from 2022-05-18 18-44-08
Screenshot from 2022-05-18 18-44-43
Screenshot from 2022-05-18 18-45-13

### DIFF
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -474,7 +474,7 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, row TransactionRow) 
 						}),
 						layout.Rigid(func(gtx C) D {
 							// vote reward
-							if row.Transaction.Type != dcrlibwallet.TxTypeVote {
+							if row.Transaction.Type != dcrlibwallet.TxTypeVote && row.Transaction.Type != dcrlibwallet.TxTypeRevocation{
 								return D{}
 							}
 
@@ -486,11 +486,6 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, row TransactionRow) 
 								Alignment:   layout.Middle,
 							}.Layout(gtx,
 								layout.Rigid(func(gtx C) D {
-									label := l.Theme.Label(values.TextSize14, "+")
-									label.Color = l.Theme.Color.Turquoise800
-									return label.Layout(gtx)
-								}),
-								layout.Rigid(func(gtx C) D {
 									ic := l.Theme.Icons.DecredSymbol2
 
 									return layout.Inset{
@@ -500,7 +495,10 @@ func LayoutTransactionRow(gtx layout.Context, l *load.Load, row TransactionRow) 
 								}),
 								layout.Rigid(func(gtx C) D {
 									label := l.Theme.Label(values.TextSize12, dcrutil.Amount(row.Transaction.VoteReward).String())
-									label.Color = l.Theme.Color.Turquoise800
+									label.Color = l.Theme.Color.Orange
+									if row.Transaction.Type == dcrlibwallet.TxTypeVote {
+										label.Color = l.Theme.Color.Turquoise800
+									}
 									return label.Layout(gtx)
 								}),
 							)
--- a/ui/page/components/components.go
+++ b/ui/page/components/components.go
@@ -131,13 +131,13 @@ func TransactionTitleIcon(l *load.Load, wal *dcrlibwallet.Wallet, tx *dcrlibwall
 
 	switch tx.Direction {
 	case dcrlibwallet.TxDirectionSent:
-		txStatus.Title = "Sent"
+		txStatus.Title = values.String(values.StrSent)
 		txStatus.Icon = l.Theme.Icons.SendIcon
 	case dcrlibwallet.TxDirectionReceived:
-		txStatus.Title = "Received"
+		txStatus.Title = values.String(values.StrReceived)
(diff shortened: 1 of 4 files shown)


ui/page/components/components.go
```diff
@@ -111,82 +111,93 @@ func UniformHorizontalPadding(gtx layout.Context, body layout.Widget) layout.Dim
...
+					txStatus.TicketStatus = dcrlibwallet.TicketStatusLive
+					txStatus.ProgressBarColor = l.Theme.Color.Primary
+					txStatus.ProgressTrackColor = l.Theme.Color.LightBlue4
+					txStatus.Background = l.Theme.Color.Primary50
+				} else if wal.TxMatchesFilter(tx, dcrlibwallet.TxFilterExpired) {
+					txStatus.Title = "Expired"
+					txStatus.Icon = l.Theme.Icons.TicketExpiredIcon
+					txStatus.Color = l.Theme.Color.GrayText2
+					txStatus.TicketStatus = dcrlibwallet.TicketStatusExpired
+					txStatus.Background = l.Theme.Color.Gray4
```
Sirmorrison:
```suggestion
	case dcrlibwallet.TxDirectionSent:
		txStatus.Title = values.String(values.StrSent)
		txStatus.Icon = l.Theme.Icons.SendIcon
	case dcrlibwallet.TxDirectionReceived:
		txStatus.Title = values.String(values.StrReceived)
		
```

Lets localize all string variables as they where initially.

Sirmorrison (changes requested):

ui/page/components/components.go
```diff
@@ -111,82 +111,93 @@ func UniformHorizontalPadding(gtx layout.Context, body layout.Widget) layout.Dim
...
+							txStatus.Title = values.String(values.StrTicketVotedTitle)
+							txStatus.Icon = l.Theme.Icons.TicketLiveIcon
+							txStatus.Color = l.Theme.Color.Turquoise700
+							txStatus.TicketStatus = dcrlibwallet.TicketStatusVotedOrRevoked
+							txStatus.ProgressBarColor = l.Theme.Color.Turquoise300
+							txStatus.ProgressTrackColor = l.Theme.Color.Turquoise100
+							txStatus.Background = l.Theme.Color.Success2
+						} else {
+							txStatus.Title = values.String(values.StrTicketRevokedTitle)
+							txStatus.Icon = l.Theme.Icons.TicketLiveIcon
```
Sirmorrison, line 189: Live and revoked tickets, have similar icons... is this intentional?
image

Sirmorrison: this is seen on the staking page.

dreacot: Screenshot from 2022-06-24 11-18-20

Decrediton uses same icon for tickets whether revoked or voted, so i did same.

Personally i feel we need to have the designers run through the tickets live cycle and provide is unique icons for all, plus i don't think the same method that populate the transactions list, should also populate the ticket list. there are overlaps.

Initially i was making the change of separating the methods in this PR, but later determined it would be best suited in another PR and properly detailed

Sirmorrison: So if i understand correctly, the current icon for revoked tickets does not apply here then.

I feel regardless, we should maintain out icons for revoked as it was pending when we get designs for this by the design team. Except you have some concerns about it so we look into it objectively.

About the different methods for transaction and tickets, could you help expand on that so we know what we are up against and to help determine the best approach. it will also help us properly prioritize the recommended task.

dreacot: No the icon is only reflected on a revocation transaction. but all tickets have same icon

Right now, there's no standard icon in place, i mean if we want to we can revert it, but I think this works for now, i could probably work on this aspect for the v2 UI
Screenshot from 2022-06-29 20-39-26

so for godcr the icons are only for revoked/voted transactions but not tickets, i think the idea was to make tickets have a distinguishable icon


we should have 2 methods `renderTransactions` and `renderTicket`, naming is just for example purposes,  also a different method populating the ticket/transaction details page. just to give an example off the top of my head, the color change on the tickets list was intended for the transaction list only. also using the same methods doesn't allow for the transaction list, the ticket list and their details page to have different titles

Sirmorrison (approved):

PR #898, opened 2022-04-22
Don't allow same account for mixed and unmixed account for manual cssp set up.

### LINKED ISSUE #856: UI should not allow same account to be selected for mixed and unmixed in cspp manual setup.
UI should not allow same account to be selected for mixed and unmixed in cspp manual setup. Currently an error triggers after the user clicks 'Set up' but it's cleaner UI experience to disable the button instead.

godcr-sameacct

crux25:
This PR closes #856. It disables the set-up button on the cssp manual setup page if the unmixed and mixed account selector is the same account. It also sets different accounts for mixed/unmixed accounts when the manual setup page is navigated to.

### DIFF
--- a/ui/page/components/account_selector.go
+++ b/ui/page/components/account_selector.go
@@ -94,9 +94,9 @@ func (as *AccountSelector) Handle() {
 }
 
 // SelectFirstWalletValidAccount selects the first valid account from the
-// first wallet in the SortedWalletList. If selectedWallet is not nil,
-// the first account for the selectWallet is selected.
-func (as *AccountSelector) SelectFirstWalletValidAccount(selectedWallet *dcrlibwallet.Wallet) error {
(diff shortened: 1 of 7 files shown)


ui/page/components/account_selector.go
```diff
@@ -138,6 +138,28 @@ func (as *AccountSelector) SelectFirstWalletValidAccount(selectedWallet *dcrlibw
...
+			if as.accountIsValid(account) && account.Number != accountID {
+				as.SetSelectedAccount(account)
+				as.callback(account)
+				return nil
+			}
+		}
+	}
+
+	return errors.New("no valid account found")
+}
```
Sirmorrison: The entire logic here can fit into `SelectFirstWalletValidAccount()`. This is code duplication and not optimal.

Sirmorrison (changes requested):

Sirmorrison (approved):

PR #899, opened 2022-04-23
Get recent address when Receive page is navigated to

### LINKED ISSUE #863: privacy of wallet affected as receive page doesn't update the wallet's current address
for an address to be displayed on the receive page a call is made to the `wallet.CurrentAddress()` method.
The upstream comment on that method sates explicitly that;
```
// CurrentAddress gets the most recently requested payment address from a wallet.
// If the address has already been used (there is at least one transaction
// spending to it in the blockchain or dcrd mempool), the next chained address
// is returned.
func (w *Wallet) CurrentAddress(account uint32) (stdaddr.Address, error) {
```

currently on godcr, once the receive page is opened, an instance of the receive page is maintained, meaning that the said method on https://github.com/planetdecred/godcr/blob/master/ui/page/receive_page.go#L101, only gets called once on the first launch of the receive page, and not on other visits to the receive page.

This hinders `wallet.CurrentAddress()` from performing as intended, and therefore affecting the privacy of the wallet(s) involved.

**How to Replicate:**
1. Open the receive page and copy the address shown.
2. Send funds to that address and wait for the funds to be confirmed. (2 confirmations)
3. Revisit the receive page, you'd notice you're still presented with the same initial address. You should be presented with a new address at this point.

A possible solution to this is to call the `wallet.CurrentAddress()` method each time the receive page is navigated to, or to create a new instance of the receive page on each visit

devchoplife:
Resolves #863 
This PR implements the following:
- Prevent access to send and receive pages until sync is complete. This is to prevent an unsynced account from calling the function that checks if an address has been used and returns a new one. Wallets that are not synced to the tip would not have enough information to get the last transaction 
- Generate a new instance of receive page whenever it is navigated to. This is to ensure that transaction addresses are not reused which improves wallet privacy 
- Add notification messages to let users know that sync needs to be completed before send and receive pages can be accessed

### DIFF
--- a/ui/page/main_page.go
+++ b/ui/page/main_page.go
@@ -400,7 +400,7 @@ func (mp *MainPage) HandleUserInteractions() {
 				if mp.receivePage == nil {
 					mp.receivePage = NewReceivePage(mp.Load)
 				}
-				pg = NewReceivePage(mp.Load)
+				pg = mp.receivePage
 			}
 
 			if pg.ID() == mp.currentPageID() {
--- a/ui/page/main_page.go
+++ b/ui/page/main_page.go
@@ -408,9 +408,11 @@ func (mp *MainPage) HandleUserInteractions() {
 			}
 
 			if !mp.WL.MultiWallet.IsSynced() {
-				mp.Toast.Notify("Wallet not synced")
-			} else {
 				mp.ChangeFragment(pg)
+			} else if mp.WL.MultiWallet.IsSyncing() {
+				mp.Toast.Notify("Wallet is syncing, please wait")
+			} else {
+				mp.Toast.Notify("Wallet not synced")
 			}
 		}
 	}
--- a/ui/page/main_page.go
+++ b/ui/page/main_page.go
@@ -407,7 +407,13 @@ func (mp *MainPage) HandleUserInteractions() {
 				continue
 			}
 
-			mp.ChangeFragment(pg)
+			if !mp.WL.MultiWallet.IsConnectedToDecredNetwork() {
+				mp.Toast.Notify("Connect to the decred network and wait for sync to complete")
+			} else if mp.WL.MultiWallet.IsSyncing() {
+				mp.Toast.Notify("Wait for sync to complete")
+			} else {
+				mp.ChangeFragment(pg)
+			}
 		}
 	}
 
--- a/ui/page/main_page.go
+++ b/ui/page/main_page.go
@@ -410,9 +410,9 @@ func (mp *MainPage) HandleUserInteractions() {
 			if mp.WL.MultiWallet.IsSynced() {
 				mp.ChangeFragment(pg)
 			} else if mp.WL.MultiWallet.IsSyncing() {
-				mp.Toast.Notify("Wallet is syncing, please wait")
+				mp.Toast.NotifyError("Wallet is syncing, please wait")
 			} else {
-				mp.Toast.Notify("Not connected to the Decred network")
+				mp.Toast.NotifyError("Not connected to the Decred network")
 			}
 		}
 	}
--- a/ui/page/main_page.go
+++ b/ui/page/main_page.go
@@ -407,10 +407,8 @@ func (mp *MainPage) HandleUserInteractions() {
 				continue
 			}
 
-			if !mp.WL.MultiWallet.IsConnectedToDecredNetwork() {
-				mp.Toast.Notify("Connect to the decred network and wait for sync to complete")
-			} else if mp.WL.MultiWallet.IsSyncing() {
-				mp.Toast.Notify("Wait for sync to complete")
+			if !mp.WL.MultiWallet.IsSynced() {
+				mp.Toast.Notify("Wallet not synced")
 			} else {
 				mp.ChangeFragment(pg)
 			}
--- a/ui/page/main_page.go
+++ b/ui/page/main_page.go
@@ -407,12 +407,12 @@ func (mp *MainPage) HandleUserInteractions() {
 				continue
 			}
 
-			if !mp.WL.MultiWallet.IsSynced() {
+			if mp.WL.MultiWallet.IsSynced() {
 				mp.ChangeFragment(pg)
 			} else if mp.WL.MultiWallet.IsSyncing() {
 				mp.Toast.Notify("Wallet is syncing, please wait")
 			} else {
-				mp.Toast.Notify("Wallet not synced")
+				mp.Toast.Notify("Not connected to the Decred network")
 			}
 		}
 	}
--- a/ui/page/main_page.go
+++ b/ui/page/main_page.go
@@ -400,7 +400,7 @@ func (mp *MainPage) HandleUserInteractions() {
 				if mp.receivePage == nil {
 					mp.receivePage = NewReceivePage(mp.Load)
 				}
-				pg = mp.receivePage
+				pg = NewReceivePage(mp.Load)
 			}
 
 			if pg.ID() == mp.currentPageID() {
--- a/ui/page/receive_page.go
+++ b/ui/page/receive_page.go
@@ -3,6 +3,7 @@ package page
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"image"
 	"image/color"
 	"time"
@@ -141,6 +142,7 @@ func (pg *ReceivePage) OnNavigatedTo() {
 	currentAddress, err := selectedWallet.CurrentAddress(pg.selector.SelectedAccount().Number)
 	if err != nil {
 		log.Errorf("Error getting current address: %v", err)
+		pg.Toast.NotifyError(fmt.Sprintf("Error getting current address: %v", err))
 	} else {
 		pg.currentAddress = currentAddress
 	}
--- a/ui/page/receive_page.go
+++ b/ui/page/receive_page.go
@@ -145,9 +145,8 @@ func (pg *ReceivePage) OnNavigatedTo() {
 		pg.Toast.NotifyError(fmt.Sprintf("Error getting current address: %v", err))
 	} else {
 		pg.currentAddress = currentAddress
+		pg.generateQRForAddress()
 	}
-
-	pg.generateQRForAddress()
 }
 
 func (pg *ReceivePage) generateQRForAddress() {
--- a/ui/page/receive_page.go
+++ b/ui/page/receive_page.go
@@ -137,6 +137,15 @@ func (pg *ReceivePage) OnNavigatedTo() {
 	pg.selector.ListenForTxNotifications(pg.ctx)
 	pg.selector.SelectFirstWalletValidAccount(nil) // Want to reset the user's selection everytime this page appears?
 	// might be better to track the last selection in a variable and reselect it.
+	selectedWallet := pg.multiWallet.WalletWithID(pg.selector.SelectedAccount().WalletID)
+	currentAddress, err := selectedWallet.CurrentAddress(pg.selector.SelectedAccount().Number)
+	if err != nil {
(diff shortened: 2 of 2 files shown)


crux25 (changes requested):
Moved comment to relevant code section.

crux25 (changes requested):
Comment moved to relevant code section.

ui/page/main_page.go
```diff
@@ -400,14 +400,20 @@ func (mp *MainPage) HandleUserInteractions() {
 				if mp.receivePage == nil {
 					mp.receivePage = NewReceivePage(mp.Load)
 				}
-				pg = mp.receivePage
+				pg = NewReceivePage(mp.Load)
```
crux25: Generating a new instance of the page every time a user visits the receive page is going to produce an unpredictable back-stack. Also if a user should navigate back to the previously generated page on the back-stack they'll have the old receiving address which sort of defeats your purpose.
What you should do is withing the receiving page `OnNavigatedTo()` retrieve the current selected wallet and call `wallet.CurrentAddress()` on the wallet, then assign the result to the page's `currentAddress`.

ui/page/main_page.go
```diff
@@ -400,14 +400,20 @@ func (mp *MainPage) HandleUserInteractions() {
...
 			}
 
 			if pg.ID() == mp.currentPageID() {
 				continue
 			}
 
-			mp.ChangeFragment(pg)
+			if !mp.WL.MultiWallet.IsConnectedToDecredNetwork() {
```
crux25: Implement this logic on the individual pages and not here
Currently send page doesn't permit sending when godcr isn't connected to the network. So no point implementing that logic.

I will suggest that the receive page should let users generate address after account discovery for imported accounts and at any point for other accounts created on the app.

crux25 (changes requested):

ui/page/main_page.go
devchoplife: This raises a privacy concern in the sense that addresses might be reused when a wallet is not synced to the tip as the information available is not sufficient to determine if an address has been used before.

crux25: Should be best to balance security and usability. If a user tries to retrieve an address when the wallet isn't fully synced the user should get a notification about the security implications, but should be able to retrieve an address irrespective. In some cases users will be more concerned about not waiting to sync to get a receiving address than reusing an address.

Sirmorrison: The receive page should not be accessible until wallets are full my synced… you can test the mobile android or iOS for more information

ui/page/main_page.go
```diff
@@ -407,7 +407,13 @@ func (mp *MainPage) HandleUserInteractions() {
...
+				mp.Toast.Notify("Connect to the decred network and wait for sync to complete")
```
```suggestion
			if !mp.WL.MultiWallet.IsSynced() {
				mp.Toast.Notify("Wallet not synced")
```

ui/page/main_page.go
dreacot: For the first point you made, if the send page isn't usuable, then might as well prevent them from accessing it and letting them know why, seams like a cleaner approach

Then, while I get your point that it could be frustrating to wait for sync before you can get a receiving address, even if you get an address before sync completes, sending funds there wouldn't reflect until sync is done, so they might as well wait

ui/page/receive_page.go
```diff
@@ -137,6 +137,15 @@ func (pg *ReceivePage) OnNavigatedTo() {
 	pg.selector.ListenForTxNotifications(pg.ctx)
 	pg.selector.SelectFirstWalletValidAccount(nil) // Want to reset the user's selection everytime this page appears?
 	// might be better to track the last selection in a variable and reselect it.
+	selectedWallet := pg.multiWallet.WalletWithID(pg.selector.SelectedAccount().WalletID)
+	currentAddress, err := selectedWallet.CurrentAddress(pg.selector.SelectedAccount().Number)
+	if err != nil {
+		log.Errorf("Error getting current address: %v", err)
+	} else {
```
Sirmorrison: Why are we only logging the error instead of returning it?

Sirmorrison (changes requested):

ui/page/main_page.go
```diff
@@ -407,7 +407,13 @@ func (mp *MainPage) HandleUserInteractions() {
 				continue
 			}
 
-			mp.ChangeFragment(pg)
+			if !mp.WL.MultiWallet.IsSynced() {
+				mp.ChangeFragment(pg)
+			} else if mp.WL.MultiWallet.IsSyncing() {
+				mp.Toast.Notify("Wallet is syncing, please wait")
+			} else {
+				mp.Toast.Notify("Wallet not synced")
```
dreacot:
```suggestion
				mp.Toast.Notify("Not connected to Decred network")
```

ui/page/receive_page.go
```diff
@@ -137,6 +138,16 @@ func (pg *ReceivePage) OnNavigatedTo() {
...
 	pg.selector.SelectFirstWalletValidAccount(nil) // Want to reset the user's selection everytime this page appears?
 	// might be better to track the last selection in a variable and reselect it.
+	selectedWallet := pg.multiWallet.WalletWithID(pg.selector.SelectedAccount().WalletID)
+	currentAddress, err := selectedWallet.CurrentAddress(pg.selector.SelectedAccount().Number)
+	if err != nil {
+		log.Errorf("Error getting current address: %v", err)
+		pg.Toast.NotifyError(fmt.Sprintf("Error getting current address: %v", err))
+	} else {
+		pg.currentAddress = currentAddress
+	}
```
dreacot: In regards displaying the error to the user, I'm not sure it's required, as this is not a user generated action and can be confusing, seems better to just log the response only

errors are usually useful when the user has tried to perform an action, and they are told the result of that action, or directed to what needs to be done, telling the user there's an error getting the current address is confusing as they still see an address and are left perplexed

```suggestion
	if err != nil {
		log.Errorf("Error getting current address: %v", err)
	} else {
		pg.currentAddress = currentAddress
	}
```

ui/page/receive_page.go
```diff
@@ -137,6 +138,16 @@ func (pg *ReceivePage) OnNavigatedTo() {
...
+
+	pg.generateQRForAddress()
```
dreacot: although not related to your PR it makes sense to modify this method to accept an address as an argument, in which it would be used to generate the QR code, doesn't make much sense to update the address separately and generate the QR code separately when they are directly related

or to avoid feature creeping, an issue can be created and it can be done in another PR

dreacot (dismissed):

ui/page/receive_page.go
dreacot: @Sirmorrison @itswisdomagain thoughts?

ui/page/main_page.go
Sirmorrison: My concern with this is what happens after you are connected, and you try to access the page?

dreacot: `if !mp.WL.MultiWallet.IsSynced() {
				mp.ChangeFragment(pg)`

It goes to the page like on the code, my suggestion was just to change the text, from "wallet not synced"

Sirmorrison: alright.. lets get this implemented and we move from there.

ui/page/receive_page.go
Sirmorrison: My take on errors is basically different. it provides the user with facts that an error has occurred while trying out a process rather than leave them in limbo hoping for result. 

looking at the code, it is an `if` `else` scenario meaning this should happen or this and as such we are expecting 2 outcomes. when one fails, user need to be notified so they dont keep waiting for the output of the other. 

displaying the error would not really matter if it does not have impact on the UX, however in this regard it does.

Address will not the generated and the user will not know why and as such, nothing will be displayed on the UI for the user to work with.

https://github.com/planetdecred/godcr/blob/f31dfb8d0d23bad0aee029985177b3e9a4624092/ui/page/receive_page.go#L208

ui/page/receive_page.go
Sirmorrison: I think this approach works for me though. Its really not too much of an issue and the address are used from different methods.

ui/page/receive_page.go
dreacot: alright good point

ui/page/receive_page.go
```diff
@@ -105,7 +106,7 @@ func NewReceivePage(l *load.Load) *ReceivePage {
 				pg.currentAddress = currentAddress
 			}
 
-			pg.generateQRForAddress()
+			pg.generateQRForAddress(currentAddress)
```
Sirmorrison: Please revert this... 
Currently we already have a `pg.currentAddress` i dont see the need to pass parameters around.

`pg,currentAddress` is set before `pg.generateQRForAddress()` is called so no need for he extra `currentAddress` param.

Although i see we may have some challenge here though especially when there is an error with `currentAddress`.

Kindly move `pg.generateQRForAddress()`into the `if else` statement. That way, we only try to generate the QrCode when there is an address available.

ui/page/receive_page.go
```diff
@@ -355,7 +366,7 @@ func (pg *ReceivePage) HandleUserInteractions() {
 		}
 
 		pg.currentAddress = newAddr
-		pg.generateQRForAddress()
+		pg.generateQRForAddress(newAddr)
```
Sirmorrison: revert

ui/page/receive_page.go
```diff
@@ -137,10 +138,20 @@ func (pg *ReceivePage) OnNavigatedTo() {
...
+	} else {
+		pg.currentAddress = currentAddress
+	}
+
+	pg.generateQRForAddress(currentAddress)
 }
 
-func (pg *ReceivePage) generateQRForAddress() {
-	qrCode, err := qrcode.New(pg.currentAddress)
+func (pg *ReceivePage) generateQRForAddress(walletAddress string) {
```
Sirmorrison: revert

ui/page/receive_page.go
```diff
@@ -137,10 +138,20 @@ func (pg *ReceivePage) OnNavigatedTo() {
...
+	selectedWallet := pg.multiWallet.WalletWithID(pg.selector.SelectedAccount().WalletID)
+	currentAddress, err := selectedWallet.CurrentAddress(pg.selector.SelectedAccount().Number)
+	if err != nil {
+		log.Errorf("Error getting current address: %v", err)
+		pg.Toast.NotifyError(fmt.Sprintf("Error getting current address: %v", err))
+	} else {
+		pg.currentAddress = currentAddress
+	}
+
+	pg.generateQRForAddress(currentAddress)
```
Sirmorrison:
```suggestion
	} else {
		pg.currentAddress = currentAddress
		pg.generateQRForAddress()
	}
```

Sirmorrison (changes requested):

ui/page/main_page.go
```diff
@@ -407,7 +407,13 @@ func (mp *MainPage) HandleUserInteractions() {
 				continue
 			}
 
-			mp.ChangeFragment(pg)
+			if mp.WL.MultiWallet.IsSynced() {
+				mp.ChangeFragment(pg)
+			} else if mp.WL.MultiWallet.IsSyncing() {
+				mp.Toast.Notify("Wallet is syncing, please wait")
+			} else {
+				mp.Toast.Notify("Not connected to the Decred network")
```
```suggestion
				mp.Toast.NotifyError("Wallet is syncing, please wait")
			} else {
				mp.Toast.NotifyError("Not connected to the Decred network")
```

Sirmorrison (changes requested):
Functionality looks good just whats left and we are good to go

Sirmorrison (approved):

PR #904, opened 2022-04-26
don't use default account during stake shuffle.

### LINKED ISSUE #903: Remove 'default' as a possible choice in the manual privacy setup
the `default` account should not be shown as a possible choice for the mixed and unmixed account setup.

Screen Shot 2022-04-25 at 11 29 02 AM

crux25:
This PR closes #903, specifically:
1. It doesn't count the default account when determining if manual/auto-setup should be used.
2. When on manual set up page it doesn't prefill default account.
3. It disables set up button if the default account is set for either the mixed or unmixed account.

### DIFF
--- a/ui/page/components/account_selector.go
+++ b/ui/page/components/account_selector.go
@@ -94,9 +94,9 @@ func (as *AccountSelector) Handle() {
 }
 
 // SelectFirstWalletValidAccount selects the first valid account from the
-// first wallet in the SortedWalletList, skipping any account whose number is in filterAccounts slice.
+// first wallet in the SortedWalletList
 // If selectedWallet is not nil, the first account for the selectWallet is selected.
-func (as *AccountSelector) SelectFirstWalletValidAccount(selectedWallet *dcrlibwallet.Wallet, filterAccounts []int32) error {
+func (as *AccountSelector) SelectFirstWalletValidAccount(selectedWallet *dcrlibwallet.Wallet) error {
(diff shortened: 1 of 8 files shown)


ui/page/components/account_selector.go
```diff
@@ -94,9 +94,9 @@ func (as *AccountSelector) Handle() {
 }
 
 // SelectFirstWalletValidAccount selects the first valid account from the
-// first wallet in the SortedWalletList, skipping any account whose ID is == filter.
+// first wallet in the SortedWalletList, skipping any account whose number is in filterAccounts slice.
```
Sirmorrison: From your comment, it say skip any account whose ID is found... I believe this should mea n not displayed among options or can be selected.

I can however still see and select the default wallet.
I think the best solution is not to have it among the options to be selected. this reduces chances of errors.

image

i see that the set up button is greyed out yeah.. lets just give the users the available options only.

Sirmorrison: From your comment, it say skip any account whose ID is found... I believe this should mea n not displayed among options or can be selected.

I can however still see and select the default wallet.
I think the best solution is not to have it among the options to be selected. this reduces chances of errors.

image

i see that the set up button is greyed out yeah.. lets just give the users the available options only.

Sirmorrison (changes requested):

Sirmorrison (approved):

PR #907, opened 2022-04-28
rework key event handling

itswisdomagain:
This is part of a larger body of work to refactor the codebase.

The current use of channels to notify pages and modals of key events
is inefficient because to work effectively, pages and modals will need to
start goroutines to intercept key events as soon as it is sent into the
channel.

Most of the uses of the event channel do not currently constantly check if
the event channel has received a new event, only checking the channel when
`HandleUserInteractions` is called and that is only called when some other
action causes a `FrameEvent` to be generated. Inefficient, and was responsible
for a bug on the info modal that was recently fixed using a goroutine.

To work effectively, pages and modals should be able to identify key events as
they happen, not only when some other action (like moving the mouse or clicking
a button) triggers a FrameEvent. This solution achieves this aim without using
goroutines or channels.

### DIFF
--- a/ui/decredmaterial/theme.go
+++ b/ui/decredmaterial/theme.go
@@ -287,32 +287,28 @@ func HandleEditorEvents(editors ...*widget.Editor) (bool, bool) {
 	return submit, changed
 }
 
-func SwitchEditors(keyEvent chan *key.Event, editors ...*widget.Editor) {
-	select {
-	case event := <-keyEvent:
-		if event.Name == key.NameTab && event.Modifiers != key.ModShift && event.State == key.Press {
-			for i := 0; i < len(editors); i++ {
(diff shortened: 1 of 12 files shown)


Sirmorrison (changes requested):
Nice work done so far.

Some concern i have.
When the modal is opened, the editors in background are still active. they should be disabled when the modal is opened. 

Active screen should only be the modal anythings else should be disabled.
ezgif com-gif-maker(16)

itswisdomagain:
> Active screen should only be the modal anythings else should be disabled.

I've disabled key event handling on the main page if a modal is displayed. If the desire is to disable the entire page when a modal is displayed, a new issue should be open and that resolved in a separate PR.

Sirmorrison (approved):
LGTM

PR #922, opened 2022-05-10
Multi: Fix total transaction display bug, code clean up

### LINKED ISSUE #915: Incorrect sum of transactions on the statistic page
On the statistics page the transactions count is always 0.

transactions
Transactions Page

stats-page
Stats Page

Sirmorrison:
Fix #915
- Fix total transaction display bug
- remove redundant codes and cleaned up multiple files

image


wallet/commands.go
```diff
@@ -73,76 +16,112 @@ func (wal *Wallet) HaveAddress(address string) (bool, string) {
...
-	timeLeft := time.Duration(r) * time.Second
-	if q > 0 {
-		return fmt.Sprintf("%dd%s", q, timeLeft.String())
-	}
-	return timeLeft.String()
-}
+// RescanBlocks rescans the multiwallet
+// func (wal *Wallet) RescanBlocks(walletID int) error {
+// 	return wal.multi.RescanBlocks(walletID)
+// }
```
dreacot: we have a lot of comment out codes, let's remove them, if the work can't be done in this PR, then it might be worth using what we have now, the comments are too much to ignore

ui/page/components/components.go
```diff
@@ -744,3 +744,20 @@ func CalculateTotalWalletsBalance(l *load.Load) (dcrutil.Amount, dcrutil.Amount,
...
+	q, r := divMod(totalTimeLeft, 24*60*60)
+	timeLeft := time.Duration(r) * time.Second
+	if q > 0 {
+		return fmt.Sprintf("%dd%s", q, timeLeft.String())
+	}
+	return timeLeft.String()
+}
+
+// divMod divides a numerator by a denominator and returns its quotient and remainder.
+func divMod(numerator, denominator int64) (quotient, remainder int64) {
```
dreacot, line 759: this function name could be better, the comment is noted, but it's best to make function names as readable as possible.
`numeratorByDenominator`, or something more readable

dreacot (changes requested):

wallet/commands.go
Sirmorrison: Would remove the entire file in subsequent code clean up… I feel the file is has served it’s use but it won’t be in this PR. Left the comments for reference just in case.

ui/page/components/components.go
Sirmorrison: Subsequent code clean up as well. 
trying to minimize the changes made on this Pr.

wallet/commands.go
dreacot: noted, don't think the reference is needed though, it just makes the code untidy, anyone willing to make reference can probably backtrack to commit messages, or the PR the change was made

ui/page/components/components.go
dreacot: understood regarding the further code cleanup, i believe readability in this case is worth the changes

dreacot (approved):

PR #932, opened 2022-05-18
Don't clear mixer config when spend from unmixed accounts is enabled.

### LINKED ISSUE #926: Prevent allow spending from unmixed account from clearing mixer config
Currently, when the allow spend from unmixed accounts in  the stake shuffle page is toggled, the mixer config settings cleared and user have to use the manual mixer config to remap the wallets.

Current we are using the mixerConfigset bool variable.

The effect of using this is that, it sets the mixer config value to a false state and prevents the privacy page from recognizing that mixer is already set.

The objective is to use a different config `SpendUnconfirmedFunds` so when checking for if mixer config is set, the page will not ask you to set up mixer accounts again

image

What should happen is to use a multiwallet config bool setting to set the option to spend from umixed accounts and not clear the mixer config settings.

crux25:
This PR Keeps the stake shuffle configs when spend from unmixed accounts is enabled. Closes #926.

### DIFF
--- a/ui/page/staking/purchase_modal.go
+++ b/ui/page/staking/purchase_modal.go
@@ -313,8 +313,8 @@ func (tp *stakingModal) initializeAccountSelector() {
 			// Imported and watch only wallet accounts are invalid for sending
 			accountIsValid := account.Number != dcrlibwallet.ImportedAccountNumber && !wal.IsWatchingOnlyWallet()
 
-			if wal.ReadBoolConfigValueForKey(dcrlibwallet.AccountMixerConfigSet, false) {
-				// privacy is enabled for selected wallet
+			if wal.ReadBoolConfigValueForKey(load.SpendUnmixedFundsKey, false) {
+				// Spending from unmixed accounts is disabled for wallet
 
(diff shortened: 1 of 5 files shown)


ui/page/staking/purchase_modal.go
```diff
@@ -313,8 +313,8 @@ func (tp *stakingModal) initializeAccountSelector() {
 			// Imported and watch only wallet accounts are invalid for sending
 			accountIsValid := account.Number != dcrlibwallet.ImportedAccountNumber && !wal.IsWatchingOnlyWallet()
 
-			if wal.ReadBoolConfigValueForKey(dcrlibwallet.AccountMixerConfigSet, false) {
-				// privacy is enabled for selected wallet
+			if wal.ReadBoolConfigValueForKey(load.SpendUnmixedFundsKey, false) {
```
Sirmorrison:
```suggestion
			if !wal.ReadBoolConfigValueForKey(load.SpendUnmixedFundsKey, false) {
```

Sirmorrison (changes requested):

Sirmorrison (approved):

PR #934, opened 2022-05-19
bump gio version

### LINKED ISSUE #813: Upgrade to gioui latest release
We should upgrade to the latest release of gioui. Some issues are already fixed upstream e.g #757 - see: [Gioui List,](https://pkg.go.dev/gioui.org@v0.0.0-20220318070519-8833a6738a3b/layout#List) we should upgrade to avoid duplication of what have already been accomplished upstream.

Sirmorrison:
This fix #813 

This PR bumps gio version to the latest release.


dreacot (dismissed):

dreacot:
just a note, seeing as gio is being updated, it might be important to note the required dependencies on the official website(https://gioui.org/doc/install/linux) have also been updated.

go.mod
```diff
@@ -25,6 +25,117 @@ require (
 	golang.org/x/text v0.3.7
 )
 
+require (
+	decred.org/cspp/v2 v2.0.0 // indirect
```
dreacot, line 29: run `go mod tidy` to remove the indirects

dreacot (changes requested):

go.mod
Sirmorrison: Nope it won’t… it came in due to the go version. 1.17 I checked it out already and it is supposed to happen. Same happened to the gioui when it was updated to the a higher go version.

it is a dependency issue.

dreacot (approved):

PR #939, opened 2022-05-22
Fix UI lag on transactions page

### LINKED ISSUE #757: having a wallet with lots of transactions lags the UI
I have a wallet with about 1300+ transactions, on the transactions page, selecting the said wallet from the dropdown immediately lags the UI significantly. looking at the scrollbar from the screenshot below, you can have an idea of how far the scroll is

While this particular case might be unlikely for most users to face, it begs to question that we might not be fetching transactions efficiently.

A solution to this would be to paginate the transactions, maybe about 20 transactions returned per page

Screenshot 2022-01-10 at 12 27 45 AM

devchoplife:
Closes #757 

This PR implements the pagination of transactions on the transactions page. The following were implemented:
- Adds a `view more` button if the number of transactions is more than 20
- Appends transactions in scores when the `view more` button is clicked until all transactions are displayed
- Hides the `view more` button when all transactions are displayed

### DIFF
--- a/ui/decredmaterial/clickable_list.go
+++ b/ui/decredmaterial/clickable_list.go
@@ -8,7 +8,7 @@ import (
 type ClickableList struct {
 	layout.List
 	theme           *Theme
-	clickables      []*Clickable
+	Clickables      []*Clickable
 	Radius          CornerRadius // this radius is used by the clickable
 	selectedItem    int
 	DividerHeight   unit.Dp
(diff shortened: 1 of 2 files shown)


dreacot:
I haven't reviewed the written code or tested the app yet, but I have to point out a view more button isn't the right way to go. It should be an infinite scroll.

Meaning, once the last item on the list has been scrolled to(or few items to the last), then the next set of transactions should be fetched and added to the list.

PR #948, opened 2022-05-23
switch to bottom nav, when on mobile devices

dreacot:
When on devices with screen width below `479px`, we can safely assume the user is on a mobile device and rearrange the layout.

This PR replaces the side bar with a bottom navigation bar when on screen-sizes below `479px`

**How can we achieve multiple layouts?**

Each page has a `Layout(gtx layout.Context) D {}` method, we can create 2 further layout to address mobile and desktop view(based on the current app width, we can say tablets are accounted for, although tests would still be carried out for such purposes).

The 2 layouts that need to be created are `layoutDesktop(gtx layout.Context) layout.Dimensions {}` and `layoutMobile(gtx layout.Context) layout.Dimensions {}`

**How do we decide which layout to show?**

We can check if the max screen width is below or equals the chosen mobile pixels (`479px`) e.g `if gtx.Constraints.Max.X <= gtx.Px(unit.Sp(400)) {}`

So for a full layout method, we should have;
```
	if gtx.Constraints.Max.X <= gtx.Px(unit.Sp(479)) {
		return mp.layoutMobile(gtx)
	}
	return mp.layoutDesktop(gtx)
}
```

**How can we compile for Mobile devices?**

**Android**
From the [Gioui Docs for building Android](https://gioui.org/doc/install/android)

`gogio -target android .`

This would generate a `godcr.apk`

**iOS**
From the [Gioui Docs for building iOS](https://gioui.org/doc/install/ios)

`gogio -o godcr.app -target ios .`

This would generate a `godcr.app`

**Screenshots**
ezgif com-optimize

### DIFF
--- a/ui/page/components/bottom_nav.go
+++ b/ui/page/components/bottom_nav.go
@@ -1,0 +1,146 @@
+package components
+
+import (
+	"gioui.org/layout"
+	"gioui.org/unit"
+
+	"github.com/planetdecred/godcr/ui/decredmaterial"
+	"github.com/planetdecred/godcr/ui/load"
+	"github.com/planetdecred/godcr/ui/values"
+)
+
+var (
+	bottomNavigationBarHeight = values.MarginPadding100
+)
+
+type BottomNavigationBarHandler struct {
+	Clickable     *decredmaterial.Clickable
+	Image         *decredmaterial.Image
+	ImageInactive *decredmaterial.Image
+	Title         string
+	PageID        string
+}
+
+type BottomNavigationBar struct {
+	*load.Load
+
+	AppBarNavItems       []BottomNavigationBarHandler
+	BottomNaigationItems []BottomNavigationBarHandler
+	CurrentPage          string
+
+	axis        layout.Axis
+	textSize    unit.Value
+	bottomInset unit.Value
+	height      unit.Value
+	alignment   layout.Alignment
+	direction   layout.Direction
+}
+
+func (bottomNavigationbar *BottomNavigationBar) LayoutBottomNavigationBar(gtx layout.Context) layout.Dimensions {
+	return layout.Stack{Alignment: layout.S}.Layout(gtx,
+		layout.Expanded(func(gtx C) D {
+			return UniformPadding(gtx, func(gtx C) D {
+				return decredmaterial.LinearLayout{
+					Width:       decredmaterial.WrapContent,
+					Height:      decredmaterial.WrapContent,
+					Orientation: layout.Horizontal,
+					Background:  bottomNavigationbar.Theme.Color.Surface,
+				}.Layout(gtx,
+					layout.Rigid(func(gtx C) D {
+						list := layout.List{Axis: layout.Horizontal}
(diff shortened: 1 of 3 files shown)


ui/page/components/bottom_nav.go
```diff
@@ -0,0 +1,144 @@
...
+// 									return bottomNavigationbar.AppBarNavItems[i].Image.Layout24dp(gtx)
+// 								})
+// 							})
+// 					}),
+// 					layout.Rigid(func(gtx C) D {
+// 						return layout.Inset{
+// 							Left: values.MarginPadding0,
+// 						}.Layout(gtx, func(gtx C) D {
+// 							return layout.Center.Layout(gtx, func(gtx C) D {
+// 								return bottomNavigationbar.Theme.Body1(bottomNavigationbar.AppBarNavItems[i].Title).Layout(gtx)
```
Sirmorrison: lets remove is not needed.

ui/page/main_page.go
```diff
@@ -636,6 +705,23 @@ func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
 	)
 }
 
+func (mp *MainPage) layoutMobile(gtx layout.Context) layout.Dimensions {
+	return layout.Flex{Axis: layout.Vertical}.Layout(gtx,
+		layout.Flexed(0.125, func(gtx C) D {
+			return mp.LayoutTopBar(gtx)
+		}),
```
```suggestion
		layout.Flexed(0.125, mp.LayoutTopBar),
```

ui/page/main_page.go
```diff
@@ -636,6 +705,23 @@ func (mp *MainPage) Layout(gtx layout.Context) layout.Dimensions {
...
+		layout.Flexed(0.75, func(gtx C) D {
+			if mp.currentPage == nil {
+				return layout.Dimensions{}
+			}
+			return mp.currentPage.Layout(gtx)
+		}),
+		layout.Flexed(0.125, func(gtx C) D {
+			return mp.bottomNavigationBar.LayoutBottomNavigationBar(gtx)
+		}),
```
```suggestion
		layout.Flexed(0.125, mp.bottomNavigationBar.LayoutBottomNavigationBar),
```

Sirmorrison (changes requested):

ui/values/dimensions.go
```diff
@@ -90,6 +90,6 @@ var (
 	TextSize34   = unit.Sp(34)
 	TextSize32   = unit.Sp(32)
 
-	AppWidth  = unit.Sp(800)
-	AppHeight = unit.Sp(600)
+	AppWidth  = unit.Sp(428)
```
Sirmorrison: The reason we had this minimum with was to protect some vital layouts. example are, 

image

image

image

I understand that this PR is only a part to a larger PR.. however, this change is going to break the current ui.

What we can do, is to revert this change and while testing your PRs and subsequent PRs, you can modify.

Sirmorrison (changes requested):

ui/values/dimensions.go
dreacot: okay noted, thanks for pointing it out

Sirmorrison (approved):

Issue #941, opened 2022-05-22
removing a watch only wallet crashed the app


dreacot:
If the watch only wallet is the only wallet in the list, attempting to delete it, causes the app to crash

```
2022-05-22 21:17:26.515 [INF] DLWL: Deleting Wallet
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x15225d4]

goroutine 25 [running]:
github.com/planetdecred/godcr/ui.(*Window).drawWindowUI(0xc0001d84b0, 0x320, 0x258, 0x320, 0x258, 0x3f8000003f800000, 0x1f2c780, 0xc0004bd718, 0xc09ac4759eb3f690, 0xadb25b05c1, ...)
        /home/kennedy/Projects/src/github.com/dreacot/godcr/ui/window.go:264 +0x154
github.com/planetdecred/godcr/ui.(*Window).displayWindow(0xc0001d84b0, 0xc09ac4759eb3f690, 0xadb25b05c1, 0x2cea180, 0x3f8000003f800000, 0x320, 0x258, 0x0, 0x0, 0x0, ...)
        /home/kennedy/Projects/src/github.com/dreacot/godcr/ui/window.go:224 +0x198
github.com/planetdecred/godcr/ui.(*Window).HandleEvents(0xc0001d84b0)
        /home/kennedy/Projects/src/github.com/dreacot/godcr/ui/window.go:187 +0x198
main.main.func2(0xc0001d84b0, 0xc00023ad00)
        /home/kennedy/Projects/src/github.com/dreacot/godcr/main.go:80 +0x2b
created by main.main
        /home/kennedy/Projects/src/github.com/dreacot/godcr/main.go:79 +0x4c5
```

Sirmorrison:
Resolved

PR #954, opened 2022-05-26
Watch-only wallet delete bug

### LINKED ISSUE #942: after deleting a watch only wallet, the watch-only section still shows
Screenshot from 2022-05-22 22-16-09

crux25:
This PR fixes the watch-only wallet crash delete bug. Hides watch-only wallet layout after the last watch-only wallet is deleted and shows loader widget to indicate that a process is going on when a watch-only wallet is being deleted.  Closes #942 and #941.

https://user-images.githubusercontent.com/4796738/170577854-4c6324ad-9b1d-4320-ac71-b4b2685bcf58.mp4

### DIFF
--- a/ui/modal/info_modal.go
+++ b/ui/modal/info_modal.go
@@ -105,9 +105,9 @@ func (in *InfoModal) CheckBox(checkbox decredmaterial.CheckBoxStyle, mustBeCheck
 	return in
 }
 
-func (pm *InfoModal) SetLoading(loading bool) {
-	pm.isLoading = loading
-	pm.modal.SetDisabled(loading)
+func (in *InfoModal) SetLoading(loading bool) {
+	in.isLoading = loading
+	in.modal.SetDisabled(loading)
 }
 
 func (in *InfoModal) Title(title string) *InfoModal {
--- a/ui/modal/info_modal.go
+++ b/ui/modal/info_modal.go
@@ -7,6 +7,8 @@ import (
 	"gioui.org/io/key"
 	"gioui.org/layout"
 	"gioui.org/text"
+	"gioui.org/widget/material"
+
 	"github.com/planetdecred/godcr/ui/decredmaterial"
 	"github.com/planetdecred/godcr/ui/load"
 	"github.com/planetdecred/godcr/ui/values"
@@ -39,8 +41,10 @@ type InfoModal struct {
 	mustBeChecked bool
 
 	titleAlignment, btnAlignment layout.Direction
+	materialLoader               material.LoaderStyle
 
 	isCancelable bool
+	isLoading    bool
 }
 
 func NewInfoModal(l *load.Load) *InfoModal {
@@ -51,11 +55,14 @@ func NewInfoModal(l *load.Load) *InfoModal {
 		btnPositve:   l.Theme.OutlineButton(values.String(values.StrYes)),
 		btnNegative:  l.Theme.OutlineButton(values.String(values.StrNo)),
 		isCancelable: true,
+		isLoading:    false,
 		btnAlignment: layout.E,
 	}
 
 	in.btnPositve.Font.Weight = text.Medium
 	in.btnNegative.Font.Weight = text.Medium
+
+	in.materialLoader = material.Loader(l.Theme.Base)
 
 	return in
 }
@@ -96,6 +103,11 @@ func (in *InfoModal) CheckBox(checkbox decredmaterial.CheckBoxStyle, mustBeCheck
 	in.checkbox = checkbox
 	in.mustBeChecked = mustBeChecked // determine if the checkbox must be selected to proceed
 	return in
+}
+
+func (pm *InfoModal) SetLoading(loading bool) {
+	pm.isLoading = loading
+	pm.modal.SetDisabled(loading)
 }
 
 func (in *InfoModal) Title(title string) *InfoModal {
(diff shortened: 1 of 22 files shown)


ui/modal/info_modal.go
```diff
@@ -171,22 +183,31 @@ func (in *InfoModal) HandleKeyEvent(evt *key.Event) {
...
+		}
 		isChecked := false
 		if in.checkbox.CheckBox != nil {
 			isChecked = in.checkbox.CheckBox.Value
 		}
 
 		in.positiveButtonClicked(isChecked)
+		if !in.isLoading {
+			in.DismissModal(in)
+		}
```
Sirmorrison: i dont think you need this here again. 
You already return when loading is set on line 186... 
```suggestion
	     in.DismissModal(in)

```

Sirmorrison (changes requested):

ui/modal/info_modal.go
crux25: That is needed else modal will always dismiss even when we need to show the loading widget. https://github.com/planetdecred/godcr/blob/84fa15f87dcd236993f5e402a4b56b2e48e28fb5/ui/modal/info_modal.go#L186 this is needed just in case someone is super fast and clicks the button twice before the loading widget is activated.

Sirmorrison: what you have is a for loop meaning for every click, it starts a new instance.

Sirmorrison (changes requested):

ui/modal/info_modal.go
crux25: I didn't write the for loop. I only modified it's content. I'll read the connected codes to figure out if for loop is appropriate there.

ui/modal/info_modal.go
```diff
@@ -170,23 +182,32 @@ func (in *InfoModal) HandleKeyEvent(evt *key.Event) {
 }
 
 func (in *InfoModal) Handle() {
-	for in.btnPositve.Clicked() {
-		in.DismissModal(in)
+	if in.btnPositve.Clicked() {
```
Sirmorrison: Kindly revert this.

ui/modal/info_modal.go
Sirmorrison: The review was not if you wrote it or if the for loop was wrong... it was for the redundant lines of codes to be removed.

Sirmorrison (changes requested):

dreacot (dismissed):

dreacot:
@Sirmorrison i've approved, kindly confirm if your review still stands before i merge

ui/modal/info_modal.go
```diff
@@ -171,22 +183,28 @@ func (in *InfoModal) HandleKeyEvent(evt *key.Event) {
...
-		in.DismissModal(in)
 		isChecked := false
 		if in.checkbox.CheckBox != nil {
 			isChecked = in.checkbox.CheckBox.Value
 		}
 
 		in.positiveButtonClicked(isChecked)
+		if !in.isLoading {
+			in.DismissModal(in)
+		}
```
Sirmorrison: ideally, i will expect this `isLoading` check to be at the top just before line 186. 
This way, nothing else get executed when there is a process running.
```
if in.isLoading{
     return
}
```

Sirmorrison (changes requested):
Besides the review below, the pr looks good to go.

ui/modal/info_modal.go
crux25: This has been a point of confusion, so I'll try to explain in more detail.
https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L192-L194 cannot be placed above https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L191 because this is the called back that's passed to https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L123

Think about it this way. When the positive button is clicked, an event is generated and https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L184 is called. At this point https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L47 is false because the call back passed to https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L123 hasn't been executed. At some point https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L191 -the callback- is executed and https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L47 is true. The consequence of placing that logic above the callback is that the modal will always dismiss before execution is completed - try it!!!. 

The way I figured to prevent double execution is to check if https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L47 is true twice first at the start of https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L184 execution and second after https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L191 -the callback- is executed. That's the point I was trying to get across to you earlier.

Other option is to rewrite https://github.com/planetdecred/godcr/blob/5f9db205405319305acce5ba8b7818a66d7f6ea9/ui/modal/info_modal.go#L123-L127

Sirmorrison: Alright. I think i get your point... you are trying to prevent the modal from dismissing when a process is still running.

Originally, the `positiveButtonClicked` callback just waits for the executable codes in the `PositiveButton()` where ever it is used to complete executing and send a feedback to the button to dismiss.

What we have currently are 2 methods running side by side and really does not need one to finish before the second one starts.

Before now, it was just an info modal and really was not designed what what we used it for. Hence the reason for the initial implementation as the `positiveButtonClicked` tend to execute completely before the modl dismisses.

That not withstanding, what we needed to have done was have the `positiveButtonClicked` function like that on the password modal. Have it return a `bool` value on then does the modal dismiss.

But we should check the `isLoading` at the top level on button click and return.

```
if positiveButtonClicked() {
   Dismiss()
}
```

crux25: That was my thought initially but I have to keep the task withing scope. I'll rewrite the callback to return a bool indicating if it is okay to dismiss the modal.

dreacot (dismissed):

Sirmorrison (approved):
LGTM

PR #962, opened 2022-05-31
Fixes hidden watch only menu

### LINKED ISSUE #940: watch only wallet menu hidden
If a watch only wallet is the only wallet in the list, the menu for the wallet doesn't show(hidden at the top right corner), and the scroll bar keeps showing, even if there's nothing to scroll to

Screenshot from 2022-05-22 21-11-24

Vanvians:
This resolves #940 

This makes it such that there is no more scroll bar on the side of the wallet page when there is only one thing on it. It also makes the options menu visible when its only a watch only wallet on the wallet list

image


Sirmorrison:
further work on this PR should be put on hold as we are migrating to the v2 layout.

PR #972, opened 2022-06-07
refactor page and modal navigation, remove some load methods

itswisdomagain:
Summary of key changes:
- Introduce `app.WindowNavigator` and `app.PageNavigator` interfaces to define page and modal navigation methods.
- Abstract page and modal navigation from the `ui/window.Window` object, use `app.SimpleWindowNavigator` instead.
- Abstract subpage navigation from `ui/MainPage` object, use `app.MasterPage` instead.
- Enable pages and modals perform page and modal navigations using the Navigator interfaces instead of using `ui/load/Load` methods for displaying and closing pages and modals.
- Enable pages, modals and navigators refresh the display using the `app.WindowNavigator.Reload()` method rather than `ui/load.Load.RefreshWindow()`.

The following Load methods are removed:
- RefreshWindow    func()
- ShowModal        func(Modal)
- DismissModal     func(Modal)
- ChangeWindowPage func(page Page, keepBackStack bool)
- PopWindowPage    func() bool
- ChangeFragment   func(page Page)
- PopFragment      func()
- PopToFragment    func(pageID string)
- ReloadApp        func()

Majority of the changes in this PR are minor, mostly changing the usages of those removed load methods to use the methods from the new `app` package.

This is part of a larger body of work to eliminate the load package and provide the needed operations in a more concise manner.


app/pagestack.go
```diff
@@ -0,0 +1,159 @@
...
+		pageStack.pages[l-1].OnNavigatedTo()
+	}
+	pageStack.debugLog()
+}
+
+func (pageStack *PageStack) debugLog() {
+	if l := len(pageStack.pages); l > 0 {
+		fmt.Printf("%s | page to be displayed: %s | stack: %d \n", pageStack.name, pageStack.pages[l-1].ID(), l)
+	} else {
+		fmt.Printf("%s | empty page stack \n", pageStack.name)
```
Sirmorrison, line 157: rather than use `fmt`, why not use `logger`?

ui/decredmaterial/modal.go
```diff
@@ -49,6 +59,28 @@ func (t *Theme) Modal() *Modal {
...
 
+// Dismiss removes the modal from the window. Does nothing if the modal was
+// not previously pushed into a window.
+func (m *Modal) Dismiss() {
+	// ParentWindow will only be accessible if this modal has been
+	// pushed into display by a WindowNavigator.
+	if parentWindow := m.ParentWindow(); parentWindow != nil {
+		parentWindow.DismissModal(m.ID())
+	} else {
+		panic("can't dismiss a modal that has not been displayed")
```
Sirmorrison, line 70: i have concerns with the `panic` used here.

Sirmorrison (approved):
LGTM

app/pagestack.go
itswisdomagain: debugLog was for test purposes, will remove in my next PR.

ui/decredmaterial/modal.go
itswisdomagain: This would help to catch wrong uses of `modal.Dismiss()` instead of just failing quietly.

PR #974, opened 2022-06-08
rework key event handling for updated gioui pkg

### LINKED ISSUE #960: tab behavior on restore page
Expected behavior is to move to the right to the next nubmer. Instead, it moves down to the next row.

itswisdomagain:
Resolves #960.

### DIFF
--- a/ui/page/wallets/restore_page.go
+++ b/ui/page/wallets/restore_page.go
@@ -3,6 +3,7 @@ package wallets
 import (
 	"image"
 
+	"gioui.org/io/key"
 	"gioui.org/layout"
 	"gioui.org/op/clip"
 	"gioui.org/op/paint"
@@ -254,4 +255,24 @@ func (pg *Restore) verifyHex(hex string) bool {
 	}
 
 	return true
+}
+
+// KeysToHandle returns an expression that describes a set of key combinations
+// that this page wishes to capture. The HandleKeyPress() method will only be
+// called when any of these key combinations is pressed.
+// Satisfies the load.KeyEventHandler interface for receiving key events.
+func (pg *Restore) KeysToHandle() key.Set {
+	if pg.tabIndex == 0 {
+		return pg.seedRestorePage.KeysToHandle()
+	}
+	return ""
+}
+
+// HandleKeyPress is called when one or more keys are pressed on the current
+// window that match any of the key combinations returned by KeysToHandle().
+// Satisfies the load.KeyEventHandler interface for receiving key events.
+func (pg *Restore) HandleKeyPress(evt *key.Event) {
+	if pg.tabIndex == 0 {
+		pg.seedRestorePage.HandleKeyPress(evt)
+	}
 }
--- a/app/pagestack.go
+++ b/app/pagestack.go
@@ -1,7 +1,6 @@
 package app
 
 import (
-	"fmt"
 	"sync"
 )
 
@@ -63,7 +62,6 @@ func (pageStack *PageStack) Push(newPage Page) {
 
 	pageStack.pages = append(otherPages, newPage)
 	newPage.OnNavigatedTo()
-	pageStack.debugLog()
 }
 
 // Pop removes the page at the top of the stack and gets the next page ready for
@@ -90,7 +88,6 @@ func (pageStack *PageStack) Pop() bool {
 	if l > 1 {
 		pageStack.pages[l-2].OnNavigatedTo() // get previous page ready for display
 	}
-	pageStack.debugLog()
 	return true
 }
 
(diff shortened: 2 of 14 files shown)


ui/page/wallets/restore_page.go
```diff
@@ -656,50 +656,80 @@ func (pg *Restore) HandleUserInteractions() {
...
+// that this page wishes to capture. The HandleKeyPress() method will only be
+// called when any of these key combinations is pressed.
 // Satisfies the load.KeyEventHandler interface for receiving key events.
-func (pg *Restore) HandleKeyEvent(evt *key.Event) {
+func (pg *Restore) KeysToHandle() key.Set {
+	if pg.isRestoring {
+		return "" // don't capture keys while restoring, problematic?
+	}
+	// TODO: Once user starts editing any of the input boxes, the arrow up, down
+	// and enter key signals are no longer received.
```
itswisdomagain: Only issue left: no idea how to fix yet, seems the editors from the upstream package no longer propagate arrow and enter key events to the caller even if requested.

Sirmorrison (changes requested):
https://user-images.githubusercontent.com/27733432/173068715-3a42bfd4-1b6d-40f0-8f08-269ce1ecdb5d.mov

So on the restore page, when the seed suggestion is open and the tab key is pressed, the expected behavior is auto fill the editor with the highlighted suggestion but in the scenario, it move to the editor at the bottom.

itswisdomagain:
> the expected behavior is auto fill the editor with the highlighted suggestion

That wasn't the behaviour, was it?
Can have that implemented here though. Will do.

Although not being able to use the arrow keys to select an option from the dropdown is a concern. Can we open an issue upstream for propagating arrow key events to consumers when a user is editing an Editor widget?

Sirmorrison:
> > the expected behavior is auto fill the editor with the highlighted suggestion
> 
> That wasn't the behaviour, was it? Can have that implemented here though. Will do.
> 
> Although not being able to use the arrow keys to select an option from the dropdown is a concern. Can we open an issue upstream for propagating arrow key events to consumers when a user is editing an Editor widget?

it was the behavior

itswisdomagain:
Update:
- Tab works as expected when the active input field doesn't have any text. If there's text, the default tab behaviour occurs.
- Enter selects the highlighted seed word suggestion.
- Arrow up and down keys don't work when the active input field has text.

Sirmorrison:
I think this is a conversation i would need to take upstream with gio devs.

Sirmorrison:
Ok doing some more research, i found this announcement and an alternative to how to use the global key events was provided and i believe it should help in some form.

https://go.dev/play/p/SDHy1LZRljf.

This is community discussion around the key event handler and provides various alternatives to having key issues resolved.
https://lists.sr.ht/~eliasnaur/gio/%3CCAFcc3FQNTp_UXr7oA97SsVPD7D91jSw30ZtALcT9vmopFDTeZQ%40mail.gmail.com%3E

itswisdomagain:
New Update:

> * ~Tab works as expected when the active input field doesn't have any text. If there's text, the default tab behaviour occurs.~ Tab works now whether or not the active input field has text.
> * ~Enter selects the highlighted seed word suggestion.~ Trying to get this to work opens up a can of worm. An ideal fix will require refactoring the restore page but hopefully, someone can come up with a simpler, quick fix.
> * ~Arrow up and down keys don't work when the active input field has text.~ They work now, depends on the cursor position. If the cursor is at the end, the down arrow works but not the up arrow. If the cursor is at the start, the up arrow works but not the down arrow.

itswisdomagain:
Digging deeper to figure out why the enter key stopped working as expected, I came up with https://github.com/itswisdomagain/godcr-1/commit/c9dd6fd808f49b0129207703099f009528b85ced.

The last commit that had enter key working fine on the restore page was https://github.com/itswisdomagain/godcr-1/commit/48f5e262c67408fbc0f8a831a8cb873d4b3aa00f.

The commit after that was https://github.com/itswisdomagain/godcr-1/commit/6b0470959a2710598a28341690aa5bb691655fd1 which understandably introduced issues with all key event handling.

But even if the new key event handling approach was implemented immediately after the gioui version bump, the enter key still doesn't work as expected on the restore page. https://github.com/itswisdomagain/godcr-1/commit/c9dd6fd808f49b0129207703099f009528b85ced implements the new key event handling on top of the commit that bumped the gioui version, to demonstrate that the enter key issue on the restore page seems to have come with the gioui version bump and requires more to fix than just implementing the new pattern for key event handling.

Sirmorrison:
PR looks good... issues with enter key not working on the restore page will be addressed differently.

Sirmorrison (approved):

PR #1000, opened 2022-06-20
fix governance click bug

### LINKED ISSUE #997: buttons don't respond to click events on the governance page
since the app package was implemented, buttons on the governance sub pages(Proposals, Consensus), no longer respond to click events.

Sirmorrison:
fix #997 

Previously, the user interaction on the governance sub pages were not being handled by the master page(Governance page), so all click events were ignored. The solution was to initialize the `HandleUserInteractions()` method of the sub pages from the governance page.

Fix app crash when the reject dropdown option is selected. This was as a result of integer from the error. 0/0 = infinity converting this to floating points

### DIFF
--- a/ui/page/components/votebar_widget.go
+++ b/ui/page/components/votebar_widget.go
@@ -260,12 +260,13 @@ func (v *VoteBar) layoutIconAndText(gtx C, lbl decredmaterial.Label, count int, 
 				return lbl.Layout(gtx)
 			}),
 			layout.Rigid(func(gtx C) D {
-				percentage := (count / v.totalVotes) * 100
+				count := float64(count)
+				percentage := (count / float64(v.totalVotes)) * 100
 				if percentage < 0 {
 					percentage = 0
 				}
-				percentageStr := strconv.FormatFloat(float64(percentage), 'f', 1, 64) + "%"
-				countStr := strconv.FormatFloat(float64(count), 'f', 0, 64)
+				percentageStr := strconv.FormatFloat(percentage, 'f', 1, 64) + "%"
+				countStr := strconv.FormatFloat(count, 'f', 0, 64)
 
 				return v.Theme.Body1(fmt.Sprintf("%s (%s)", countStr, percentageStr)).Layout(gtx)
 			}),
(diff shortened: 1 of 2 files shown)


ui/page/components/votebar_widget.go
```diff
@@ -260,12 +260,13 @@ func (v *VoteBar) layoutIconAndText(gtx C, lbl decredmaterial.Label, count int,
 				return lbl.Layout(gtx)
 			}),
 			layout.Rigid(func(gtx C) D {
-				percentage := (count / v.totalVotes) * 100
+				count := float64(count)
+				percentage := (count / float64(v.totalVotes)) * 100
 				if percentage < 0 {
```
itswisdomagain, line 265: Can `count` or `v.totalVotes` be less than 0? If that's not possible, this check is useless.

itswisdomagain (approved):

itswisdomagain:
Not sure what was causing `app crash when the reject dropdown option is selected`, based on the changes in this PR. Can you call that out?

ui/page/components/votebar_widget.go
Sirmorrison: I think when implementing this earlier, we ran into this issue hence the check.

Sirmorrison:
> Not sure what was causing `app crash when the reject dropdown option is selected`, based on the changes in this PR. Can you call that out?

yeah.. the was as a result of integer from the error. `0/0 = infinity` converting this to floating points, resolved the issue.

ui/page/components/votebar_widget.go
itswisdomagain: Weird, but okay. I think it'd be good to log if either count or v.totalVotes were negative.

itswisdomagain:
> result of integer from the error. `0/0 = infinity`

should ideally NOT do a division if the denominator is 0

dreacot:
`Fix app crash when the reject dropdown option is selected`. can you update the PR description on what the cause of the crash is, and what the solution is

Sirmorrison:
> `Fix app crash when the reject dropdown option is selected`. can you update the PR description on what the cause of the crash is, and what the solution is

Done

dreacot (dismissed):

ui/page/components/votebar_widget.go
dreacot: @Sirmorrison are we going to be logging the output in this PR?

Sirmorrison: No need logging it as i dont see the use.

dreacot (approved):

PR #1027, opened 2022-07-28
Activate Proposal search

### LINKED ISSUE #715: Search does not return results on Proposals page
When a search query is entered in the search field on proposals page, no results are returned. I tried clicking on the search icon and there were still no results

crux25:
This PR activates proposal search functionality and closes #715


Sirmorrison:
I believe the search functionality would have best been implement on dcrlibwallet. Just like we have the tx filters.

crux25:
> I believe the search functionality would have best been implement on dcrlibwallet. Just like we have the tx filters.

Any specific reason you believe this belongs in dcrlibwallet?

Sirmorrison:
> > I believe the search functionality would have best been implement on dcrlibwallet. Just like we have the tx filters.
> 
> Any specific reason you believe this belongs in dcrlibwallet?

I feel it would be easier and faster to so the sorting directly from the DB and we only have filtered data on the GUI.
What we currently have is pulling all data to the GUI from the backend and the sorting through them.

But again, i would like to understand the approach used for the search.. are we using instant search/sort/filter ie using regex where the filter happens per word typed or we completely type and use a search button?
N?B i have not gone through the logic in details yet hence the question.

crux25:
> > > I believe the search functionality would have best been implement on dcrlibwallet. Just like we have the tx filters.
> > 
> > 
> > Any specific reason you believe this belongs in dcrlibwallet?
> 
> I feel it would be easier and faster to so the sorting directly from the DB and we only have filtered data on the GUI. What we currently have is pulling all data to the GUI from the backend and the sorting through them.
> 
> But again, i would like to understand the approach used for the search.. are we using instant search/sort/filter ie using regex where the filter happens per word typed or we completely type and use a search button? N?B i have not gone through the logic in details yet hence the question.

Read the PR to understand it's implementation.

Sirmorrison:
> > > > I believe the search functionality would have best been implement on dcrlibwallet. Just like we have the tx filters.
> > > 
> > > 
> > > Any specific reason you believe this belongs in dcrlibwallet?
> > 
> > 
> > I feel it would be easier and faster to so the sorting directly from the DB and we only have filtered data on the GUI. What we currently have is pulling all data to the GUI from the backend and the sorting through them.
> > But again, i would like to understand the approach used for the search.. are we using instant search/sort/filter ie using regex where the filter happens per word typed or we completely type and use a search button? N?B i have not gone through the logic in details yet hence the question.
> 
> Read the PR to understand it's implementation. 


`This PR activates proposal search functionality and closes https://github.com/planetdecred/godcr/issues/715` this is your PR description... not much to read.

crux25:
> 

I am referring to the implementation.

PR #1057, opened 2023-12-05
Update README.md

dreacot:


xaur:
How about making it more visible? Make it bold, add a warning emoji (⚠️), and mention that Cryptopower includes everything GoDCR had and more (I assume it is)?

dreacot:
> How about making it more visible? Make it bold, add a warning emoji (⚠️), and mention that Cryptopower includes everything GoDCR had and more (I assume it is)?

i think linking should just be fine, once the link is clicked all the information needed is there

xaur:
Thats big and visible now, thanks!

